PR #1, opened 2017-06-15
added script to analyze latency scans

cbravo135:


latency/LatencyScanData.py
```diff
@@ -0,0 +1,86 @@
+#!/bin/env python2.7
```
jsturdy: unless you specifically need python 2.7, it's better to just make this ```#!/bin/env python```; as is, it will fail on any machine that doesn't have python 2.7

latency/LatencyScanData.py
```diff
@@ -0,0 +1,86 @@
...
+import numpy as np
+from array import array
+from optparse import OptionParser
+
+#Define and parse options
+parser = OptionParser()
+
+parser.add_option("-i", "--infilename", type="string", dest="filename", default="LatencyScanData.root",
+                  help="Specify Input Filename", metavar="filename")
+parser.add_option("-o", "--outfilename", type="string", dest="outfilename", default="SCurveFitData.root",
```
jsturdy: and if you're making the previous change, then the default here should be changed too

PR #2, opened 2017-07-12
Renaming Latency Analysis Script & Including Analysis Options

bdorney:
Addressing renaming suggested in:

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/99


bdorney:
Will submit a new PR based on discussion in July 18th GEM Operations & Online SW meeting which moves all analysis tools from vfatqc-python-scripts to gem-plotting-tools

Issue #3, opened 2017-07-19
Move Analysis Tools from vfatqc-python-scripts to this repo


bdorney:
Based on discussion we had in GEM Operations & Online SW Meeting on July 18th we decided to migrate all analysis based tools from vfatqc-python-scripts to this repo.

This impacts issues:
https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/84
https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/97
https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/99
https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/101

And also pull request:
https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/98

PR #4, opened 2017-07-19
Migrating Analysis Tools from vfatqc-python-scripts to gem-plotting-tools

bdorney:
Addressing:

https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/3
https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/99
https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/104

Also addressing:

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/98

Also built on:

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/100

Tested:

http://cmsonline.cern.ch/cms-elog/998496

Requires joint PR:

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/105

User should have following `$SHELL` variables declared:

`$BUILD_HOME`
`$DATA_PATH`
`$ELOG_PATH`

Then execute:

`setup/paths.sh`

Then you'll be ready to go.

Description of how to use ana_scans.py to analyze scan data is found:

https://twiki.cern.ch/twiki/bin/view/CMS/GEMDOCDoc#How_to_Produce_Scan_Plots

Note ana_scans.py was tested separately on a clone of $DATA_PATH on the gemdqm machine.  However I didn't make an E-Log documenting this (rookie mistake!).

Requesting review from @cbravo135 and @mexanick


anaUltraScurve.py
```diff
@@ -0,0 +1,355 @@
...
+
+if options.SaveFile:
+    masks = []
+    fitSums = {}
+    for vfat in range (0,24):
+        fitThr = []        
+        fitENC = []
+        strList = []
+        masks.append([])
+        for ch in range (0, 128):
```
cbravo135: This was not correctly merged. This should now be a loop over strips and also lookup_table should be changed to chToStrLUT.

bdorney: I didn't understand why lookup_table was seemingly split into:
`chToStrLUT = []
strToChLUT = []`

Looking at the code it wasn't clear why this was done.  Test outputs I ran are attached.  To me it seems like this implementation is "okay."

fitsummary
summary

cbravo135: fitSums x-axis is labeled Strips but the resulting output is in channel units this way. I do things for a reason, one of the two LUT replaces lookup_table and the other is new so I could make fitSums in strip units easily.

bdorney: The resulting x-axis is the same in both cases...if it's a concern over axis labeling then I would prefer the implementation shown in anaUltraThresholds.py:

https://github.com/bdorney/gem-plotting-tools/blob/master/anaUltraThreshold.py

Lines 81-119

If I understand this correctly it makes the output plots x-axis based on the three user input options automatically:
->strips
->vfat ch
->panasonic pins

Might be good to unify this.

cbravo135: Not true, look at my code harder. The code this way will definitely be in channel units.

This first version of fitSums is only in strip units, I will add other options when I clean up the code that takes care of switching between the three cases.

anaUltraScurve.py
```diff
@@ -0,0 +1,355 @@
...
+    pass
+if GEBtype == 'short':
+    #intext = open(buildHome+'/vfatqc-python-scripts/macros/shortChannelMap.txt', 'r')
+    intext = open(buildHome+'/gem-plotting-tools/setup/shortChannelMap.txt', 'r')
+    pass
+
+for i, line in enumerate(intext):
+    if i == 0: continue
+    mapping = line.rsplit('\t')
+    #chToStrLUT[int(mapping[0])][int(mapping[2]) - 1] = int(mapping[1])
```
cbravo135: Please use my more complete set of LUTs.

anaUltraScurve.py
bdorney: Look at the output plot, the distributions are following each other.  The x-axis labels may be wrong but the value plotted in fitSummary appears to follow Summary.  For example look at VFAT0, the spike matches in the two positions.  Also VFAT3 and VFAT4 match pretty well.  The rest appear to also.

I doubt this is just a coincidence.  Again it seems like the problem is the label on the x-axis

anaUltraScurve.py
bdorney: Again, would prefer the implementation shown in anaUltraThresholds.py:

https://github.com/bdorney/gem-plotting-tools/blob/master/anaUltraThreshold.py

Lines 81-119 to automatically ensure the correct mapping & x-axis label.

cbravo135: can we please change the name of lookup_table to chToStrLUT because it is more descriptive. lookup_table is a terrible variable name. I might as well call all my integers int1 int2 int3...

anaUltraScurve.py
cbravo135: that is right, strList fixes it, I forgot I did that. We should add panList and chList so we can control which we pass to build the TGraphErrors and can control the units of the x-axis then. As the code is now, it will always be in strip units.

cbravo135: Might be better to have one variable here instead of strList which gets passed to the TGraphErrors constructor which is filled depending on the users choice.

anaUltraScurve.py
bdorney: Fair point, will do, but I will use "strip" to avoid confusion with "string"

anaUltraScurve.py
```diff
@@ -0,0 +1,356 @@
...
+
+canv.SaveAs(filename+'/Summary.png')
+
+canv = r.TCanvas('canv','canv',500*8,500*3)
+canv.Divide(8,3)
+r.gStyle.SetOptStat(0)
+for vfat in fitSums.keys():
+    r.gStyle.SetOptStat(0)
+    canv.cd(vfat+1)
+    fitSums[vfat].Draw('ap')
```
lmoureaux: Crashes here when not using the -f option:

```Traceback (most recent call last):
  File "/home/louis/Documents/CERN Summer School/GEM/gem-plotting-tools/anaUltraScurve.py", line 367, in <module>
    for vfat in fitSums.keys():
NameError: name 'fitSums' is not defined```

Comment (bdorney) [inline anaUltraScurve.py]: At @cbravo135 is there are usage case where we would not want to use the option `-f` this seems like the primary use case...

Comment (cbravo135) [inline anaUltraScurve.py]: It is an easy fix anyways right? Just put the if protection like the rest of the blocks for the fitting option. The idea is without -f it only makes the Summary plot and runs faster since it is skipping all the fitting part. This plot doesn't make sense unless you do the fits, I just forgot to put the if part to protect it.

Comment (bdorney) [inline anaUltraScurve.py]: yes in fact I've already patched it, just making some updates to anaUltraLatency so it will make summary plot as a 8x3 grid

File: anaUltraScurve.py [hunk 4]
```diff
@@ -326,18 +323,20 @@ def overlay_fit(VFAT, CHAN):
...
-r.gStyle.SetOptStat(0)
-for vfat in fitSums.keys():
+if options.SaveFile:
+    canv = r.TCanvas('canv','canv',500*8,500*3)
+    canv.Divide(8,3)
     r.gStyle.SetOptStat(0)
-    canv.cd(vfat+1)
-    fitSums[vfat].Draw('ap')
-    canv.Update()
+    for vfat in fitSums.keys():
```
cbravo135: no reason to have .keys() here, that is what python uses by default when you iterate over a dict.

anaUltraLatency.py
```diff
@@ -83,4 +83,3 @@
```
cbravo135: Looks like you still haven't added the 8x3 summary plot yet, so I will wait to approve.

PR #5, opened 2017-07-21
Use the MAD algorithm to detect hot channels in `anaUltraScurve`

lmoureaux:
The MAD algorithm is currently used to detect hot channels in `anaUltraThreshold`. Use it in `anaUltraScurve`.

## Description
This PR uses the Median Absolute Deviation algorithm on the variable `threshold - ztrim * noise` to detect hot channels (hot channels have small values of `threshold - ztrim * noise`). This algorithm is already used on threshold data; this PR adds support for it on S-curve data.
In addition to the MAD itself, the following changes are also performed:
* Dead channels are detected from the fit results and explicitly masked (if they start producing data, they won't affect the measurements).
* The output ROOT tree contains information about why a channel was masked, in the form of a bitmask called `maskReason`.
* One additional plot is produced, called `PrunedSummary.png`. It is the equivalent of `Summary.png` but with hot channels masked.

### Types of changes

## How Has This Been Tested?

The changes were tested on `cosmicstandtif` Python 2.6.6, using data from http://cmsonline.cern.ch/cms-elog/998348. All hot channels are detected (at least those that I can see by looking at the data). Some "good" channels are masked too; the same behavior was already observed in `anaUltraThreshold`. The "fake hot" rate is of order 1/1000.

The `z-score`-dependence of the hot channel count was evaluated, and the default `z-score` of 3.5 looks close to optimal:
Number of detected hot channels vs `z-score`

It was checked that `anaUltraThreshold.py` produces the same results after fixing the bug in `isOutlier*` functions.

The change from 25 to 15 iterations in fit() was decided after looking at the data, and finding no significant improvement in the best `chi^2` after 10 iterations.

### Screenshots
S-curve summary:
summary
Pruned version:
prunedsummary

## Checklist:

- [ ] My change requires a change to the documentation. (There is no doc in the repo)
- [ ] I have updated the documentation accordingly. (There is no doc in the repo)


anaUltraScurve.py
```diff
@@ -284,44 +345,49 @@ def overlay_fit(VFAT, CHAN):
         pass
     pass
 
-canv = r.TCanvas('canv','canv',500*8,500*3)
-legend = r.TLegend(0.75,0.7,0.88,0.88)
-if not options.PanPin:
-    canv.Divide(8,3)
-    r.gStyle.SetOptStat(0)
-    for vfat in range(0,24):
+def saveSummary(vSum, vSum2, name='Summary'):
```
bdorney: Line 348, why not add the function "saveSummary()" to "anautilities.py" and add arguments for it taking input lists of histograms as arguments.

Seems like:
> anaUltraScurve.py
> anaUltraThreshold.py
> anaUltraLatency.py

Could all make use of this and this could also be used by future scripts.

Functionality should be user supplies list of histograms as an argument indexed by VFAT position, then saveSummary() will just make a single output canvas in the standard 3x8 (row x col) plot.  Then this function could be called for each list of histograms.

bdorney:
Also can you retest using native python?  As I mentioned on mattermost:

> On the p5 machines we have `python` which is version 2.6.6.  When machines are migrated to CC7 it will be version 2.7.12 but this has not happened yet so we still need to support v2.6.6

anaUltraScurve.py
```diff
@@ -210,16 +226,61 @@ def overlay_fit(VFAT, CHAN):
     trimrange_list[event.vfatN][event.vfatCH] = event.trimRange
     pass
```
lmoureaux: Running over all events only once is possible but would be a huge change by itself. I suggest to leave this optimization to another PR in order to make reviewing changes easier.

Getting the `strip` and the `pan_pin` for the event can only be done on an event-by-event basis. I don't see any other way to do it except always filling the histogram per-channel and shuffling bins afterwards.

Lines 224-226 can be put in any of the two "event loops", I don't see any good reason to touch these lines.

anaUltraScurve.py
```diff
@@ -210,16 +226,61 @@ def overlay_fit(VFAT, CHAN):
...
+    for vfat in range(0, 24):
+        trimValue = np.zeros(128)
+        dead = np.zeros(128)
+        for ch in range(0, 128):
+            # Get fit results
+            threshold[0] = scanFits[0][vfat][ch]
+            noise[0] = scanFits[1][vfat][ch]
+            pedestal[0] = scanFits[2][vfat][ch]
+            # Identify dead channels
+            dead[ch] = (threshold[0] == 8 and pedestal[0] == 8)
```
lmoureaux: I think we want to mask them in any case, the wording is bad. I'll dig into better dead channel identification, which would also help reducing the number of "fit data is empty" warnings.

anaUltraScurve.py
```diff
@@ -210,16 +226,61 @@ def overlay_fit(VFAT, CHAN):
...
+            # Compute the value to apply MAD on for each channel
+            trimValue[ch] = threshold[0] - options.ztrim * noise[0]
+        masksDead.append(dead)
+        # Determine outliers
+        hot = isOutlierMADOneSided(trimValue, thresh=options.zscore,
+                                   rejectHighTail=False).flatten()
```
lmoureaux: `isOutlierMADOneSided` returns an array like `[[0], [1], [2], ...]`, and I don't care about the second dimension. I don't think I can change `isOutlierMADOneSided` without breaking `anaUltraThreshold`

bdorney: Okay, I see, don't do this.  We should make the package as uniform as possible, having two different ways of implementing this usage will be harder to maintain over the longterm.

Please either:
- Change your implementation to match what is done in anaUltraThreshold.py, or
- Change anaUltraThreshold.py to use the implementation shown here.

I strongly favor the first option.  To me the use of "flatten" is unnecessary and confusing.  Please update.

anaUltraScurve.py
```diff
@@ -94,6 +95,10 @@
     myT.Branch( 'ROBstr', ROBstr, 'ROBstr/I' )
     mask = array( 'i', [ 0 ] )
     myT.Branch( 'mask', mask, 'mask/I' )
+    maskDead = array( 'i', [ 0 ] )
+    myT.Branch( 'maskDead', maskDead, 'maskDead/I' )
+    maskOutlier = array( 'i', [ 0 ] )
+    myT.Branch( 'maskOutlier', maskOutlier, 'maskOutlier/I' )
```
lmoureaux: > The output ROOT tree contains information about why a channel was masked.

anaUltraScurve.py
```diff
@@ -249,9 +310,9 @@ def overlay_fit(VFAT, CHAN):
             noise[0] = param1
             fitENC.append(vToQm*param1*options.ztrim)
             pedestal[0] = param2
-            if noise[0] > 20.0 or ped_eff[0] > 50.0: mask[0] = True
-            else: mask[0] = False
-            masks[vfat].append(mask[0])
+            maskOutlier[0] = masks[vfat][chan]
+            maskDead[0] = int(masksDead[vfat][ch])
+            mask[0] = masks[vfat][chan]
```
lmoureaux: `mask` should have been the logical OR of `maskDead` and `maskOutlier`. Fixing.

anaUltraScurve.py
bdorney: If this is the goal then we should change how this is implemented.  These branch names are not revealing.  Instead you should have:

`mask`
`maskReason`

And then `maskReason` stores an enum style mask code, for example:

-1 -> not masked
0 -> Hot channel
1 -> Dead channel

etc...

Additionally what's the difference between `maskOutlier` and `mask` in the code they appear to be identical.

anaUltraScurve.py
bdorney: I'd like to see how many times the fit fails and if this corresponds with dead channels or just bad fitting.  Please investigate.  We should be sure of the reasoning behind why we need to mask things especially moving toward LS2.

anaUltraScurve.py
bdorney: See comments above, I think we should rethink the implementation to be more descriptive to the user

anaUltraScurve.py
lmoureaux: About 1% (8 out of 776) of failed fits aren't due to dead channels.

bdorney: I think it would be important to demonstrate how these were identified and also include them in the maskReason.

This would be good to put into a talk for the Commissioning & Online SW meeting.

lmoureaux: The identification was changed to be more robust, but I still have exactly 8 failed fits. I'll put some slides together.

anaUltraScurve.py
lmoureaux: The problem was in fact a bug in `isOutlierMADOneSided`, which would sometimes return 1-dimensional arrays and sometimes 2-dimensional ones. See commit fd32c4a

anaUltraScurve.py
```diff
@@ -208,18 +223,73 @@ def overlay_fit(VFAT, CHAN):
...
 if options.SaveFile:
+    scanFits = fitter.fit()
+    pass
+
+# Determine hot channels
+import numpy as np
+if options.SaveFile:
+    print 'Determining hot channels'
     masks = []
+    maskReasons = []
```
bdorney: Let's define this container as a dict instead in `anaInfo.py` let's have the dictionary be (int, string) pairs where the key value is the int and the string is the human readable reason for being masked.  Then in the tree store the key value (e.g. the int).

I think this will be easier to maintain, read, and reference in the future.

anaUltraScurve.py
```diff
@@ -122,6 +126,8 @@
 
 vSum  = ndict()
 vSum2 = ndict()
```
bdorney: This was probably myself who used these variable names and it was a bad idea in retrospect.  Can we change these to be a bit more descriptive?  At the very least we should be able to understand how `vSum2` is different from `vSum` here.

anaUltraScurve.py
```diff
@@ -122,6 +126,8 @@
...
+vSumPruned = ndict()
+vSumPruned2 = ndict()
```
bdorney: same comment here, resolve ambiguity between `vSumPruned` and `vSumPruned2`

anaUltraScurve.py
```diff
@@ -208,18 +223,73 @@ def overlay_fit(VFAT, CHAN):
...
+    for vfat in range(0, 24):
+        trimValue = np.zeros(128)
+        fitFailed = np.zeros(128, dtype=bool)
+        for ch in range(0, 128):
+            # Get fit results
+            threshold[0] = scanFits[0][vfat][ch]
+            noise[0] = scanFits[1][vfat][ch]
+            pedestal[0] = scanFits[2][vfat][ch]
+            # Compute the value to apply MAD on for each channel
+            trimValue[ch] = threshold[0] - options.ztrim * noise[0]
```
bdorney: can you add `pass` after this line to show the nested loop has finished?

anaUltraScurve.py
```diff
@@ -284,56 +353,51 @@ def overlay_fit(VFAT, CHAN):
         pass
     pass
 
-canv = r.TCanvas('canv','canv',500*8,500*3)
-legend = r.TLegend(0.75,0.7,0.88,0.88)
-if not options.PanPin:
-    canv.Divide(8,3)
+def saveSummary(vSum, vSum2, name='Summary'):
```
bdorney: Let's move this definition to `anautilities.py` that way other analysis scripts can also take advantage of it.  Seems like this type of thing could be used in:

- `anaUltraLatency.py`
- `anaUltraScurve.py`
- `anaUltraThreshold.py`

anautilities.py
```diff
@@ -41,6 +42,17 @@ def initVFATArray(array_dtype, nstrips=128):
 
     return np.zeros(nstrips, dtype=list_dtypeTuple)
 
+def make3x8Canvas(name, initialContent = None, drawOption = ''):
```
bdorney: Can you add a comment above this which describes what `initialContent` is expected to be?  Looks like some dict or list of TObjects that have a `Draw()` method but this would clarify things.

anautilities.py
```diff
@@ -57,26 +69,18 @@ def rejectOutliersMADOneSided(arrayData, thresh=3.5, rejectHighTail=True):
 #Use inter-quartile range (IQR) to reject outliers
 #Returns a boolean array with True if points are outliers and False otherwise.
 def isOutlierIQR(arrayData):
-    if len(arrayData.shape) == 1:
-        arrayData = arrayData[:,None]
-        pass
-
```
bdorney: Are you sure this works as expected with these lines removed?  I would caution against this as during testing I found without these lines unexpected behavior occurred.

anautilities.py
```diff
@@ -57,26 +69,18 @@ def rejectOutliersMADOneSided(arrayData, thresh=3.5, rejectHighTail=True):
...
-    return (arrayData < (q1 - 1.5 * IQR)) or (arrayData > (q3 + 1.5 * IQR))
+    return (arrayData < (q1 - 1.5 * IQR)) | (arrayData > (q3 + 1.5 * IQR))
 
 #Use inter-quartile range (IQR) to reject outliers, but consider only high or low tail
 #Returns a boolean array with True if points are outliers and False otherwise.
 def isOutlierIQROneSided(arrayData, rejectHighTail=True):
-    if len(arrayData.shape) == 1:
-        arrayData = arrayData[:,None]
-        pass
-
```
bdorney: same comment as above

anautilities.py
```diff
@@ -93,15 +97,8 @@ def isOutlierIQROneSided(arrayData, rejectHighTail=True):
 #See: https://github.com/joferkington/oost_paper_code/blob/master/utilities.py
 #Returns a boolean array with True if points are outliers and False otherwise.
 def isOutlierMAD(arrayData, thresh=3.5):
-    if len(arrayData.shape) == 1:
-        arrayData = arrayData[:,None]
-        pass
-
```
bdorney: same comment as above

anautilities.py
```diff
@@ -113,15 +110,10 @@ def isOutlierMAD(arrayData, thresh=3.5):
 #Use MAD to reject outliers, but consider only high or low tail
 #Returns a boolean array with True if points are outliers and False otherwise.
 def isOutlierMADOneSided(arrayData, thresh=3.5, rejectHighTail=True):
-    if len(arrayData.shape) == 1:
-        arrayData = arrayData[:,None]
-        pass
-
```
bdorney: same comment as above

macros/fitScanData.py
```diff
@@ -1,87 +1,103 @@
+class DeadChannelFinder(object):
+    def __init__(self):
+        import numpy as np
```
bdorney: Is this the correct way to do this?  Looks like multiple classes/methods inside `DeadChannelFinder` and `ScanDataFitter` require numpy.

macros/fitScanData.py
```diff
@@ -1,87 +1,103 @@
...
-                    scanFits[4][vfat][ch] = scanCount[vfat][ch]
-                    scanFits[5][vfat][ch] = -999999
-                    MinChi2Temp = fitChi2
+    def readFile(self, treeFileName):
+        import ROOT as r
+        inF = r.TFile(treeFileName)
+        for event in inF.scurveTree :
+            self.feed(event)
+
+    def fit(self):
```
bdorney: I'd like @cbravo135 to take a look at this, it seems like it's  the same but I'd like a cross check

anautilities.py
lmoureaux: This solves a bug, see the commit message of 8cf13871.

anautilities.py
```diff
@@ -105,15 +97,8 @@ def isOutlierIQROneSided(arrayData, rejectHighTail=True):
 #See: https://github.com/joferkington/oost_paper_code/blob/master/utilities.py
 #Returns a boolean array with True if points are outliers and False otherwise.
 def isOutlierMAD(arrayData, thresh=3.5):
-    if len(arrayData.shape) == 1:
-        arrayData = arrayData[:,None]
-        pass
-
     median = np.median(arrayData, axis=0)
-
-    diff = np.sum((arrayData - median)**2, axis=-1)
```
lmoureaux: The problems @bdorney experienced when removing the reshaping code for 1D arrays was most likely caused by `np.sum` here. I simply removed this line because summing along the last axis was anyway not supported in `isOutlierIQR`.

macros/fitScanData.py
lmoureaux: I advise you to look at this one commit at a time, the changes will be easier to spot and understand this way. The relevant commits are between 5404cc5d and 52ca2c93. I think that the most controversial ones will be 04dfa6b4 (removes a conditional branch) and 3639c7fd (25 to 15 iterations).

anaUltraScurve.py
lmoureaux: What remains in `saveSummary` is `ultraScurve`-specific. The factorisable part was moved to `make3x8Canvas`.

anaUltraScurve.py
lmoureaux: The tuple thing isn't going to play well with the bitmasks. I added an enum-like class in `anaInfo.py` to address the readability/maintainability concern. There is also a `humanReadable(reason)` function there which returns e.g. `'HotChannel,FitFailed'`.

macros/fitScanData.py
bdorney: Okay, I left two comments on the last two commits, other than those I think this is good.

anautilities.py
bdorney: Okay, fair point.  Let's keep it this way then.

bdorney:
Approving changes, but before merging I'd like this update presented in the next operations and online software meeting (Tuesday August 15th, no indico agenda yet) to get any final feedback from the community as these changes could impact the p5 operations.

macros/fitScanData.py
cbravo135: Looks like the output of this new class is gonna be the same as how it was done before. I do like the addition, this is gonna make is easier for me to change the fitting code to use the correct statistical error bars in the future. This is going to require an update to trimChamber in vfatqc, so we need to make sure that change is implemented and both are merged simultaneously.

anaUltraScurve.py
```diff
@@ -249,9 +321,8 @@ def overlay_fit(VFAT, CHAN):
             noise[0] = param1
             fitENC.append(vToQm*param1*options.ztrim)
             pedestal[0] = param2
-            if noise[0] > 20.0 or ped_eff[0] > 50.0: mask[0] = True
```
cbravo135: removing this line changes how we decide when to mask a channel. Is this replaced somewhere?

lmoureaux: We don't need to mask noisy channels if their operational threshold (`threshold-zTrim*noise`) is high. The MAD algorithm will detect low operational thresholds.

There is no replacement for the effective pedestal cut. I only saw S-curve plots without a white band at the bottom (ie bins with zero events) in untrimmed data after the HV incident, and most of them weren't masked by `anaUltraScurve`. Re-adding the cut wouldn't hurt, and in this case I'd even make it tighter.

(Note: changes to the trimming procedure may affect the second point, but I'd say that if trimming results in channels having to be masked, then the trimming procedure is bad.)

cbravo135: threshold - zTrim*noise is not the operational threshold and actually called the trimming point. The operational threshold is a value in charge units representing the 50% point on the scurve after set by the final threshold scan. Also, there is a point where the noise is so high, we want to mask that channel anyways.

The ped_eff cut is not exactly for "S-curve plots without a white band at the bottom". This cut catches a a couple specific scurve structures I have observed over the past year or so of studying hundreds of s-curve plots, from many different chambers, while developing these tools. We can discuss removing it at some point, but we need a good way to catch the I observed this to mask at some point. We also will need to make sure we mask channels which give no response during the scurve.

lmoureaux: > We also will need to make sure we mask channels which give no response during the scurve.

This corresponds to `maskReason = MaskReason.DeadChannel` (no data at all) or `maskReason = MaskReason.FitFailed` (not enough data to fit).

bdorney: Since @cbravo135 is now on holiday for ~3 weeks I would suggest to reimplement this line and open a separate issue/feature request where we address the concerns from both @cbravo135 and @lmoureaux.

macros/fitScanData.py
bdorney: I believe @lmoureaux has a branch where this is addressed.  I would suggest that after https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/108 is merged @lmoureaux opens a PR with the goal of simultaneously merging this PR (in `gem-plotting-tools`) and the corresponding updates to `trimChamber.py` in `vfatqc-python-scripts`

cbravo135:
I have not seen a response to thr issue of making sure trimChamber is not
broken. Please post a test showing you have properly updated trimChamber.

lmoureaux:
> I have not seen a response to thr issue of making sure trimChamber is not broken. Please post a test showing you have properly updated trimChamber.

@cbravo135 Revision 52ca2c93 takes care of backwards compatibility of `fitScanData.py`.

bdorney:
@lmoureaux since no one is using the cosmic stand now why don't you do a trim run showing which commit your `vfatqc-python-scripts` and your `gem-plotting-tools` repos are on.  Then post the successful analysis output files and the e-log shortlink.

As mentioned if your `vfatqc-python-scripts` is ahead of the master then open an issue there and subsequent pull request.

I think this will be the definitive proof that @cbravo135 is asking for (and now @jsturdy will require since @cbravo135 will be on holiday).

lmoureaux:
Trim run showing that no update to `trimChamber.py` is needed:
http://cmsonline.cern.ch/cms-elog/1004412

lmoureaux:
@jsturdy Are the results at http://cmsonline.cern.ch/cms-elog/1004412 good enough? Do you need additional information?

Issue #6, opened 2017-07-24
Bug Report: ana_scans.py


bdorney:
http://cmsonline.cern.ch/cms-elog/999538

ana_scans.py has several bugs when being called:

`runCommand(cmd, log)` doesn't seems to generate the log file but doesn't call cmd given.  Changing to runCommand(cmd) calls the cmd.  Potential issue in "runCommand?"

Under anaType 'scurve' the cmd is improperly appended

When starting to build commands before the case structure in launchAnaArgs need to get the env variable for the path (e.g. GEM_PLOTTING_PROJECT) and then use this to call the command, e.g.:

cmdPath   = os.getenv('GEM_PLOTTING_PROJECT')
cmd = ["python", "%s/%s"%(cmdPath,ana_config[anaType])]

When tool=anaUltraThreshold.py a the depreciated option of --vfatmask was called leading to an exception.

PR #7, opened 2017-07-24
Bug Fixes to ana_scans.py

bdorney:
Addressing:

https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/6

Now also addressing:

https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/8
https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/9

Tested using my account on gemdqm machine:

http://cmsonline.cern.ch/cms-elog/999738

Testing completed.  Looks good.


ana_scans.py
```diff
@@ -14,18 +14,18 @@ def launchAnaArgs(anaType, cName, cType, scandate, scandatetrim=None, ztrim=4.0,
   from anaInfo import ana_config
   from gempython.utils.wrappers import runCommand
 
+  #cmdPath   = os.getenv('GEM_PLOTTING_PROJECT')
```
cbravo135: remove this line?

ana_scans.py
```diff
@@ -14,18 +14,18 @@ def launchAnaArgs(anaType, cName, cType, scandate, scandatetrim=None, ztrim=4.0,
...
   dataPath  = os.getenv('DATA_PATH')
   dirPath   = ""
   elogPath  = "%s/%s"%(os.getenv('ELOG_PATH'),scandate)
     
   print "Analysis Requested: %s"%(anaType)
 
   #Build Commands
-  cmd = ["python", ana_config[anaType]]
+  #cmd = ["python", "%s/%s"%(cmdPath,ana_config[anaType])]
```
cbravo135: also this line?

anaUltraLatency.py
```diff
@@ -4,82 +4,150 @@
...
+from ROOT import TH1F
+r.gROOT.SetBatch(True)
+r.gStyle.SetOptStat(1111111)
+inF = r.TFile(filename+'.root',"READ")
+
+#Initializing Histograms
+print 'Initializing Histograms'
+dict_hVFATHitsVsLat = ndict()
+for vfat in range(0,24):
+    dict_hVFATHitsVsLat[vfat]   = TH1F("vfat%iHitsVsLat"%vfat,"vfat%i"%vfat,256,-0.5,255.5)
```
cbravo135: this line can use r.TH1F() and the import above should be deleted

anaUltraLatency.py
```diff
@@ -4,82 +4,150 @@
...
-errY= array( 'f' )
-errX=array('f')
+print filename
+outputfilename = options.outfilename
 
-for vfat in range(24):
-    Nhits_hs[vfat] = r.TH1D("Nhits%i_h"%vfat,"Nhits%i_h"%vfat,256,-0.5,255.5)
-    Nev_hs[vfat] = r.TH1D("Nev%i_h"%vfat,"Nev%i_h"%vfat,256,-0.5,255.5)
+import ROOT as r
+from ROOT import TH1F
```
cbravo135: remove this import, the line above imports all of root to the r namespace

ana_scans.py
```diff
@@ -83,19 +86,18 @@ def launchAnaArgs(anaType, cName, cType, scandate, scandatetrim=None, ztrim=4.0,
...
-    postCmds[0] = ["mkdir","-p","%s"%(elogPath)]
     postCmds[1] = ["cp","%s/ThresholdScanData/ThreshSummary.png"%(dirPath),
                    "%s/ThreshSummary_%s.png"%(elogPath,cName)]
     postCmds[2] = ["cp","%s/ThresholdScanData/ThreshPrunedSummary.png"%(dirPath),
                    "%s/ThreshPrunedSummary_%s.png"%(elogPath,cName)]
-    postCmds[3] = ["cp","%s/ThresholdScanData/chConfig_MasksUpdated.txt"%(dirPath),
-                   "%s/chConfig_MasksUpdated_%s.txt"%(elogPath,cName)]
-    postCmds[4] = ["cp","%s/ThresholdScanData/vfatConfig.txt"%(dirPath),
+    postCmds[3] = ["cp","%s/ThresholdScanData/vfatConfig.txt"%(dirPath),
                    "%s/vfatConfig_%s.txt"%(elogPath,cName)]
```
cbravo135: This is okay for now, but we should discuss if this makes sense. These text files are not going to be posted to the elog ever, so maybe we should have a different place to put them.

ana_scans.py
```diff
@@ -123,9 +124,9 @@ def launchAnaArgs(anaType, cName, cType, scandate, scandatetrim=None, ztrim=4.0,
...
   try:
     log = file("%s/anaLog.log"%(dirPath),"w")
-   
-    runCommand(cmd,log)
-    #runCommand(cmd)
+ 
+    #runCommand(cmd,log)
+    runCommand(cmd)
     for key in postCmds:
       runCommand(postCmds[key])
```
cbravo135: Is there a reason why postCmds is a dict and not a list? Code would be cleaner if it was a list.

anaUltraLatency.py
bdorney: Ah yeah, I agree...sloppy, fixed.

anaUltraLatency.py
bdorney: Fixed

ana_scans.py
bdorney: Implementation requires it to be a dict.

It's a dictionary of lists, runCommand requires a list of strings as inputs to be the command.  We need **multiple** commands to be executed because multiple output files need to be copied, so it's configured as a dictionary.

e.g. syntax of cp command is cp <file> <target> and to keep with the naming scheme each file requires it's own executing of runCommand, so it's own entry in the postCmds dictionary.

ana_scans.py
bdorney: If we provide another environment variable, e.g. `$CONFIG_PATH` then maybe that's the solution.

Otherwise I think it would be convenient to have a single place to have all the files so the question:

"What are the thresholds set to" has an easy single entry point to access.

For now I would propose `$ELOG_PATH` and then in a later feature request we consider `$CONFIG_PATH`.

ana_scans.py
cbravo135: This is not true of python. postCmds can still be a list and you can append lists into it. Then when you iterate over postCmds you will get the lists which define the commands to be run. The difference between a list and a dict is that a dict uses a map to associate a key to the location of some data. Dicts are built on top of the list so you can access entries with keys instead of just having to pass an int which gets the object placed with the Nth call to append on the list object.

cbravo135: The point is you can append lists into a list...

bdorney: Yes I see what you mean, misunderstanding from my side.  I thought you were asking for it to be one long list of strings where each command is just the next entry in the list rather than a list of lists where each entry is a list specifying the command.

In this case yes it could be this way and I agree it will be cleaner rather than having to continually increase the key idx.  I will update.

bdorney: Updated and tested on gemdqm machine at P5

Issue #12, opened 2017-08-10
Add xdaq Latency Scan python routine to repo


bdorney:
The xdaq latency scan routine is currently sitting on the 904 NAS and is not tracked via subversion.

Suggest to incorporate the script into this repository.

Propose also to rename it following some sort of convention.  The analysis scripts that deal with the python tools all start with:

`anaUltra*`

Analysis scripts that deal with scans taken with xdaq might start with:

`anaXDAQ*`

bdorney:
We should also update the options menu of `anaXDAQLatency.py` to match with the conventions set in the rest of `cms-gem-daq-project`

For example:

```
anaXDAQLatency.py -h
Usage: anaXDAQLatency.py [options]

Options:
  -h, --help            show this help message and exit
  -i infile, --infile=infile
                        Input file to process
  --amc13=amc13         AMC13 to look at
  --amc=amc             AMC to look at
  --gtx=gtx             GTX to look at
  --min=min             Min parameter range to look at
  --max=max             Max parameter range to look at
```

Propose to change:
- `--amc` option to `-s`
- `--gtx` option to `-g`

This will match better with what users are familiar with from common options found in [standardopts.py](https://github.com/cms-gem-daq-project/cmsgemos/blob/generic-amc-release-v3/gempython/utils/standardopts.py) from `cmsgemos`.

It may also be useful to change:
- `--min` option to `--scanmin`
- `--max` option to `--scanmax`

As this will match better with common options found in [qcoptions](https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/master/qcoptions.py) from `vfatqc-python-scripts`

Issue #11, opened 2017-08-10
Add repo structure 


bdorney:
Following from cmsgemos and vfatqc-python-scripts:

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/95

Add in contributing guidelines and templates (mostly copied from cmsgemos).

Suggest to copy the files from .github of vfatqc-python-scripts and place them here:

- `CONTRIBUTING.md`
- `ISSUE_TEMPLATE.md`
- `PULL_REQUEST_TEMPALTE.md`

And make the corresponding updates to `.gitignore` e.g.:

```
# Don't ignore special files
!*.md
!*.rst

# Don't ignore subdirectories
!*/
```

Issue #10, opened 2017-08-10
Remove `cd` usage from paths.sh


bdorney:
Following:

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/109

> Sourcing the setup/paths.sh script leaves the user in a specific directory, which is not good practice, especially when used in a profile init script

For paths.sh here we should modify blocks at:

- [#L42](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/master/setup/paths.sh#L42)
- And everything below [#L55](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/master/setup/paths.sh#L55)

PR #13, opened 2017-08-10
Addressing Setup, Repo Structure, and Missing xdaq Latency Scan Application

bdorney:
Removing usage of `cd` in `setup/paths.sh`  This should address: https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/10

Adding repo structure files requested in https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/11:

- `CONTRIBUTING.md`
- `ISSUE_TEMPLATE.md`
- `PULL_REQUEST_TEMPALTE.md`

And made updates to `.gitignore` as discussed in https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/11.

Finally added anaXDAQLatency.py script to the repository as mentioned in https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/12.  This script was renamed from:

`/data/bigdisk/GEM-Data-Taking/latencyAnalysis.py`

Created by @jsturdy for analyzing latency scan data taken with xdaq.


setup/paths.sh
```diff
@@ -37,38 +37,27 @@ else
     source $BUILD_HOME/cmsgemos/setup/paths.sh
 fi
 
+# Export project
+export GEM_PLOTTING_PROJECT=$BUILD_HOME/gem-plotting-tools
+
+# Setup Path
+export PATH=$PATH:$GEM_PLOTTING_PROJECT
+export PATH=$PATH:$GEM_PLOTTING_PROJECT/macros
+export PATH=$PATH:$GEM_PLOTTING_PROJECT/setup
```
jsturdy: why is this in the `$PATH`?

setup/paths.sh
```diff
@@ -37,38 +37,27 @@ else
...
+
+# Setup PYTHONPATH
+export PYTHONPATH=$PYTHONPATH:$BUILD_HOME/vfatqc-python-scripts
+export PYTHONPATH=$PYTHONPATH:$GEM_PLOTTING_PROJECT
+export PYTHONPATH=$PYTHONPATH:$GEM_PLOTTING_PROJECT/macros
+export PYTHONPATH=$PYTHONPATH:$GEM_PLOTTING_PROJECT/setup
```
jsturdy: Why is this in the `$PYTHONPATH`?

setup/paths.sh
```diff
@@ -37,38 +37,27 @@ else
     source $BUILD_HOME/cmsgemos/setup/paths.sh
 fi
 
+# Export project
+export GEM_PLOTTING_PROJECT=$BUILD_HOME/gem-plotting-tools
```
jsturdy: Is `$GEM_PLOTTING_PROJECT` ever used outside of this script?

setup/paths.sh
```diff
@@ -37,38 +37,27 @@ else
     source $BUILD_HOME/cmsgemos/setup/paths.sh
 fi
```
jsturdy: `cmsgemos` should not be set up from this project

setup/paths.sh
jsturdy: OK, I see in my reverse review that it is... though possibly only the once? As such, would suggest a more appropriate variable name (seems only used in relation to channel mapping)

setup/paths.sh
bdorney: I don't see why it shouldn't be?

`cmsgemos` is required since `ana_scans.py` makes use of `runCommand`, and `envCheck` from `gempython` inside of cmsgemos

setup/paths.sh
bdorney: I think this variable name is fine, it matches the convention declared in:

- `cmsgemos`
- `vfatqc-python-scripts`

And allows for easy reference in future development.

setup/paths.sh
bdorney: mmhmm, yes I guess it could be removed.

setup/paths.sh
bdorney: contents of macros and setup are used in the various python scripts here.

e.g. `fitScanData.py` (in $GEM_PLOTTING_PROJECT/macros) is needed by anaUltra*.py routines and channel map info (in $GEM_PLOTTING_PROJECT/setup) is needed by anaUltra*.py

setup/paths.sh
jsturdy: I have the same problem with the name in `vfatqc-python-scripts`, and in `cmsgemos` it has a different purpose

setup/paths.sh
jsturdy: Yes, they depend, but the point is that it should *already* be set up elsewhere, this script should simply tell the user that they have not done other things in life correctly.  The same argument is why I've requested changes in `vfatqc-python-scripts`

setup/paths.sh
```diff
@@ -37,38 +37,25 @@ else
...
 
+# Export project
+export GEM_PLOTTING_PROJECT=$BUILD_HOME/gem-plotting-tools
+
+# Setup Path
+export PATH=$PATH:$GEM_PLOTTING_PROJECT
+
+# Setup PYTHONPATH
+export PYTHONPATH=$PYTHONPATH:$BUILD_HOME/vfatqc-python-scripts
```
jsturdy: As elsewhere, if this should be in the `$PYTHONPATH`, it should be set up that way in `vfatqc-python-scripts`

setup/paths.sh
jsturdy: Then I suppose it's time that project be organized more like a proper `python` package.
* `macros` has a nebulously defined name, but should be reserved for a piece of code you're actually going to run a la: `python mymacro.py`; if the module is being imported or loaded elsewhere, it's better to structure the package like a proper library, i.e., 
 ```
gem-plotting-tools/
  - main program files (probably expected to be executable and in the $PATH, could also go in a bin subdirectory)
  macros/
    - tools you may use for various tasks on their own
  fitting/
   - modules related to fitting
  mapping/
   - modules related to mapping
  tests/
    - any unit tests
 ```
* This way you can `import mapping.PanChannelMaps` (or `from mapping.PanChannelMaps import *`, or however you are using them) with only the `gem-plotting-tools` directory having been added to the `$PYTHONPATH`

setup/paths.sh
bdorney: Yes okay, I understand.  However, when requesting changes in the future it would be more useful to understand the "why" that is attached to a request.  I think this would provide a greater understanding and allow developers to have a higher change of doing things correctly "the first time"

setup/paths.sh
bdorney: Could you describe the purpose then?  I don't see why this is not a good idea, moreover this variable is now checked when setting up the `vfatqc-python-scripts` repo:

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/master/setup/paths.sh#L30

Based on the comments above it looks like the implementation is following the:

>  it should already be set up elsewhere, this script should simply tell the user that they have not done other things in life correctly

So this variable again seems necessary.

setup/paths.sh
bdorney: Sure we can implement these changes.  This is a good idea and will increase readability/maintenance in the long turn.

setup/paths.sh
jsturdy: In `cmsgemos` it is used mainly in the RPM building and `make` processes.
What code in `vfatqc-python-scripts` actually uses this variable, or depends on it being set (or is it simply a way to see if the software is present)?

bdorney: `setup/paths.sh` in `vfatqc-python-scripts` is checking that `gem-plotting-tools` has been setup correctly since `trimChamber.py` requires `gem-plotting-tools/macros/fitScanData.py` (which will be moved to `gem-plotting-tools/fitting/fitScanData.py` as requested below).

bdorney:
Unit tests successfully run using `ana_scans.py` with:
- `--anaType=latency`
- `--anaType=scurve`
- `--anaType=threshold` without `--chConfigKnown` option
- `--anaType=trim`
- `--anaType=threshold` with `--chConfigKnown` option

Looks good.

macros/plot_scurves_by_thresh.py
```diff
@@ -1,4 +1,4 @@
-from plot_scurve import *
+from macros.plot_scurve import *
```
jsturdy: This script imports `plot_scurve` but doesn't use it, probably remove the `import`?

anaUltraScurve.py
```diff
@@ -54,13 +54,12 @@
         pass
     pass
 
-buildHome = os.environ.get('BUILD_HOME')
-
+projectHome = os.environ.get('GEM_PLOTTING_PROJECT')
```
jsturdy: Quick question, is there a check that is run before this point that this variable is set (and points somewhere real)?
If not, add an issue such that checks like this are made in the future

macros/plot_vfat_and_channel_Scurve.py
```diff
@@ -1,4 +1,4 @@
-from plot_scurve import *
+from macros.plot_scurve import *
```
jsturdy: This script seems to only add options on top of the `plot_scurve` script.  If that's all there is, suggest to merge `plot_vfat_and_channel_Scurve.py` and `plot_scurve.py`, making it similar to the other plotting macros, and just make it standalone (though my quick glance at the `macros` may have missed something in how this is used).

macros/plot_scurves_by_thresh.py
bdorney: Fixed; and added the imports I assumed it was trying to take from plot_scurve.py (ROOT and OptionsParser).

jsturdy: Did you forget to push? Last commit to github was 20 hours ago...

bdorney: Yeah forgot to push

anaUltraScurve.py
bdorney: Checks made using `envCheck()` in related scripts:

- `anaUltraScurve.py`
- `anaUltraThreshold.py`
- `mapping/buildMapFiles.py`

macros/plot_vfat_and_channel_Scurve.py
bdorney: Yeah not really sure how @cbravo135 intended the tools in /macros to be used but I can have a look and try to merge appropriately.

macros/plot_scurves_by_thresh.py
```diff
@@ -1,25 +1,17 @@
...
 (options, args) = parser.parse_args()
 
 filename = options.filename
 overlay_fit = options.overlay_fit
-channel_yes = options.channel_yes
+channel_yes = options.channels
 vfat = options.vfat
 strip = options.strip
 
+import root as r
```
jsturdy: Think L14 should be:
```python
import ROOT as r
```

bdorney: Whoops...

PR #14, opened 2017-08-10
Address issue #10

lmoureaux:
In addition, add channel maps to `.gitignore` since they are automatically generated.


lmoureaux:
Closing because of  #13

Issue #15, opened 2017-08-15
Feature Request: fitScanData.py Takes Number of Injections as Input


bdorney:
## Brief summary of issue
Right now the form of the modified erf function that is used to fit s-curves has parameters that are hardcoded.  These parameters should be changed from `500` to `0.5 * nInjections`.  See [L#34](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/22cdd9bf15ebe3f8df652564d51cd943af8f79b6/fitting/fitScanData.py#L34) of fitScanData.py

However now the `--nevt` option can be passed from command line to the scan tool of interest.  Right now the `--nevt` defaults to 1000 so there is no conflict.  However if the user supplies a different value the fit may not work as planned.

This was also mentioned in:
https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/113

### Types of issue

- [ ] Bug report (report an issue with the code)
- [X] Feature request (request for change which adds functionality)

## Expected Behavior

In [L#1](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/22cdd9bf15ebe3f8df652564d51cd943af8f79b6/fitting/fitScanData.py#L1) `fitScanData()` should take an additional parameter as input: `nInjections=1000`.  The default will ensure the same behavior as present.

Then the equation of Fit at [L#34](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/22cdd9bf15ebe3f8df652564d51cd943af8f79b6/fitting/fitScanData.py#L34) should be modified from:

```
    fitTF1 = r.TF1('myERF','500*TMath::Erf((TMath::Max([2],x)-[0])/(TMath::Sqrt(2)*[1]))+500',1,253)
```

to be instead:

```
    fitTF1 = r.TF1('myERF','%i*TMath::Erf((TMath::Max([2],x)-[0])/(TMath::Sqrt(2)*[1]))+%i'%(0.5*nInjections),1,253)
```

## Current Behavior

Currently if the default value of `--nevt=1000` but if the user specifies a value outside of this the modified erf applied to the scurve data may not work as intended

## Context (for feature requests)

Ensuring this ensures correct fitting when number of events in scurve/trimming differs from default

PR #21, opened 2017-08-23
Read the event count from the SCurve data tree

lmoureaux:
Addesses issue #15 by reading the number of events from the S-curve trees.

## Description
The S-curve fit function has the number of events hard-coded to 1000, but can be changed in `ultraScurve`. Detect it from the S-curve tree instead, to enable using data produced with non-default `--nevts`.

The implementation is not as suggested in #15, but will rather read the number of events from the tree. It's IMHO less error-prone: there is no need for an additional parameter to `fitScanData` (and flag to `anaUltraScurve.py`), that could be forgotten. The fit function will "just work" unless supplied with inconsistent trees (for sample merged using `hadd`), in which case an error may even be thrown (depending on the parameters used to generate the `hadd`ed trees).

### Types of changes

## Motivation and Context

See #15.

## How Has This Been Tested?

The changes were tested using data from http://cmsonline.cern.ch/cms-elog/1005623. Cross-checks:

* The number of failed fits doesn't change (much) with the number of events. The scan at 10000 events fails to fit weird channels with no data on VFAT 9.
* The average `chi^2` of fits doesn't change linearly (as I would expect if the normalization stayed fixed) with the number of events, but is `~sqrt(nevt)` (as I would expect from statistical fluctuations):

| `nevts` | `<chi^2>` | 
|--------:|---------:|
| 10000 | 114 |
| 1000 | 25 |
| 100 | 13 |
| 10 | 10 |

## Checklist:

### DIFF
--- a/fitting/fitScanData.py
+++ b/fitting/fitScanData.py
@@ -32,12 +32,17 @@ class ScanDataFitter(DeadChannelFinder):
                 self.scanCount[vfat][ch] = 0
 
         self.fitValid = [ np.zeros(128, dtype=bool) for vfat in range(24) ]
+        self.Nev = -1
 
     def feed(self, event):
         super(ScanDataFitter, self).feed(event)
         self.scanHistos[event.vfatN][event.vfatCH].Fill(event.vcal,event.Nhits)
         if(event.vcal > 250):
             self.scanCount[event.vfatN][event.vfatCH] += event.Nhits
+        if self.Nev < 0:
+            self.Nev = event.Nev
+        else:
+            assert self.Nev == event.Nev, 'Inconsistent S-curve tree'
 
     def readFile(self, treeFileName):
         inF = r.TFile(treeFileName)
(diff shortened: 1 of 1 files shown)


jsturdy:
Is this ready for review (it's a lack of functionality in github that doesn't allow the person creating the PR to request review...)?

lmoureaux:
@jsturdy It's ready for review.

bdorney:
This issue is also pertinent to the conversation: https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/113

fitting/fitScanData.py
```diff
@@ -50,7 +55,8 @@ def fit(self):
 
         random = r.TRandom3()
         random.SetSeed(0)
-        fitTF1 = r.TF1('myERF','500*TMath::Erf((TMath::Max([2],x)-[0])/(TMath::Sqrt(2)*[1]))+500',1,253)
+        fitTF1 = r.TF1('myERF','[3]*TMath::Erf((TMath::Max([2],x)-[0])/(TMath::Sqrt(2)*[1]))+[3]',1,253)
+        fitTF1.FixParameter(3, self.Nev / 2.)
```
bdorney: I would prefer we did not add to the list of fit parameters of this TF1.  Please instead use:

```
fitTF1 = r.TF1("myERF","%i*TMath::Erf((TMath::Max([2],x)-[0])/(TMath::Sqrt(2)*[1]))+%i"%(self.Nev / 2, self.Nev / 2.),1,253)
```

And remove line 59.

lmoureaux:
Using string manipulation routines to fix parameters, added the fit summaries to the Elog.

Issue #26, opened 2017-08-23
Feature Requests: Unit Tests


bdorney:
## Brief summary of issue

In the past week we have let PR's be merged without either proper testing or proper oversight (myself included) as demonstrated by:

https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/16 
https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/24

Which were either not properly tested originally, or not re-tested after committing on top of a request.  To prevent lost time, and most importantly p5 machine being updated with bad code we should finalize a procedure and develop unit tests which should cross-check things.  I propose:

- Each PR should be supported with an e-log link that shows tests were properly done.  
- New commits after changes are requested should include an e-log link where tests have been repeated and properly passed.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [X] Feature request (request for change which adds functionality)

## Expected Behavior

I would propose the following:
1. For changes to `ana_scans.py` for each `--anaType` a re-analysis of old data should be launched
2. For changes to `anaUltra*.py` scripts see the first item,
3. For changes to `anaXDAQ*.py` scripts see the first item,
4. For changes to files in `fitting/` or `mapping/` sub-directories taking a run of `vfatqc-python-scripts/ultraScurve.py` and then analyzing the data immediately

For items 1-3 something like `scripts/unitTests_Ana_Ultra.sh` script like:

```
#!/bin/bash

# record branch name and commit hash
cd $GEM_PLOTTING_PROJECT
git status 2>&1 | tee -a unit_test_log.txt
git show 2>&1 | tee -a unit_test_log.txt

# perform tests
ana_scans.py --scandate=current --anaType=latency  2>&1 | tee -a unit_test_log.txt
sleep 2
ana_scans.py --scandate=current --anaType=scurve  2>&1 | tee -a unit_test_log.txt
sleep 2
ana_scans.py --scandate=current --anaType=threshold  2>&1 | tee -a unit_test_log.txt
sleep 2 
ana_scans.py --scandate=current --anaType=trim  2>&1 | tee -a unit_test_log.txt

# Done
echo "Unit Tests Completed, Upload:"
echo $PWD/unit_test_log.txt
echo "To the E-Log or Github PR along with the images in:"
echo $ELOG_PATH
```

With a similar script e.g. `scripts/unitTests_Ana_XDAQ.sh`

[...]

Issue #24, opened 2017-08-23
Fitting is broken


lmoureaux:
Fitting was broken in commit bf5d7f2.

## Brief summary of issue

```
[...]
  File "/afs/cern.ch/user/l/lmoureau/CosmicStandSw/gem-plotting-tools/fitting/fitScanData.py", line 58, in fit
    fitTF1 = r.TF1('myERF','%f*TMath::Erf((TMath::Max([2],x)-[0])/(TMath::Sqrt(2)*[1]))+%f'%(self.Nev/2.),1,253)
TypeError: not all arguments converted during string formatting
```

### Types of issue

- [x] Bug report (report an issue with the code)
- [ ] Feature request (request for change which adds functionality)

## Expected Behavior

No crash

## Current Behavior

Crashes

### Steps to Reproduce

1. Try to fit S-curves using `anaUltraScurve` or `trimChamber`

## Your Environment

* Version used: master
* Shell used: bash

PR #27, opened 2017-08-24
[RFC] Integrate gem-plotting-tools with Travis CI

lmoureaux:
Travis CI is an online platform for continuous integration, free-to-use for open source projects. It is deeply integrated with Github and can block merge just like landscape can. This PR adds support for Travis CI for gem-plotting-tools.

## Description
Travis is very generic and can be told to run arbitrary commands. The "build" fails if any of the commands fails, which is the ideal behavior for unit tests. Travis uses Ubuntu 14.04 as their default environment, and I don't change this.

This PR adds a `.travis.yml` file to enable Travis. This file and the scripts in `.travis` setup the following testing environment:

* A Linux machine based on Ubuntu 14.04 (Travis' default; some cosmetic changes were needed for the code to run there)
* ROOT packages from the ROOT website [edited: was Ubuntu's ROOT packages]
* Python 2.6 and python 2.7 (tests will be run in both environments)
* `gem-plotting-tools`' dependencies (as per `requirements.txt`)
* `cmsgemos` from the default branch [edited: was `master`]
* `cmsgemos`' dependencies (as per `requirements.txt`) [edited: new]

In addition, this PR adds a test of `ana_scans.py --runType=scurve`. The test will catch syntax errors and fatal exceptions in `ana_scans.py` and `anaUltraScurve.py`, as well as any file they import. The test will *not* catch other problems. Some changes to `ana_scans.py` were required in order to make it fail when the subcommands fail.

If this PR gets merged, an administrator can subscribe to Travis and enable it within minutes.

### Types of changes

## Motivation and Context
See #26 for context. Having Travis run every tool for every commit would catch code typos early.

## How Has This Been Tested?
The changes were continuously tested on Travis itself (and only there). See [here](https://travis-ci.org/lmoureaux/gem-plotting-tools/jobs/268001263) for an example of the test running without error, [here](https://travis-ci.org/lmoureaux/gem-plotting-tools/jobs/267992000) for the output of Travis catching bug #24.

### Screenshots
This is the current Travis result for my branch `travis`...
[Build Status](https://travis-ci.org/lmoureaux/gem-plotting-tools)

## Checklist:


lmoureaux:
This PR isn't ready yet, but I'd like to get comments on the architecture before more tests get added to it.

bdorney:
> Ubuntu's ROOT packages

Can you confirm this is:

```
ROOT 5.34/36 (heads/v5-34-00-patches@v5-34-34-76-g57bae4c, Apr 08 2016, 12:05:00 on linuxx8664gcc)
```

> Python 2.6 and python 2.7 (tests will be run in both environments):

Can you confirm these are:

```
Python 2.6.6
Python 2.7.5
```

As these are the defaults on `slc6` and `cc7` operating systems.

> cmsgemos from branch master

Please use `generic-amc-release-v3` of cmsgemos instead, found [here](https://github.com/cms-gem-daq-project/cmsgemos/tree/generic-amc-release-v3).

One question.  How does Travis CI update? e.g. if there is an update to `cmsgemos` what does one do to ensure that this update is included?

Also I think we need to expand this unit testing:

> In addition, this PR adds a test of ana_scans.py --runType=scurve. The test will catch syntax errors and fatal exceptions in ana_scans.py and anaUltraScurve.py, as well as any file they import. The test will not catch other problems. Some changes to ana_scans.py were required in order to make it fail when the subcommands fail.

Let's expand this to include:

- `anaUltraThreshold.py`
- `anaUltraLatency.py`
- `anaXDAQLatency.py`

Also can the input files that we use as the "test data" be stored in the same file structure that `ana_scans.py` expects.  This could also help us test `chamberInfo.py`.

.travis.yml
```diff
@@ -0,0 +1,22 @@
...
+        packages:
+        - root-system
+        - root-plugin-hist-*
+        - libroot-*-dev
+        - libroot-bindings-python5.34
+
+before_script:
+    - .travis/install_cmsgemos.sh
+    # Copy Travis chamber info
+    - cp .travis/chamberInfo.py mapping/chamberInfo.py
```
bdorney: Looking at the file below this will work for testing.  But as a counter question how do you intend to use this to test potential changes to `chamberInfo.py` itself?

For example what happens if someone adds a dictionary to this file or changes the default behavior.  Seems like we'd then have to change `.travis/chamberInfo.py` as well to test this.

I think we should rely on using the default `mapping/chamberInfo.py` or at least one of the other links (e.g. `chamberInfo_904.py` or `chamberInfo_qc8.py` with a preference for the first).

.travis.yml
```diff
@@ -0,0 +1,22 @@
...
+    - source .travis/setupenv.sh
+
+script:
+    - ana_scans.py --scandate=current --anaType=scurve
```
bdorney: As mentioned we should expand to test the full range of scripts.  This here only tests `anaUltraScurve.py` and one piece of `ana_scans.py`.

.travis/data/travis/.gitignore
```diff
@@ -0,0 +1,9 @@
+*
+
+# Don't ignore this file
+!.gitignore
+
+# Don't ignore data files
+!scurve
+!scurve/current
+!scurve/current/SCurveData.root
```
bdorney: Why is a duplicate `.gitignore` file needed at this location: `.travis/data/travis/`?

.travis/install_cmsgemos.sh
```diff
@@ -0,0 +1,6 @@
+#! /usr/bin/env bash
+
+# Travis script to install cmsgemos
+
+cd $TRAVIS_BUILD_DIR/..
+git clone --depth 1 https://github.com/cms-gem-daq-project/cmsgemos.git
```
bdorney: I see...so everytime someone makes a PR this is going to clone `cmsgemos`.  So bugs in `cmsgemos` might cause a unit test to this repo to fail?  One needs to then keep this in mind when interpreting results.

.travis/setupenv.sh
```diff
@@ -0,0 +1,22 @@
+#! /usr/bin/env bash
+
+# Script to setup data paths
+
+# For Ubuntu 14.04: add pyroot to PYTHON_PATH
+export PYTHONPATH="/usr/share/python-support/root:/usr/lib/python2.7/dist-packages/"
```
bdorney: Let's test on `python 2.6.6` for now.

anaUltraLatency.py
```diff
@@ -1,4 +1,4 @@
-#!/bin/env python
+#! /usr/bin/env python
```
bdorney: What is the rationale for this change (and in subsequent files: `anaUltraScurve.py`, `anaUltraThreshold.py`, `anaXDAQLatency.py`, `ana_scans.py`, and `anautilities.py`)?

Are you sure this change, while possibly motivated for making `Travis CI` work, will not impact performance on production machines? Specifically the P5 machines?

ana_scans.py
```diff
@@ -125,15 +125,22 @@ def launchAnaArgs(anaType, cName, cType, scandate, scandatetrim=None, ztrim=4.0,
   try:
     log = file("%s/anaLog.log"%(dirPath),"w")
  
-    #runCommand(cmd,log)
-    runCommand(cmd)
+    #returncode = runCommand(cmd,log)
+    returncode = runCommand(cmd)
+    if returncode != 0:
+      print "Error: command exited with non-zero code %d" % returncode
+      return returncode
```
bdorney: What's the reason behind this?  This block is already inside a `try:` block so any error results in the code executing the `except` portion and issues an error message already.

Please revert.

ana_scans.py
```diff
@@ -217,13 +224,20 @@ def launchAnaArgs(anaType, cName, cType, scandate, scandatetrim=None, ztrim=4.0,
       print("Normal termination")
       pool.close()
       pool.join()
+      print "Results:", res.get()
+      for returncode in res.get():
+        if returncode != 0:
+          sys.exit(returncode)
+        pass
```
bdorney: Again could you describe the rationale here?  We're inside a `try` block already?

Also what are the possible `returncode` values?

lmoureaux:
>> Ubuntu's ROOT packages
>
> Can you confirm this is:
> ```
> ROOT 5.34/36 (heads/v5-34-00-patches@v5-34-34-76-g57bae4c, Apr 08 2016, 12:05:00 on linuxx8664gcc)
> ```

```
amd64 root-system-bin amd64 5.34.14-1build1
```

Same architecture, major and minor version. Patch number is different, though. I don't know of a simple way to install a specific ROOT version on Debian (and I don't want to compile it).

>> Can you confirm these are:
>
> Python 2.6.6
> Python 2.7.5

Python 2.6.13 and 2.7.13 (more recent patch versions, no behavior change expected). I can look into specifying patch numbers, but I'm not sure whether Travis still supports 2.6.6 (since even 2.6.13 is deprecated as it's not supported by the Python team anymore).

>> cmsgemos from branch master
>
> Please use generic-amc-release-v3 of cmsgemos instead, found here.

Checked, it uses the default branch, which is generic-amc-release-v3 for now. Should I specify the branch explicitly?

> One question. How does Travis CI update? e.g. if there is an update to cmsgemos what does one do to ensure that this update is included?

The latest `cmsgemos` if fetched from git every time the test is run.

.travis/data/travis/.gitignore
lmoureaux: I didn't want to add many long paths to the main `.gitignore`. I chose to move all data-related ignore directives into another file.

bdorney:
> Same architecture, major and minor version. Patch number is different, though. I don't know of a simple way to install a specific ROOT version on Debian (and I don't want to compile it).

Compiled sources are available, for example for `v5.34.36`:
https://root.cern.ch/content/release-53436

> Checked, it uses the default branch, which is generic-amc-release-v3 for now. Should I specify the branch explicitly?

No let's rely on the correct setup of the `cmsgemos` repo to point git to this branch as default.  This can only be changed by an admin of the `cmsgemos` repo which I think includes only @jsturdy presently.

> The latest cmsgemos if fetched from git every time the test is run.

Yeah I saw after looking at the code.

.travis/data/travis/.gitignore
bdorney: I think this will be harder to maintain in the long run...now developers have to check 2 files instead of one.  And a new developer may not even know that a second `.gitignore` exists.  Let's revert this and use only one file.

jsturdy:
Sorry to not jump on this earlier, but it would be better if you can set it up to work with the docker container I set up for `cmsgemos` CI (work in progress, and once it's sorted, the plan was to simply plug in the other packages)
* Here's the `travis.yml` I've been working with: https://github.com/jsturdy/cmsgemos/blob/jsturdy-travis-int-test/.travis.yml (I also have templates in `vfatqc-python-scripts`
* Here's the docker I currently test against: https://gitlab.cern.ch/sturdy/gemdaq_ci_worker

.travis/setupenv.sh
lmoureaux: There is no distribution of `pyroot` built against Python 2.6 in Debian, but it seems binary compatible.

.travis/install_cmsgemos.sh
lmoureaux: Right. We could specify a specific revision (or tag?), but would have to update it manually.

anaUltraLatency.py
lmoureaux: Debian/Ubuntu don't have `/bin/env` in any supported version. Performance won't be affected at all, `env` is just a tool that searches `$PATH` and standard locations for an executable and launches it. The only possible problem would be `SL` to remove `/usr/bin/env` in a future version, but IIRC there is some standard for Linux distros that says `/usr/bin` should be the default for executables.

ana_scans.py
lmoureaux: Non-zero return values and exceptions aren't the same thing.

ana_scans.py
lmoureaux: `returncode` is the return value from the called command, and will be non-zero if it fails (Python returns non-zero on exceptions, SIGSEGV, SIGKILL, ...). For this to work, we should ensure that our tools also return non-zero on "soft" failures (i.e. no crash), but that's a different issue.

.travis.yml
lmoureaux: Your PR for `chamberInfo.py` isn't merged yet, so mine isn't built on top of it :-) The current default is for P5, meaning 12x more data in the repo, which I don't want. My plan is to move this file to `mapping/chamberInfo_Travis.py` when merging your changes.

Also, the default for 904 includes detector names, which could change at some point. I don't want the build to break just because we switched detectors, which is why I used a custom version. We can in principle mix data from different detectors in the repo (as long as they don't have different `vfatmask`s, which currently limits the number of available options).

IMHO updates to the data files should be limited to file format changes.

ana_scans.py
bdorney: Sure but I think this should be a separate PR. Could you revert here and open a new issue that is related to this? That would allow us to be a bit more organized and focused in our PR's

ana_scans.py
bdorney: As above let's do this in a separate PR with a new issue

.travis.yml
bdorney: I see your points.  Let's not build off unnerved PR's your right. And let's not rely on potentially transient setups.

However the detectors in P5 will not changed (even post LS2 there will be a GEMINIm27L1). And it is for the production level machines that we should be concerned with. I think ensuring the integrity tests are valid for the real system is the primary goal so let's at least have the P5 system and the /current directory for each. This would be the most representative case.

.travis/install_cmsgemos.sh
bdorney: No let's keep it this way and rely the HEAD of cmsgemos being correct

.travis/setupenv.sh
bdorney: Again check @jsturdy's implementation to see if this sort of issue is resolved

.travis.yml
lmoureaux: Using P5 data would also mean that PRs that depend on file format changes would have to be tested on P5 data (or fake P5 data :-1:, or we'd have to merge with tests failing :-1:). Generating this data would slow down development a lot (it could even deadlock in case of circular dependencies).

lmoureaux:
In today's update:

* Using ROOT code downloaded directly from the ROOT website
* Install `cmsgemos`' Python dependencies (none atm)
* Cleanup `.gitignore`s as requested
* The test script now checks that all output files are present and non-empty
* Ensures that tests are run from `$DATA_PATH`
* Changes to `ana_scans.py` were moved to PR #29. The code is kept here because it's needed for this branch to work properly. I will revert just before merging `master` once PR #29 gets merged. 

I considered using @jsturdy's Docker image, but I [couldn't even run `ls`](https://travis-ci.org/lmoureaux/gem-plotting-tools/builds/268343236) from inside the container. As I don't have the expertise to fix it (both on Docker and on how to setup `cmsgemos`), I went ahead with Ubuntu as the second-best solution. The tests themselves will be easily reusable on Docker anyway.

I considered using `tox`, but it looked like overkill (and overlapping with Travis). I considered using Python's own `unittest` module but IMHO it's a lot of code for little added value.

Still not addressed:
* `/bin/env` -> `/usr/bin/env` (?)
* Using the exact same Python dot versions as in prod (not supported by Travis, I don't think it's worth the effort given that the Docker image will provide a well-controlled environment)
* Use P5 data (unsure, see discussion above)
* Test other scripts than `anaUltraScurve.py` (will start working on it once the principle is :+1:'ed, we might as well have on PR/tested script)

jsturdy:
I tracked down a couple of issues with the docker container that was being used, and have managed to get things looking pretty much as expected from within the `travis` environment.  If you can let me know what you would need to add to the container to complete your tests (unless you just want to run the installation commands in your `travis` script), I can update the container.

* In the `docker` image you can just use `yum install root\*` to get the version of `root` provided with `slc6`/`cc7`
* Extra `python` dependencies should be handled *only* through a `requirements.txt` or a `requirements-dev/prod/test.txt` and `pip`
* It would be best to have the test data and directories set up in the container itself, or, at the very least, pulled via `wget` into the `travis` build, rather than stored in the repo
* While adding the test script (there should really be a single test driver to test all aspects of the package, which was why I had suggested `tox`) it would be good to add the test coverage (`coveralls`, `codecov`, etc., you can also look at my [integrations](https://github.com/jsturdy/gem-plotting-tools/tree/jsturdy-integrations) branch for some testing I had been setting up, though not yet operational)

One other thing that would be optimal would be to separate the unit testing from the CI/CD setup, such that the unit tests can be run independently of the CI/CD, and will then easily plug into a functional CI/CD framework (there will be two, by virtue of the fact that the `gitlab` mirrors are using `gitlab-ci`, while the `github` repositories will use `travis`)

While moving along these lines, the best way forward will be for you to regularly do a `git rebase` to the mainline branch, which should be straightforward as long as you're not modifying any of the code unrelated to the `travis` aspects (which this PR shouldn't be doing anyway...)

jsturdy:
closing as "On Hold"

Issue #28, opened 2017-08-25
ana_scans.py doesn't fail if one of the subcommand fails


lmoureaux:
## Brief summary of issue

`ana_scans.py` invokes other commands such as `anaUltraScurve.py`. However, `ana_scans.py` doesn't notice when one of those commands fails, and will happily keep running. `ana_scans.py` will exit with status code zero even if somewhat went horribly wrong. It should instead follow the usual conventions and return a non-zero status code.

### Types of issue
- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

## Expected Behavior
If one of the subcommands fails, I expect `ana_scans.py` to at least report a non-zero status code.

## Current Behavior
The status code is always zero (except maybe when killed).

## Context (for feature requests)
Was trying to use `ana_scans.py` to setup unit tests on Travis (see #27).

## Your Environment
* Version used: master, master prior to bug #24 got fixed
* Python version: 2.6.x, 2.7.x
* More details in #27

PR #23, opened 2017-08-23
Addressing issue #22 - Symlinks for mapping/chamberInfo.py

bdorney:
## Description

Addressing: https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/22

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Motivation and Context

Quality of life improvement.

## How Has This Been Tested?

Tested: http://cmsonline.cern.ch/cms-elog/1005667

## Checklist:

- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

PR #29, opened 2017-08-25
Let ana_scans fail when any command fails

lmoureaux:
## Description
Addresses #28 

### Types of changes

## Motivation and Context

Needed for unit testing, see #27 and #28

## How Has This Been Tested?

Tested on Travis with `--anaType=scurve`. Worked fine both in case of success and in case of failure of the main analysis (output logs [here](https://travis-ci.org/lmoureaux/gem-plotting-tools/jobs/268001263) (success), and [here](https://travis-ci.org/lmoureaux/gem-plotting-tools/jobs/267992000) (analysis fails)).

## Checklist:

- [x] I have added tests to cover my changes (not in this PR though).
- [x] All new and existing tests passed (not in this PR though).

### DIFF
--- a/ana_scans.py
+++ b/ana_scans.py
@@ -125,15 +125,22 @@ def launchAnaArgs(anaType, cName, cType, scandate, scandatetrim=None, ztrim=4.0,
   try:
     log = file("%s/anaLog.log"%(dirPath),"w")
  
-    #runCommand(cmd,log)
-    runCommand(cmd)
+    #returncode = runCommand(cmd,log)
+    returncode = runCommand(cmd)
+    if returncode != 0:
+      print "Error: command exited with non-zero code %d" % returncode
+      return returncode
     for item in postCmds:
-      runCommand(item)
+      returncode = runCommand(item)
+      if returncode != 0:
+        print "Error: command exited with non-zero code %d" % returncode
+        return returncode
       pass
   except CalledProcessError as e:
     print "Caught exception",e
+    return -1
     pass
-  return
+  return 0
 
 if __name__ == '__main__':
   import sys,os,signal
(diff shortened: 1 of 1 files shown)


ana_scans.py
```diff
@@ -125,15 +125,22 @@ def launchAnaArgs(anaType, cName, cType, scandate, scandatetrim=None, ztrim=4.0,
   try:
     log = file("%s/anaLog.log"%(dirPath),"w")
  
-    #runCommand(cmd,log)
-    runCommand(cmd)
+    #returncode = runCommand(cmd,log)
+    returncode = runCommand(cmd)
```
jsturdy: Why did the logging get removed?

ana_scans.py
```diff
@@ -217,13 +224,20 @@ def launchAnaArgs(anaType, cName, cType, scandate, scandatetrim=None, ztrim=4.0,
...
       pool.close()
       pool.join()
+      print "Results:", res.get()
+      for returncode in res.get():
+        if returncode != 0:
+          sys.exit(returncode)
+        pass
     except KeyboardInterrupt:
       print("Caught KeyboardInterrupt, terminating workers")
       pool.terminate()
```
jsturdy: Why no `sys.exit(-1)` here?  Or, why present in the other `except` blocks?

ana_scans.py
lmoureaux: Wasn't me ;-) See bug #6

ana_scans.py
lmoureaux: I used `sys.exit(-1)` to indicate an exception in `ana_scans.py`. When `ana_scans.py` works properly, the exit code of (the first failing) subcommands is just forwarded to the caller.

ana_scans.py
jsturdy: OK, I guess I'll force @dorney to fix it in #23 since he introduced it originally

ana_scans.py
jsturdy: OK, so you're not interested in a `-1` when you get a `KeyboardInterrupt`?

lmoureaux: Not for unit tests, but that's inconsistent behavior

ana_scans.py
bdorney: Addressed in 311371684a8de283af669bf83ed0139eb4e76af5

lmoureaux:
No, as I wouldn't be able to test it.

PR #52, opened 2017-10-23
Updating TBranch Names for Analysis Scripts

bdorney:
## Description

Accounts for changes in branche names:

- `lat` -> `latency`
- `vthr` -> `vth1`

as described here: https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/144

### Types of changes

## Motivation and Context

See discussion: https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/144

## How Has This Been Tested?

Taking new data and analyzed it, see screenshots: http://cmsonline.cern.ch/cms-elog/1017334

## Checklist:

### DIFF
--- a/anaUltraLatency.py
+++ b/anaUltraLatency.py
@@ -50,12 +50,12 @@ latMin = 1000
 latMax = -1
 nTrig = -1
 for event in inF.latTree:
-    dict_hVFATHitsVsLat[int(event.vfatN)].Fill(event.lat,event.Nhits)
-    if event.lat < latMin and event.Nhits > 0:
-        latMin = event.lat
+    dict_hVFATHitsVsLat[int(event.vfatN)].Fill(event.latency,event.Nhits)
+    if event.latency < latMin and event.Nhits > 0:
+        latMin = event.latency
         pass
-    elif event.lat > latMax:
-        latMax = event.lat
+    elif event.latency > latMax:
+        latMax = event.latency
         pass
 
     if nTrig < 0:
--- a/anaUltraScurve.py
+++ b/anaUltraScurve.py
@@ -105,8 +105,8 @@ if options.SaveFile:
     myT.Branch( 'panPin', panPin, 'panPin/I' )
     trimRange = array( 'i', [ 0 ] )
     myT.Branch( 'trimRange', trimRange, 'trimRange/I' )
-    vthr = array( 'i', [ 0 ] )
-    myT.Branch( 'vthr', vthr, 'vthr/I' )
+    vth1 = array( 'i', [ 0 ] )
+    myT.Branch( 'vth1', vth1, 'vth1/I' )
     trimDAC = array( 'i', [ 0 ] )
     myT.Branch( 'trimDAC', trimDAC, 'trimDAC/I' )
     threshold = array( 'f', [ 0 ] )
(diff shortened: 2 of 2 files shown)


bdorney:
> I think in some places (`ultraThreshold`) `vth1` and `vth2` are set separately and `vthr` is `vth2`-`vth1`, as it is in VFAT2. Granted this is different from VFAT3, but consistency can still be achieved, can you check that this change doesn't introduce any semantic discrepancies?

Yeah I had a similar thought as well and was also concerned so I checked when I was designing this.  In the master branch we previously had:

- `vth1` is `VThreshold1` in `ultraThreshold.py` and is `VThreshold1` in `ultraLatency.py`
- `vth2` is `VThreshold2` in `ultraThreshold.py` but never used; and is `VThreshold2 in `ultraLatency.py`
- `vthr` (is `vth1` in [ultraScurve.py](https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/699d4286180315111fdfb0e54821b29c2c4b30e1/ultraScurve.py#L162))
- `vth` this is `vt1 - vt2` in [ultraThreshold.py](https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/699d4286180315111fdfb0e54821b29c2c4b30e1/ultraThreshold.py#L120).

So this change does not cause a systematic discrepancy and reduces complexity/ambiguity with regard to our branches for `VThreshold1`, `VThreshold2` and their difference.

jsturdy:
So to summarize:
* All trees currently have branches for `VThreshold1/2` (VFAT2 specific) 
* `vthr` is proposed to change to be `VThreshold1` which is:
  * incorrect (but accidentally fine) for V2 (though @cbravo135 likes to mention "the chip designers gave us this handle, we shouldn't code it out of utility")
  * correct for V3 (modulo the change in name of the register)

bdorney:
I don't follow what you mean by:

> vthr is proposed to change to be VThreshold1 which is:

Right now in the master branch the TBranch `vthr` is `VThreshold1`.  My proposal is to reduce ambiguity by stopping the usage of different names for the same quantity.  Maybe we are talking about the same thing and I haven't understood.

For handling the case of `V2b` and `V3` I would propose that the TBranch names have the same names but a table showing the correspondence be added to both the `vfatqc-python-scripts` repo and the `gem-plotting-tools` repo.  For example see:

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/145

jsturdy:
What I'm trying to say is that `vthr` is used to mean the "threshold" applied to the chip as a whole, in all places.
However, in `ultraThreshold` it is done "correctly", while in `ultraScurve` (maybe others) it is done incorrectly, but it is accidentally correct (since `VT2` is currently, always set to 0).  If the value written into the `tree` for `vthr` is changed to `VT1` alone, then it becomes "incorrect" (for `VFAT2`)

I definitely agree, `vthr` *should* be the "threshold" applied to the chip as a whole (whatever that means, but semantically consistent), so I would propose that it be changed (everywhere) as:
* in `VFAT2` `VT2-VT1` (has the downside that the value is now negative, while it was in the past positive in some places, but has the upside that it is the correct representation of the circuit)
* in `VFAT3` (whatever the new register is...)

PR #66, opened 2018-02-07
Adding threshold scan control to the analysis

jsturdy:
## Description
Updates the analysis control to allow analysis of various different types of `VT1` scans

### Types of changes

## Motivation and Context
The `ultraThreshold` module can be configured to take:
* VT1 scan vs channel using tracking data
* VT1 scan vs VFAT using tracking data
* VT1 scan vs VFAT using trigger data (s-bits), which allows much faster scanning of the high threshold tails

## How Has This Been Tested?
* http://cmsonline.cern.ch/cms-elog/1026783
* New configurations are being generated using analysis of both per-channel and per VFAT w/ trigger data

### DIFF
--- a/anaInfo.py
+++ b/anaInfo.py
@@ -3,7 +3,9 @@ import string
 ana_config = {
         "latency":"anaUltraLatency.py",
         "scurve":"anaUltraScurve.py",
-        "threshold":"anaUltraThreshold.py",
+        "thresholdch":"anaUltraThreshold.py",
+        "thresholdvftrk":"anaUltraThreshold.py",
+        "thresholdvftrig":"anaUltraThreshold.py",
         "trim":"anaUltraScurve.py"
         }
 
@@ -17,6 +19,9 @@ tree_names = {
         "scurve":("SCurveData.root","scurveTree"),
         "scurveAna":("SCurveData/SCurveFitData.root","scurveFitTree"),
         "threshold":("ThresholdScanData.root","thrTree"),
+        "thresholdch":("ThresholdScanData.root","thrTree"),
+        "thresholdvftrig":("ThresholdScanData.root","thrTree"),
+        "thresholdvftrk":("ThresholdScanData.root","thrTree"),
         "thresholdAna":("ThresholdScanData/ThresholdPlots.root","thrAnaTree"),
         "trim":("SCurveData_Trimmed.root","scurveTree"),
         "trimAna":("SCurveData_Trimmed/SCurveFitData.root","scurveFitTree")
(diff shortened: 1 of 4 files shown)


anaInfo.py
```diff
@@ -17,6 +19,9 @@
         "scurve":("SCurveData.root","scurveTree"),
         "scurveAna":("SCurveData/SCurveFitData.root","scurveFitTree"),
         "threshold":("ThresholdScanData.root","thrTree"),
```
bdorney: I think you forgot to remove line 21.

ana_scans.py
```diff
@@ -163,7 +163,7 @@ def launchAnaArgs(anaType, cName, cType, scandate, scandatetrim=None, ztrim=4.0,
   from anaoptions import parser
 
   parser.add_option("--anaType", type="string", dest="anaType",
-                    help="Analysis type to be executed, from list {'latency','scurve','threshold','trim'}", metavar="anaType")
+                    help="Analysis type to be executed, from list {'latency','scurve','thresholdch','thresholdvftrig','thresholdvftrk','trim'}", metavar="anaType")
```
bdorney: Actually instead of having the list printed manually can we just use string concatenation & list comprehension here to insert `ana_config.keys()`? This would ensure a copy/paste error is not committed and be easier to maintain.

bdorney:
It seems you are just changing the dictionaries and tools for filepath look-up and not the analysis tool itself.  e.g. `anaUltraThreshold.py` remains unchanged.

Does this file script work out of the box on threshold scans taken without the `--perchannel` option?  I would be surprised if so.

anaInfo.py
jsturdy: Hmm, I may have left it to ensure that legacy behaviour remained, but can't remember; for sure with the check against valid options, it will never be used, so can probably be removed

jsturdy:
* Yes, this is correct.  In principle, there is no needed change to the analysis, there will be empty plots, yes, but the plot of interest, the per VFAT VT1 scan distribution is still made (`ProjectionY` over a histogram that fills only one of the channel bins)
* "Works" yes, produces sensible `vfatConfig.txt` no, but this is quite strange, as manually the operation that is done in the creation of this file works on the output `ROOT` file.  In the generated `vfatConfig.txt`
  * `vfatID` is always `-1`
  * `vt1` is always `0`
* Currently trying to figure out the above, which was not noticed initially, since the output plots were sensible.

bdorney:
looks like the issue with `vfatID` always being `-1` is resolved by: https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/155

Is the `vt1` issue still outstanding?

bdorney:
Seems legit.

Issue #64, opened 2018-01-22
Feature Request: 3x8 Grid Plots Ordered by Physical VFAT Position


bdorney:
## Brief summary of issue

Right now the 3x8 plots that all our plotting software makes orders the VFATs as (from left-to-right):

- Row 1: VFAT0 -> VFAT7
- Row 2: VFAT8 -> VFAT15
- Row 3: VFAT16 -> VFAT23

However if you physically look at the detector you would have to "invert" this plot to match the physical location of the VFATs when facing the readout PCB.  Suggest to change to:

- Row 1: VFAT16 -> VFAT23
- Row 2: VFAT8 -> VFAT15
- Row 3: VFAT0 -> VFAT7

### Types of issue

- [ ] Bug report (report an issue with the code)
- [X] Feature request (request for change which adds functionality)

## Expected Behavior

The 3x8 grid plot more follows physical layout of chips on the detector when facing the readout PCB.

## Current Behavior

The 3x8 grid plot is inverted w.r.t. the physical layout of chips on the detector when facing the readout PCB

## Context (for feature requests)

It's going to make our lives much easier when mass production and quality control starts given that there will be inexperienced users.  Having an "at a glance" representation that matches the physical layout will reduce complexity and potential errors in the long-term.

PR #80, opened 2018-04-11
Addresses #64: 3x8 Grid TCanvas Objects Now Follow Physical VFAT Layout

bdorney:
## Description

With superchamber construction coming soon we will be doing a lot of benchtop testing before/after assembly of detectors.  To reduce confusion I have made the TCanvas objects that produced 3x8 grid plots follow the physical VFAT ordering when looking from above the detector with the narrow (wide) base of the trapezoid on your left (right). 

### Types of changes

The "Breaking Change" may not be the correct designation, but it *will* change the order on how plots are shown and may catch certain users off guard.  Overall though I believe this will **save** trouble in the long run.

## Motivation and Context

When looking at 3x8 grid plots produced by our tools the VFAT numbering in the plot does not match the physical layout of the detector, e.g. VFATs 0->7 appear where VFATs 16-23 are located.  During the mass production of superchambers and subsequent in lab testing by non-experts this has a strong possibility to generate confusion and/or mistakes. 

## How Has This Been Tested?

Reanalyzed old data.

summary

## Checklist:

### DIFF
--- a/anaSBitThresh.py
+++ b/anaSBitThresh.py
@@ -47,12 +47,13 @@ for event in inF.rateTree :
 #Save Output
 outF.cd()
 from anautilities import make3x8Canvas
+from mapping.chamberInfo import chamber_vfatPos2PadIdx
 canv_RateSummary = make3x8Canvas('canv_RateSummary', vRate, 'hist')
 dirVFATPlots = outF.mkdir("VFAT_Plots")
 dirRatePlots1D = dirVFATPlots.mkdir("Rate_Plots_1D")
 for vfat in range(0,24):
-    canv_RateSummary.cd(vfat).SetLogy()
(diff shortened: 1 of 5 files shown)


jsturdy:
Can you add an old plot and describe the confusion you feel will arise? (I want to see if I can similarly confuse myself with the new layout)

bdorney:
So for example when we were reviewing the old `GEMINIm01L1` we wanted to replace [VFAT4](http://cmsonline.cern.ch/cms-elog/1027123) because there was a concern it had burned inputs.  At first the wrong chip (`VFAT20`) was removed because the operator wanted to reference the plot and not the physical layout.  This image shows the original ordering which does *not* follow the physical detector layout:

vfatordermismatchin3x8grid

Issue #83, opened 2018-04-16
Feature Request: Algorithmic Identification of Burned/Disconnected VFAT2 Inputs


bdorney:
## Brief summary of issue

Based on [slide 22](https://indico.cern.ch/event/721622/contributions/2968019/attachments/1631961/2602748/BDorney_GEMDAQMtg_20180412_BurnedVFATInputs.pdf) `vfat2` channels that have been burned during operation have a characteristic s-curve width.  This s-curve width could be used to mark the channel as having a "burned input" and thus could be factored into our "% active channels" list when reporting to CMS.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [X] Feature request (request for change which adds functionality)

## Expected Behavior

The `maskReason` class of `anaInfo.py` is updated to have a new field `DisconnectedInput` (it could also be called `BurnedInput` but we would need to know a priori if the channel was always like this or was assigned this classifier after some time).  This could be added:

https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/6445dcb44ad50ef49fb2c48eed9e49171ce72861/anaInfo.py#L41-L54

The hex assignment could be `BurnedInput = 0x20` which would be `0b100000`.  This keeps with the standard of the `maskReason` (maybe we don't mask these but we have a human readable identifier) class where each reason corresponds to a bit in the overall maskReason value (recall channels can have more than one maskReason).

Then in `anaUltraScurve.py` the "create reason array" block could be updated to search for channels with this behavior:

https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/6445dcb44ad50ef49fb2c48eed9e49171ce72861/anaUltraScurve.py#L361-L376

Here the solution would be something like:

```
reason[4.14E-02 < scanFitResults[1][vfat] and scanFitResults[1][vfat] < 1.09E-01] = MaskReason.DisconnectInput
```

Here the `4.14e-2` comes from the lower bound of the high purity burned sample and `1.09e-1` comes from the upper bound of the literature value.  Numbers taken from [slide 22](https://indico.cern.ch/event/721622/contributions/2968019/attachments/1631961/2602748/BDorney_GEMDAQMtg_20180412_BurnedVFATInputs.pdf). However these cut values could be influenced based on the result of #82.

## Current Behavior

Presently channels which have become burned, or were always disconnected, are erroneously included in our "live channel fraction" when reporting to CMS.

## Context (for feature requests)

[...]

Issue #77, opened 2018-03-29
Discussion: Charge in fC For HighNoise or HighEffPed?


bdorney:
## Brief summary of issue

When analyzing s-curves the results will now be calculated in fC.  Before the `maskReason` for `HighNoise` and `HighEffPed` was determined in DAC units of `VCAL` or `CFG_CAL_DAC`.  This presents a problem as there is chip-to-chip variations.  Additionally there are differences between the `vfat3` and `vfat2` calibration modules so the step size of the DAC unit (in `mV`) is also different.

I think we should make a push for determining what `HighNoise` and `HighEffPed` classify as in terms of `fC` instead of DAC units.  It would be good if we could also keep the MIP signal charge in mind. 

Maybe this is a job for the DPG team but since it's in the code thought perhaps should open an issue page.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [ ] Feature request (request for change which adds functionality)
- [X] Discussion

## Expected Behavior

`HighNoise` and `HighEffPed` classifiers should be defined on a value of `fC`.

## Current Behavior

In the `master` branch they compared against hard coded values that are in terms of DAC units.  This means that a `HighEffPed` in one chip will mean a different amount of charge in another chip:

https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/13607095bcc0895f3a2e34c342644a10400d5fd9/anaUltraScurve.py#L286-L287

In the `develop` branch these quantities are in `fC` by default but compared against the old DAC unit cuts:

https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/anaUltraScurve.py#L373-L374

Probably need some sort of study that defines this.  

## Context (for feature requests)

At some point CMS will ask us for our dead channel fraction again and may ask more detailed question than just "channels masked/unmasked."

bdorney:
So based on results shown here https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/78 the effective pedestal cut of 50 is pretty high.

However this cut is an "event count" so it should scale with the number of events that were used to take the s-curve.  For example in the plots shown in https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/78 only 100 events were used to generate the s-curve.  

So this cut should probably be re-cast to be a percentage of the input number of events used.

For example:

```
maxPedPercentage = 0.05
effPedThreshold = maxPedPercentage * scurveEventCount
[...]

PR #87, opened 2018-04-21
New Feature: LSF Batch Submission Tools

bdorney:
## Description

Now possible for a developer to setup the computing environment on `lxplus` and take advantage of the LSF Batch Submission for reanalyzing large amounts of calibration data.  Examples have been provided in the updated `README.md`.

Also few quality of life improvements, in all cases see updated `README.md` for further details or examples:
- It is no longer necessary to provide `plotTimeSeries.py` a template file, it will automatically choose the last two weeks worth of scandates,
- `plotTimeSeries.py` can now be used to automatically generate a `listOfScanDates.txt` file with the two-column format for a range of scandates supplied,
- `gemPlotter.py` can be passed the `--skipBadFiles` argument to ignore `scandates` in the input `listOfScanDates.txt` file which generate an `IOError` (E.g. `TFile` is a zombie, `TTree` was not written, etc...),
- `package4Docker.py` can be configured to only package raw data with the `--onlyRawData` argument, and
- `package4Docker.py` can ignore errors in tarball creation due to `IOError`'s with the `--ignoreFailedReads` argument.

Finally addresses:
https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/83
https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/77

For the above the command line arguments are described in the updated `README.md`.  Default values are chosen as:

- `HighEffPed` if effective pedestal is 5% of injected pulse count,
- `HighNoise` if the scurve_sigma is greater than `1 fC`,
- `DeadChannel` if following the window discussed in #83.

Note a channel assigned `DeadChannel` is not masked, but the value is tracked in the `maskReason` branch.

### Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Motivation and Context

In hunting for burned VFAT2 inputs it was necessary to re-analyze the entire s-curve dataset to ensure the stored values would be in charge units.  It is not practical to do this by hand.

## How Has This Been Tested?

Following, to completion, the example shown in the `README.md` which uses the scripts that have been modified here.

https://mattermost.web.cern.ch/cms-gem-daq/pl/gfgd19f3jprk8ng9xkqu7dtffe
Here we see these channels for this VFAT were always dead.

## Checklist:

- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.


bdorney:
Additionally @jsturdy please review for accuracy the description shown in this [table](https://github.com/bdorney/gem-plotting-tools/blob/feature_clusterModeLxplus/README.md#deriving-channel-configuration) for describing the derivation of the channel config for the `v2b` case in P5.

macros/clusterAnaScurve.py
```diff
@@ -0,0 +1,234 @@
...
+    envCheck('ELOG_PATH')
+
+    buildHome= os.getenv('BUILD_HOME')
+    dataPath = os.getenv('DATA_PATH')
+    elogPath = os.getenv('ELOG_PATH')
+
+    # Get info from input file
+    from anautilities import getDirByAnaType, filePathExists, makeListOfScanDatesFile, parseListOfScanDatesFile
+    if (listOfScanDatesFile is None and options.chamberName is not None):
+        makeListOfScanDatesFile(options.chamberName, options.anaType, ztrim=options.ztrim)
```
mexanick: I would add the possibility to set start/end scan dates here. I.e. I have no list of scan dates file (or have a wrong one), but I want to have one automatically created (updated) for certain desired dates. In a current implementation the new list will be created for all dates from the oldest one till today

macros/clusterAnaScurve.py
```diff
@@ -0,0 +1,234 @@
...
+        jobScript.write('echo `gcc --version | grep "gcc"`\n')
+        jobScript.write('source /opt/rh/python27/enable\n')
+        jobScript.write('source /afs/cern.ch/sw/lcg/contrib/gcc/4.8.4/x86_64-slc6/setup.sh\n')
+        jobScript.write('export BUILD_HOME=%s\n'%buildHome)
+        jobScript.write('export DATA_PATH=%s\n'%dataPath)
+        jobScript.write('export ELOG_PATH=%s\n'%elogPath)
+        jobScript.write('source %s/cmsgemos/setup/paths.sh\n'%buildHome)
+        jobScript.write('source %s/gem-plotting-tools/setup/paths.sh\n'%buildHome)
+        jobScript.write('source $BUILD_HOME/venv/slc6/py27/bin/activate\n')
+        jobScript.write('echo `python --version`\n')
```
mexanick: This and next line are copy of lines 143-144. Do we need them in both places?

macros/clusterAnaScurve.py
```diff
@@ -0,0 +1,234 @@
...
+            pass
+        else:
+            runCommand(jobCmd)
+            time.sleep(1)
+            pass
+        pass # end loop over listChamberAndScanDate
+
+    print("Job submission completed")
+    print("To check the status of your jobs execute:")
+    print("")
```
mexanick: Is there any particular reason to use `print("")` instead of adding \n in preceding print statement?

macros/plotTimeSeries.py
```diff
@@ -1,22 +1,17 @@
...
-    with open(template,"rt") as fin:
-        with open(ofname, "wt") as fout:
-            for line in fin:
-                fout.write(line.replace("GEMINI", chamberName)) 
-
-def makePlots(chamberName, scanType, vt1bump, elog_path):
-    call_command = 'gemPlotter.py --infilename='+getDirByAnaType(scanType,chamberName)+'listOfScanDates_gemPlotter.txt --anaType=scurveAna --branchName=threshold --make2D --alphaLabels -c -a'
+def makePlots(chamberName, anaType, vt1bump, elog_path):
+    call_command = 'gemPlotter.py --skipBadFiles --infilename='+getDirByAnaType(anaType,chamberName)+'listOfScanDates.txt --anaType=scurveAna --branchName=threshold --make2D --alphaLabels -a --axisMax=10'
+    os.system(call_command)
```
mexanick: please consider refactoring the gemPlotter so it can be called in a pythonic way

macros/clusterAnaScurve.py
bdorney: Yes this can be implemented.

macros/clusterAnaScurve.py
bdorney: They were exceptionally useful in debugging as on the worker node the python and gcc versions were not being set correctly.  If you object then I would prefer to remove the lines at 143-144 and keep 152-153 (as they reflect the version of python and gcc that the job is executed with)

macros/clusterAnaScurve.py
bdorney: I find readability of the code enhanced in this way as I often miss special characters such as `\n`.

macros/plotTimeSeries.py
bdorney: For sure that will not come in this PR or an upcoming PR on a short-time scale.  The bulk of the work being done by `gemPlotter.py` is already "pythonic" since it is separated into two functions:

- `arbitraryPlotter(...)`,
- `arbitraryPlotter2D(...)`.

If one wanted they could import these, and then build around importing that; if this is desired then these functions could be moved to `anautilities.py` or an independent `plottingUitilties.py` file.

But additional re-factoring of this tool I think is a lower priority compared to other critical tasks at present.

macros/clusterAnaScurve.py
jsturdy: you can also use triple quotes for multi-line strings

Issue #96, opened 2018-05-22
Bug Report: Problems generating full list of scan dates for plotTimeSeries.py


QuantumKittydynamics:
### Types of issue

- [X] Bug report (report an issue with the code)
- [ ] Feature request (request for change which adds functionality)

## Expected Behavior

When I type the following command:

python plotTimeSeries.py --listOfScanDatesOnly --startDate=2017.01.01

for each detector defined in chamber_config.values() the listOfScanDAtes.txt file should be found at $DATA_PATH/<ChamberName>/scurve/

## Current Behavior

Instead, it halts the process and outputs the following error:

Traceback (most recent call last):
  File "plotTimeSeries.py", line 26, in <module>
    from gempython.utils.wrappers import envCheck
ImportError: No module named gempython.utils.wrappers

### Steps to Reproduce (for bugs)

1. python plotTimeSeries.py --listOfScanDatesOnly --startDate=2017.01.01

## Context (for feature requests)

I am trying to generate the full list of scan dates since 2017 installation for the various detectors to study the presence of burned VFAT inputs in time-series SCurve plots.

## Your Environment

* Version used:
* Shell used: Mac terminal

mexanick:
@QuantumKittydynamics, 
please report:
1. PC where you tried to run it
2. Account from which you've tried to run
3. How the package was installed (system-wide install of RPM or local github clone?)
4. Version of the package if installed as RPM or github branch/commit

bdorney:
Additionally please include the commands, from the very beginning, and their outputs.

bdorney:
Possibly related to:

https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/94
https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/95

QuantumKittydynamics:
1. My MacBook Pro
2. estarlin on LXPlus
3. Local github clone
4. commit dd3c00f47844ac620279ac57e6dd1339d188583b (I think)

mexanick:
If you connect to a certain PC over ssh in your macbook terminal and then run the code there - it is that PC of our interest. So as far as I can understand, a correct answer to the first question should be ''lxplus"
Now if you are using the local github clone of latest, please cross check the commit by running `git show`
Then if you have indicated the commit correctly, you must build a package and install it under the virtual environment (this is valid for lxplus, on other PCs, like gem904* or P5 ones there could be other options)

bdorney:
[...]

Issue #95, opened 2018-05-18
Update clusterAnaScurve.py for new computing environment.


bdorney:
## Brief summary of issue

Since `gem-plotting-tools` has been updated to be more of a package that is installed or downloaded as a standalone release the `clusterAnaScurve.py` tool no longer properly sets up the computing env on the worker node of lxplus.

Additionally CERN IT is moving lxplus away from using the scheduler `bsub` to moving `HTCondor`:

http://batchdocs.web.cern.ch/batchdocs/

Now might be a good time to migrate to `HTCondor`.

## Expected Behavior

The worker node on lxplus should:
1. Download a given release and all supporting software,
2. Unpack it and install it,
3. Set the env
4. Launch the analysis
5. Return the results

You may have to adapt this procedure to instead download all tarballs/zip files, and then send them to the worker node instead of having the worker node do this. 

Presently right now the worker node is using AFS directories this typically is "frowned" upon and angry emails may be sent from cluster administers.  Perhaps we should alter this.

## Current Behavior

The computing env setup by `clusterAnaScurve.py` is no longer compatible with how the package is installed as either a rpm, zip, or tarball.  Specifically the following lines need to be updated:  

https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/dd3c00f47844ac620279ac57e6dd1339d188583b/macros/clusterAnaScurve.py#L150-L163

Additionally the example shown in the `README.txt` needs to be updated:
https://github.com/cms-gem-daq-project/gem-plotting-tools/#full-example-for-p5-s-curve-data

## Context (for feature requests)

Need cluster computing abilities to re-analyze all s-curve data taken in P5.

bdorney:
Partially related to:

https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/94

In that the documentation for using `clusterAnaScurve.py` may need to be updated.

bdorney:
Support for the lxplus batch submission system was added by #105.  Given limited resources let's close this for now; it can be reopened later if `HTCondor` goes live.

Issue #102, opened 2018-05-30
Bug Report: macros lack executable permissions post release install


bdorney:
## Brief summary of issue

Following instructions [developer](https://gist.github.com/jsturdy/4f9bccbdd18d953f73777e492a71fd24) or [user](https://gist.github.com/jsturdy/a82a6b7b7d497ae8ef98ee736119c6b5) instructions for setting up a scripts placed under:

```
$VENV_ROOT/lib/python2.7/site-packages/gempython/gemplotting/macros 
```

Lack executable permissions.  Originally reported by @lmoureaux at [here](https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/96#issuecomment-393224126)

### Types of issue

- [X] Bug report (report an issue with the code)
- [ ] Feature request (request for change which adds functionality)

## Expected Behavior

Scripts found at:

```
$VENV_ROOT/lib/python2.7/site-packages/gempython/gemplotting/macros 
```

Should have executable permissions.

## Current Behavior

Lack of executable permissions:

```
% ll venv/slc6/py27_postpkg/lib/python2.7/site-packages/gempython/gemplotting/macros 
total 182K
-rw-r--r--. 1  zh 2.9K May 30 22:20 summary_plots.pyc                                                                                                                                                                                    
-rw-r--r--. 1  zh 2.8K May 30 22:20 summary_plots.py
-rw-r--r--. 1  zh 5.1K May 30 22:20 scurvePlottingUtitilities.pyc
-rw-r--r--. 1  zh 5.2K May 30 22:20 scurvePlottingUtitilities.py
-rw-r--r--. 1  zh  527 May 30 22:20 plot_vfat_summary.pyc
-rw-r--r--. 1  zh  296 May 30 22:20 plot_vfat_summary.py
-rw-r--r--. 1  zh  643 May 30 22:20 plot_vfat_and_channel_Scurve.pyc
-rw-r--r--. 1  zh  420 May 30 22:20 plot_vfat_and_channel_Scurve.py
-rw-r--r--. 1  zh 3.6K May 30 22:20 plotTimeSeries.pyc
-rw-r--r--. 1  zh 4.2K May 30 22:20 plotTimeSeries.py
-rw-r--r--. 1  zh 1.4K May 30 22:20 plot_scurves_by_thresh.pyc
-rw-r--r--. 1  zh 1.1K May 30 22:20 plot_scurves_by_thresh.py
[...]

PR #105, opened 2018-05-31
Update clusterAnaScurve.py for the new environment

lmoureaux:
## Description

From the commit message:

    Fix clusterAnaScurve.py to work with packages
    
    Copy all environment variables and activate the virtualenv in the batch job.
    Expected to work with Python 2.6 and 2.7, regardless of how the user's
    virtualenv was setup.

No behavior change, therefore no need to update the documentation.

Depends on #103.

### Types of changes

## Motivation and Context

This addresses the urgent part of #95. HTCondor and not using `afs` left for later.

## How Has This Been Tested?

* Used [this](https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/96#issuecomment-393208232) `virtualenv` with Python 2.7
* Started the script on all data for GEMINIm30L2, killed it after three jobs were submitted.
* Checked that `anaUltraScurve.py` is being called
* Checked that the summary plot of the one `anaUltraScurve.py` that didn't crash is produced (see below)
* Launched an analysis of all S-curves to run overnight

screenshot_20180531_231132

## Checklist:

- [ ] All new and existing tests passed. --> there's a test bot, right?


bdorney:
> Used this virtualenv with Python 2.7

Can you try again using: https://gist.github.com/bdorney/5f3965bb8b26858a91ef8087137d1ec9

Example call:

```
source setup_gemdaq.sh -c 0.3.1 -g 1.0.0 -G 5 -v 2.0.0 -V 3 -p $PWD/venv
```

> Checked that the summary plot of the one anaUltraScurve.py that didn't crash is produced (see below)

For those that did crash is the reason understood? e.g. is it related to the software or for example a bad input file.

bdorney:
Additionally before this is approved I'd like the instructions in the `README.md` to be updated to reflect necessary changes (this can come as a later commit):

I think only the area under [setup](https://github.com/cms-gem-daq-project/gem-plotting-tools/#setup) requires changes from what I've seen in this PR.

lmoureaux:
>> Used this virtualenv with Python 2.7
>
> Can you try again using: https://gist.github.com/bdorney/5f3965bb8b26858a91ef8087137d1ec9

* Installed using:
    ```
    python -m pip install -U rpm/dist/gempython_gemplotting-1.0.0.tar.gz
    ```
* `chmod +x`'ed the scripts
* Produced the map files
* Launched one non-failing job again

=> Everything works as expected

>> Checked that the summary plot of the one anaUltraScurve.py that didn't crash is produced (see below)
>
> For those that did crash is the reason understood? e.g. is it related to the software or for example a bad input file.

It complains about the S-curve tree not being present, and the input file size is <700 bytes. Bad input file.

> Additionally your tests built the rpm after developing the festure?

Couldn't get the rpm to build in my `afs` area (even before my changes): `make rpm` fails but produces the `pip` package anyway. Not sure if it's because development on `lxplus` isn't supported.

The rpm was built in Travis. I don't know if it works and have no way to test this (ie no admin access to any slc6/cc7 machine).

> Additionally before this is approved I'd like the instructions in the README.md to be updated to reflect necessary changes (this can come as a later commit):
> I think only the area under setup requires changes from what I've seen in this PR.

The updated `clusteraAnaScurve ` would work in the old environment. It's independent of how one sets up its `virtualenv`, as long as there is one.

Updating the README would require a lot more testing (incl. at P5, where I've been unable to create a working env without a hack).

lmoureaux:
> Can you please try again and report back on issues encountered? Showing the exact steps would be useful.

I had no issue with the setup script on `lxplus`:

* Download the script: `wget raw.githubusercontent.com/whereever-the-script-was`
* `chmod +x` the script
* `source setup_gemdaq.sh -c 0.3.1 -g 1.0.0 -G 5 -v 2.0.0 -V 3 -p GEM/venv` from `bash`
* `python -m pip install -U rpm/dist/gempython_gemplotting-1.0.0.tar.gz` from the root of my local checkout
* `chmod +x` the `gemplotting` scripts (#102)
* Produce the map files again
* Run -- everything works

lmoureaux:
>> The updated clusteraAnaScurve would work in the old environment. It's independent of how one sets up its virtualenv, as long as there is one.
>
> Right, updating the twiki to include the instructions on how to do this is what's being asked.
>
>> Updating the README would require a lot more testing (incl. at P5, where I've been unable to create a working env without a hack).
> 
> Indeed, this is what's being asked. 

In order to keep this discussion focused, I opened #109 about updating the README. I'll test again when the environment is set in stone.

lmoureaux:
## Force-pushed changes

* Rebased on top of `develop`
* Fix a bug introduced upstream

## Testing update

Now tested with setup instructions from the [new README] on `lxplus`:

```
<follow README instructions>
chmod +x venv/slc6/py2.7/lib/python2.7/site-packages/gempython/gemplotting/macros/*.py
clusterAnaScurve.py --anaType=scurve -i /afs/cern.ch/work/l/lmoureau/GEM/data/gemdata/GEMINIm30L2/scurve/listOfScanDates.txt
```

Three jobs submitted, all produced the expected output (2 fails because of bad data, 1 success with plot produced).

anautilities.py
```diff
@@ -66,7 +66,7 @@ def getDirByAnaType(anaType, cName, ztrim=4):
         pass
 
     # Check Paths
-    from ..utils.wrappers import envCheck
+    from gempython.utils.wrappers import envCheck
```
jsturdy: Simultaneous changes, but the others were merged in earlier, can you revert yours, or make them conform to the current `develop`? (also L128, L335)

bdorney:
> @bdorney, anything else you are still waiting on?
Will plan on putting this into a stable released version tonight

For the purposes of `clusterAnaScurve.py` I think this is suitable.  The issues I raised here are being addressed by #109.

anautilities.py
lmoureaux: If you're talking about commit `2ac2945`, read again: this *reverts* your change (which crashed the program). This PR is based on today's `develop`.

Context: `anautilities.py` is installed in `gempython/gemplotting/utils`, while `wrappers.py` is in `gempython/utils`

jsturdy: ah, ok, then the relative import should be `...utils.wrappers`

lmoureaux: For my own education: `N` leading dots are translated into `N-1` `../` relative to the current file ?

lmoureaux:
@dorney I force-pushed to update the last commit, will need your approval again

anautilities.py
jsturdy: For [reference](https://www.python.org/dev/peps/pep-0328/)

Issue #104, opened 2018-05-31
packageFiles4Docker.py is missing from the package


lmoureaux:
## Brief summary of issue

The package `gempython_gemplotting-1.0.0.tar.gz` doesn't contain `packageFiles4Docker.py`. This file is referenced in the README instructions to analyze P5 data.

### Types of issue

- [x] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

## Expected Behavior

After installing the package and setting up `$PATH`, `packageFiles4Docker.py` should be available.

## Current Behavior

`packageFiles4Docker.py` isn't even present in the package.

### Steps to Reproduce (for bugs)

1. [Install](https://gist.github.com/jsturdy/a82a6b7b7d497ae8ef98ee736119c6b5) the `gemplotting` package
2. Run `find path/to/your/venv | grep packageFiles4Docker.py`
3. Nothing is found

## Context (for feature requests)

Trying to get some data to analyze to tackle #95

## Your Environment

* Version used: 1.0.0
* Shell used: `bash`

Issue #107, opened 2018-06-01
Bug Report: make rpm fails to build the package on lxplus


lmoureaux:
## Brief summary of issue

Even though `lxplus` isn't a supported development environment, some of the tools cannot be run on the P5 machines (eg because they require cluster access). Developing and testing on different  machines would be painful.

`make rpm` crashes on `lxplus`.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

## Expected Behavior

`make rpm` works on lxplus.

## Current Behavior

[As recommended](https://gist.github.com/jsturdy/4f9bccbdd18d953f73777e492a71fd24#gistcomment-2605577), the following wasn't done in a `virtualenv`:
```
$ cd GEM/gem-plotting-tools/
$ LANG=C make && LANG=C make rpm
PythonModules=["gempython.gemplotting", "gempython.gemplotting.utils", "gempython.gemplotting.fitting", "gempython.gemplotting.macros", "gempython.gemplotting.mapping" ]
Using BUILD_HOME=/afs/cern.ch/user/l/lmoureau/GEM
OS Detected: slc6
mkdir -p pkg/gempython/gemplotting/utils
PythonModules=["gempython.gemplotting", "gempython.gemplotting.utils", "gempython.gemplotting.fitting", "gempython.gemplotting.macros", "gempython.gemplotting.mapping" ]
Using BUILD_HOME=/afs/cern.ch/user/l/lmoureau/GEM
OS Detected: slc6
mkdir -p pkg/gempython/gemplotting/utils
mkdir -p pkg/gempython/scripts
cp -rf README.md LICENSE CHANGELOG.md MANIFEST.in requirements.txt pkg/gempython/gemplotting
cp: cannot stat `CHANGELOG.md': No such file or directory
make: [preprpm] Error 1 (ignored)
cp -rf README.md LICENSE CHANGELOG.md MANIFEST.in requirements.txt pkg
cp: cannot stat `CHANGELOG.md': No such file or directory
make: [preprpm] Error 1 (ignored)
Running _setup_update target
PackagePath ackagePath
mkdir -p /afs/cern.ch/user/l/lmoureau/GEM/gem-plotting-tools/rpm/RPMBUILD
if [ -e /afs/cern.ch/user/l/lmoureau/GEM/gem-plotting-tools/setup.py ]; then \
[...]

PR #108, opened 2018-06-04
Updates to packaging

jsturdy:
## Description
Few tweaks based on feedback of current packaged files
* Add `packageFiles4Docker.py` to built packages (cf. #104)
* Update `import`s syntax in `anautilities`
* Not addressing #102, as a better pythonic solution should be preferred (e.g., utilize the `entry_points` feature of `setuptools`), which requires more thought and changes
* Add a `requirements-dev.txt` file for the minimum necessary dependencies to be able to build the package (`pip install -r requirements-dev.txt` into your development `virtualenv`, or `pip install --user -r requirements-dev.txt` into your user area)

## How Has This Been Tested?
Update to `anautilities` has not been, but is a semantic change only, a proof-of-principle of proper relative imports that will be extended to other package specific modules (where appropriate, e.g., not in tools that utilize a `__main__`).
The other update is non-functionality change.


jsturdy:
Added also updated `gembuild` `submodule` for `pip`-only `make` targets for building/developing inside a `virtualenv`

requirements-dev.txt
```diff
@@ -0,0 +1,3 @@
+importlib
+setuptools>25,<=38
```
lmoureaux: I used `setuptools<38` to get `make rpm` to work with Python 2.6 (see [here](https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/107#issuecomment-394319947)). It wasn't working with `<39`.

jsturdy: Do you know which version got pulled in?

anautilities.py
```diff
@@ -66,7 +66,7 @@ def getDirByAnaType(anaType, cName, ztrim=4):
         pass
 
     # Check Paths
-    from gempython.utils.wrappers import envCheck
+    from ..utils.wrappers import envCheck
```
bdorney: Probably my ignorance here but I don't really follow what this (change to line 69) does; or the motivation for it?

I guess it's related to the discussion:

>Update imports syntax in anautilities

> Update to anautilities has not been, but is a semantic change only, a proof-of-principle of proper relative imports that will be extended to other package specific modules (where appropriate, e.g., not in tools that utilize a __main__).

jsturdy: `anautilities.py` lives in `gempython/gemplotting/`
It wants to `import` from `wrappers.py` which lives in `gempython/utils`
Because `anautilities.py` will **never** be called standalone (no `__main__`), it is a "part of the package", so it should reference all `import`s from the `gempython` namespace relatively (wrt where it lives) so it imports from `..utils.wrappers` (i.e., look in the package structure in the directory above where I am for `utils/wrappers.py`)
This type of `import` *should* (and **can**) only be done as deep as the package itself is (don't `import .....somewhere.outside.of.gempython`), and **must** not be done for anything that will be called either as a standalone, lives-on-the-`PATH` executable, nor for a module with a `__main__` (executable via `python -m <modulename>`, as those pieces are not a "part" of the package in the same sense (and can be located outside of the package itself)

bdorney: Ah so the `import ..some.package`

Is like calling `cd ..`.  I understand.

Issue #61, opened 2018-01-17
"Bug" Report: Error Bars on anaUltra*.py Scripts


bdorney:
## Brief summary of issue

We've migrated all our DAQ (and I guess also DQM?) machines to cc7 which defaults to ROOT v6.X.  In this version of ROOT by default TH1::Sumw2() is automatically set, see the [TH1 Class Documentation](https://root.cern.ch/doc/master/classTH1.html#a8cd3b44217c8fec1d74046ed2ac153c2).  This causes error bars of some of our histograms to not be set correctly.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [ ] Feature request (request for change which adds functionality)

## Expected Behavior

For a counting experiment (almost all of our scan routines) the error bar should go as `sqrt(N)` where `N` is the bin content.

## Current Behavior

Due to the differences between ROOT 6.X and 5.34 the `TH1::Sumw2()` method is automatically called.

Our analysis routines usually use the [Int_t TH1::Fill(Double_t x,Double_t w )](https://root.cern.ch/doc/master/classTH1.html#a498de8e0804e75fc75e62dc14a3bb62d) method for storing values in the histogram where `x` and `w` come from the input `TTree` object.  However this is only being called once per bin so the error bar ends up equalling `w` e.g. the sum of weights squared is a sum over one number.

See for example:

- anaUltraLatency.py [Line 55](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/6f831848c4a39298a8d700e45eee87f58bed7838/anaUltraLatency.py#L55)
- anaUltraScurve.py [Line 249](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/6f831848c4a39298a8d700e45eee87f58bed7838/anaUltraScurve.py#L249) 
- anaUltraThreshold.py [Line 112](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/6f831848c4a39298a8d700e45eee87f58bed7838/anaUltraThreshold.py#L112)
 so in all instances where we are filling 

### Steps to Reproduce (for bugs)

1. Run any anaUltra script on a machine where ROOT is v6.X or higher.
2. Look at the error bars on the produced plots

### Proposal 1
Call [TH1::SetDefaultSumw2(False)](https://root.cern.ch/doc/master/classTH1.html#a8cd3b44217c8fec1d74046ed2ac153c2
[...]

PR #110, opened 2018-06-06
Maintenance: Update CFG_CAL_DAC to Charge Conversion for v3

bdorney:
## Description

The `vfat3` team has decided to change the convention for the calibration coefficients `calDAC2Q_m` and `calDAC2Q_b`.  Previously in voltage step mode the "inverted" `CFG_CAL_DAC` register value was used, i.e.:

```
Q = calDAC2Q_m * (256 - CFG_CAL_DAC) + calDAC2Q_b
``` 

Now the raw register value will be used, e.g.:

```
Q = calDAC2Q_m * (CFG_CAL_DAC) + calDAC2Q_b
```

### Types of changes

- [X] Maintenance (updates to the software that are transparent to the user)

## Motivation and Context

This was reported by Henri who is developing the `vfat3` production testing:

>My idea was to make the fit on the raw register value so the charge equation for the voltage mode would become:
>Q=m*CAL_DAC+b
>which in my mind would be more straight forward and avoid confusion about when the inversion is used and when not.

## How Has This Been Tested?

Took a new scurve and analyzed the data.

summary

Note VFAT5 & 12 were masked in this test.

## Checklist:

### DIFF
--- a/fitting/fitScanData.py
+++ b/fitting/fitScanData.py
@@ -48,10 +48,16 @@ class ScanDataFitter(DeadChannelFinder):
             self.scanFitResults[6][vfat] = np.zeros(128, dtype=bool)
             for ch in range(0,128):
                 self.scanCount[vfat][ch] = 0
-                self.scanFuncs[vfat][ch] = r.TF1('scurveFit_vfat%i_chan%i'%(vfat,ch),'[3]*TMath::Erf((TMath::Max([2],x)-[0])/(TMath::Sqrt(2)*[1]))+[3]',
-                        self.calDAC2Q_m[vfat]*1+self.calDAC2Q_b[vfat],self.calDAC2Q_m[vfat]*253+self.calDAC2Q_b[vfat])
-                self.scanHistos[vfat][ch] = r.TH1D('scurve_vfat%i_chan%i_h'%(vfat,ch),'scurve_vfat%i_chan%i_h'%(vfat,ch),
-                        254,self.calDAC2Q_m[vfat]*0.5+self.calDAC2Q_b[vfat],self.calDAC2Q_m[vfat]*254.5+self.calDAC2Q_b[vfat])
+                if self.isVFAT3:
+                    self.scanFuncs[vfat][ch] = r.TF1('scurveFit_vfat%i_chan%i'%(vfat,ch),'[3]*TMath::Erf((TMath::Max([2],x)-[0])/(TMath::Sqrt(2)*[1]))+[3]',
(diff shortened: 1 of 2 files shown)


fitting/fitScanData.py
```diff
@@ -68,7 +74,7 @@ def feed(self, event):
                 if(event.vcal > 254):
                     self.scanCount[event.vfatN][event.vfatCH] += event.Nhits
             else:
-                charge = self.calDAC2Q_m[event.vfatN]*(256-event.vcal)+self.calDAC2Q_b[event.vfatN]
+                charge = self.calDAC2Q_m[event.vfatN]*(event.vcal)+self.calDAC2Q_b[event.vfatN]
                 if((256-event.vcal) > 254):
```
BenjaminRS: Hi Brian, I am (very slowly) still looking into bug #61 and so have been converting charge to chargebin later in this script. Here should the 256 not also be removed from this line? So it becomes identical to L74?

bdorney: In this particular case no.  We are trying to count the number of `scurves` that have been processed.  So all we are interested in is the raw register value.  In the case of voltage step mode (e.g. `isCurrentPulse == False`) the height of the cal pulse is high when the register value is low.

fitting/fitScanData.py
```diff
@@ -119,8 +125,12 @@ def fit(self, debug=False):
         random = r.TRandom3()
         random.SetSeed(0)
         for vfat in range(0,24):
-            fitTF1 = r.TF1('myERF','[3]*TMath::Erf((TMath::Max([2],x)-[0])/(TMath::Sqrt(2)*[1]))+[3]',
-                        self.calDAC2Q_m[vfat]*1+self.calDAC2Q_b[vfat],self.calDAC2Q_m[vfat]*253+self.calDAC2Q_b[vfat])
+            if self.isVFAT3:
+                fitTF1 = r.TF1('myERF','[3]*TMath::Erf((TMath::Max([2],x)-[0])/(TMath::Sqrt(2)*[1]))+[3]',
+                            self.calDAC2Q_m[vfat]*253+self.calDAC2Q_b[vfat],self.calDAC2Q_m[vfat]*1+self.calDAC2Q_b[vfat])
```
mexanick: Please forgive me my ignorance, but how the factor `253` appeared here?
and also why do we need multiplication by factor `1`: `self.calDAC2Q_m[vfat]*1`?

bdorney: We need the axis to be in charge units.  With the updated convention Henri has put forward `calDAC2Q_m` will be negative and work with the raw `CFG_CAL_DAC` register.  In voltage pulse mode low `CFG_CAL_DAC` is high charge, and high `CFG_CAL_DAC` is low charge.  The factors [253,1] then specify the charge daq range which is something like [0, 60] fC.

PR #111, opened 2018-06-08
Ignore files generated by "make rpm"

lmoureaux:
## Description

Add the following entries to .gitignore:

```
pkg/LICENSE
pkg/MANIFEST.in
pkg/README.md
pkg/gempython
pkg/requirements.txt
rpm
```

### Types of changes

## Motivation and Context

After running `make rpm` or `make pip`, `git` thinks that the tree is dirty.

## How Has This Been Tested?

Ran `git status` after `make pip` and `make rpm` to ensure that I didn't miss any file.

## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,12 @@
 
 # Don't ignore github directory
 !.github
+
+# Ignore packaging files
+pkg/LICENSE
+pkg/MANIFEST.in
+pkg/README.md
+pkg/gempython
+pkg/requirements.txt
+rpm
+


bdorney:
In the interim as a reminder never use `git add -a .`

lmoureaux:
> Actually it would be even better if you force `.gitignore` to ignore everything with certain exceptions (like do not ignore *.py or *.h etc).

That's what the existing `.gitignore` does, but `make` creates Python and Markdown files in a "source" directory (`pkg`), and an additional directory (`rpm`). In your example, I can see the following:
```
# Ignore RPM directories
*/rpm
```
This is close to what I do in this PR.

mexanick:
actually, l37 in the example is not quite required... What I would rather like to see:
```
# Ignore everything
*
# Don't ignore subdirectories
!*/
# Don't ignore source files, etc
!*.gitignore
!*.c
!*.cpp
!*.cc
!*.C
!*.h
!*.hh
!*.hpp
!*.py
```

mexanick:
in this case even `git commit -a` won't really add too much )

lmoureaux:
@mexanick Please take a look at the [existing `.gitignore`](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/.gitignore).

mexanick:
ehm, how come then that pkg/<smth> and rpm/<smth> become trackable by git? Did someone add them? If so, `git rm` will help =)

lmoureaux:
> how come then that pkg/ and rpm/ become trackable by git?
```
# Don't ignore subdirectories
!*/
```

In addition, `pkg` contains tracked files that are needed to build the package.

jsturdy:
Sorry I didn't see this in time to prevent this merge, but you implemented it in a non-optimal way...
As @lmoureaux noted, you can *easily* ignore a specific directory even after `!*/` in the `.gitignore`, so I'm not sure why `pkg` as a whole wasn't ignored, and the few tracked files unignored...
Also, I'm not sure why this was an issue in the first place...
I would probably be in favour of reverting this and if deciding to go forward, doing it "right"

bdorney:
As you wish.

> ...this in time to prevent this merge...

To prevent further issues in the future perhaps we could set a designated time per week that approved PR's get merged.

I think the weekly Friday ops meeting might be an ideal time period.

Issue #88, opened 2018-04-23
Feature Request: Identifying Original maskReason & Date Range of Burned Input


bdorney:
## Brief summary of issue

We have 30720 channels in CMS presently.  And we need some automated way for tracking when changes occur.  Ideally this should be via the DB but this is not up and running (yet...).  But we have an automated procedure to make time series plots from the entire calibration dataset, see #87. 

So my proposal is to use these time series plots to identify:

- The original `maskReason` that is applied to a channel,
- The smallest date range for when a channel made the switch from "healthy" to burned, e.g. for [this](https://mattermost.web.cern.ch/cms-gem-daq/pl/ks7fa3yf67nnjejw3uhtk6eije) example that would be `2017.05.10.20.41` to `2017.05.31.09.21`.

### Issue with Channel Mask
Right now when a channel is masked it's `maskReason` will be recorded in the output `TTree` of the analysis scan.  However once a channel is masked the history will be somewhat "lost."  This is because the channel will be masked at the time of acquisition so the s-curve analysis will see that a fit to it fails (because there is no data).  The `maskReason` will then be assigned [FitFailed](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/0ce86672ddac3affee27b80517eb34e8cd50b029/anaInfo.py#L51).  This is because the scan software does not know the history.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [X] Feature request (request for change which adds functionality)

## Expected Behavior

To accomplish this a secondary analysis tool should be made, `macros/timeHistoryAnalyzer.py`, which should should have a function like:

```
def analyzeTimeSeriesPlot(timeSeriesHisto, obsType, deadChanCutLow=4.14E-02, deadChanCutHigh=1.09E-01):
"""
timeSeriesHisto - TH2 object created by gemPlotter.py when making time series plots
obsType - The observable to be investigate from the set {noise, ..., maskReason, etc...} see for example the observables produced here https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/0ce86672ddac3affee27b80517eb34e8cd50b029/macros/plotTimeSeries.py#L10-L26
deadChanCutLow - lower bound on s-curve width for identifying a dead channel, value in fC
deadChanCutHigh - higher bound on s-curve width for identifying a dead channel, value in fC
"""
# import numpy
# import nesteddict from gem sw

chanArray = nesteddict()
# Load info from the histo
[...]

Issue #68, opened 2018-03-10
Structure of RPM package


jsturdy:
### Information
* The packages and modules will be packaged and end up in the system `PYTHONPATH`, e.g., (default `python` on `cc7):  `/usr/lib/python2.7/site-packages/`
  * I've been toying with adding a global namespace `gempython`, under which there would be the `gemplotting`, `vftaqc`, etc., modules, but this is still to be finalized into a proposal, and then discussed
  * This would necessarily involve a small redesign of the `import` statements, but I have for the most part decided on a workable solution for people who are doing private development and **not** using the production software
* Executable scripts, e.g., `amc_info_uhal.py` will **all** go into `/opt/cmsgemos/bin`, to make it simple to add this to the `PATH` variable and have the script ready

#### Current RPM structure (**UPDATED 12-03-2018**)
```
gempython_gemplotting     /opt/cmsgemos/bin/anaInfo.py
gempython_gemplotting     /opt/cmsgemos/bin/anaUltraLatency.py
gempython_gemplotting     /opt/cmsgemos/bin/anaUltraScurve.py
gempython_gemplotting     /opt/cmsgemos/bin/anaUltraThreshold.py
gempython_gemplotting     /opt/cmsgemos/bin/anaXDAQLatency.py
gempython_gemplotting     /opt/cmsgemos/bin/ana_scans.py
gempython_gemplotting     /opt/cmsgemos/bin/anaoptions.py
gempython_gemplotting     /opt/cmsgemos/bin/anautilities.py
gempython_gemplotting     /usr/lib/python2.7/site-packages/gempython/__init__.py
gempython_gemplotting     /usr/lib/python2.7/site-packages/gempython/__init__.pyc
gempython_gemplotting     /usr/lib/python2.7/site-packages/gempython/__init__.pyo
gempython_gemplotting     /usr/lib/python2.7/site-packages/gempython/gemplotting/__init__.py
gempython_gemplotting     /usr/lib/python2.7/site-packages/gempython/gemplotting/__init__.pyc
gempython_gemplotting     /usr/lib/python2.7/site-packages/gempython/gemplotting/__init__.pyo
gempython_gemplotting     /usr/lib/python2.7/site-packages/gempython/gemplotting/fitting/__init__.py
gempython_gemplotting     /usr/lib/python2.7/site-packages/gempython/gemplotting/fitting/__init__.pyc
[...]

PR #112, opened 2018-06-10
Feature: time history analyzer

lmoureaux:
## Description

Add a new macro, `plotTimeSeries.py`, as requested in #88.

The macro works in three steps:

* [Bad scan removal](#bad-scan-removal): Scans that failed to produce consistent results are removed.
* [Range detection](#range-detection): The time evolution of each channel is searched for successive scans with consistent behavior. A set of such scans is called a (time) range. What kind of behavior is searched for is used-defined.
* [Analysis](#range-analysis): The properties of "ranges" are computed and printed.

### Bad scan removal

Scans that pass any the following cuts are removed:

* The average noise over the entire detector is lower than 0.1 fC (or `--minScanAvgNoise`). This cuts scans with none or very few channels responding.
 Here's the distribution of noise for all scans of GEMINIm30L2:
average_noise_distribution
0.1 was chosen as default instead of, say, 0.22 because other detectors may have lower noise, and maybe we'll manage to lower the noise. It's not a problem if some "bad" scans remain as long as most of them are removed.
* The fraction of masked channels is above 7% (or `--maxScanMaskedFrac`)

### Range detection

The time evolution of each channel is searched for successive scans with consistent behavior. A set of such scans "bad" scans for a given channel is called a (time) range; the definition of bad is user defined (see below).

Range finding starts with a list of scans, where each scan is marked as "good" or "bad". The definition of "bad" depends on what's being searched for (and "good" is always defined as "not bad"). The start of a range is determined by:

* Starts with a "bad" scan (see below)
* The channel wasn't "bad" in the previous scan (e.g going good to bad)

Then the range continues and the end of the range is determined by 5 consecutive good scans appearing (option: `--numEndScans`). To prevent the printing of spurious ranges due to transient effects ranges with less than 4 "bad" scans in total are suppressed (option: `--minBadScans`). A "range" found by this algorithm can have include some "good" scans.

As a side-effect, channels with sparse "bad" behavior, such as 88 in GEMINIm30L2, are also extracted. This can be controlled by tightening the cuts in the algorithm above.

Three definitions of "bad" are currently available:

* `mask`: the channel under consideration is masked
* `maskReason`: the channel under consideration has a non-zero `maskReason`
*  `zeroInputCap`: the channel under consideration has an scurve width that is consistent with zero input capacitance ( `4.14E-02 < scurevWidth < 1.09E-01 fC`). The precise values can be controlled using the `--minNoise` and `--maxNoise` options.

This is somewhat rough and not exactly what asked for in #88, but it finds problematic ranges correctly, including some with sparse bad scans, see eg channels 88-89 below.

### Analysis

For every "range" found in each of the VFATs, the following properties are computed and printed:

| Column header | Meaning |
|:-|:-|
| `ROBstr` or `vfatCH` | Strip number and VFAT channel, respectively |
| Last known good | Date and time of the last good scan before the range ("never" if the range starts at the first scan) |
| Range begins | Start date and time |
| Range ends | End date and time ("never" if the range includes the lastest scan) |
| #scans | Total number of scans (good and bad) |
| masked% | Percentage of `#scans` where the channel is "masked" not to be confused with "bad (useful to investigate channels that behave badly once in a while) |
| Initial `maskReason` | `maskReason` for the first scan in the range |
| Other subsequent `maskReason`s | `maskReason` not present for the first scan but found in a later scan in the same range |

A summary table of initial `maskReason` vs VFAT is also printed at the end.

### Types of changes

## Motivation and Context

See issue #88

## How Has This Been Tested?

Initially tested on GEMINIm30L2, with all S-curves taken before 1st of June. Ranges found matched existing features in data:

* When using the `mask` definition: they correspond to consecutive scans with the corresponding channel masked
* When using the `maskReason` definition: they correspond to consecutive scans where the corresponding channel has non-zero `maskReason`
* When using the `noise` definition: they correspond to consecutive scans with the corresponding channel blank in the plot (ie with noise below 0.05 fC)

Now testing on GEMINIm29L1 data provided by @bdorney and compared to slides 22-31 [here](https://indico.cern.ch/event/726585/contributions/2990039/attachments/1643406/2626210/DAQ_Meeting_-_05-03-18_extComments.pdf). The following configuration was used:

* `maskReason` ranges
* `--onlyCurrent` to select ranges that span until the latest scan
* Looking at entries with `DeadChannel` in either initial or later `maskReason`s

[Full output](https://github.com/cms-gem-daq-project/gem-plotting-tools/files/2106260/GEMINIm29L1-timeHistoryAnalyzer.txt)
. Conclusions:

* The number of dead channels found by the new tool is *close to* what's reported in the slides, though not exactly the same.
* The start dates *often* (not always) match what's reported in the slides.
* Most dead channels start with `DeadChannel` straight away
* Some ranges starting with `HotChannel` have `DeadChannel` in the additional `maskReason`s. This is also observed in GEMINIm30L2. Reporting because it's *not* observed for other initial `maskReason`s.

Excerpt from the output:

------

## VFAT 23

|  `ROBstr`  | Last known good   | Range begins     | Range ends   |  #scans  |  Masked%  | Initial `maskReason`   | Other subsequent `maskReason`s   |
|:----------:|:------------------|:-----------------|:-------------|:--------:|:---------:|:-----------------------|:---------------------------------|
|     18     | 2017.10.11.11.24  | 2017.10.13.12.53 | never        |   127    |    100    | HotChannel,FitFailed   |                                  |
|     19     | 2017.10.11.11.24  | 2017.10.13.12.53 | never        |   127    |     0     | DeadChannel            |                                  |
|     20     | 2017.10.11.11.24  | 2017.10.13.12.53 | never        |   127    |     0     | DeadChannel            |                                  |
|     31     | 2017.10.11.11.24  | 2017.10.13.12.53 | never        |   127    |     0     | DeadChannel            |                                  |
|     96     | 2017.06.15.15.10  | 2017.06.16.14.35 | never        |   156    |    58     | HotChannel,HighNoise   | DeadChannel                      |

## Initial maskReason summary

The table below shows the distribution of the initial maskReason for ranges found in each VFAT.
Note that a single range is counted as many times as it has maskReasons.

|    |  HotChannel  |  FitFailed  |  DeadChannel  |  HighNoise  |  HighEffPed  |
|:--:|:------------:|:-----------:|:-------------:|:-----------:|:------------:|
| 0  |      0       |      0      |       2       |      0      |      0       |
| 1  |      0       |      0      |       1       |      0      |      0       |
| 2  |      3       |      3      |       0       |      0      |      0       |
| 3  |      2       |      1      |       0       |      1      |      0       |
| 4  |      0       |      0      |       0       |      0      |      0       |
| 5  |      3       |      5      |      13       |      0      |      0       |
| 6  |      2       |     34      |       0       |      3      |      0       |
| 7  |      25      |      4      |      26       |     14      |      1       |
| 8  |      1       |      1      |       9       |      0      |      0       |
| 9  |      2       |      2      |       1       |      0      |      0       |
| 10 |      5       |      5      |       2       |      0      |      0       |
| 11 |      4       |      4      |       0       |      0      |      0       |
| 12 |      0       |      0      |       0       |      0      |      0       |
| 13 |      2       |      2      |       0       |      0      |      0       |
| 14 |      0       |      0      |       0       |      0      |      0       |
| 15 |      9       |      4      |      15       |      0      |      0       |
| 16 |      2       |      3      |       2       |      0      |      0       |
| 17 |      4       |      2      |       0       |      0      |      0       |
| 18 |      2       |      2      |       0       |      0      |      0       |
| 19 |      6       |      4      |       2       |      0      |      0       |
| 20 |      0       |      0      |       0       |      0      |      0       |
| 21 |      3       |      3      |       0       |      0      |      0       |
| 22 |      6       |      3      |       0       |      0      |      0       |
| 23 |      2       |      1      |       3       |      1      |      0       |

------

## Checklist:

    - Code is well documented
    - Call interface and concepts are not yet


lmoureaux:
The code is now ready for initial review, but please check the *TODO* items in the description.

macros/timeHistoryAnalyzer.py
```diff
@@ -0,0 +1,457 @@
...
+        """
+        numMaskedChannels = np.count_nonzero(self.mask, (0, 1))
+        badScans = np.logical_or(numMaskedChannels == 0,
+                                 numMaskedChannels / 24. / 128 > maxMaskedChannelFraction)
+        self.dates = self.dates[np.logical_not(badScans)]
+        self.mask = self.mask[:,:,np.logical_not(badScans)]
+        self.maskReason = self.maskReason[:,:,np.logical_not(badScans)]
+        self.noise = self.noise[:,:,np.logical_not(badScans)]
+
+if __name__ == '__main__':
```
bdorney: Separate the functions that are part of the API from the actual executable portion as discussed [here](https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/68#issuecomment-373132733).

These could go in either `anautilities.py` or another `library` style file.

macros/timeHistoryAnalyzer.py
```diff
@@ -0,0 +1,457 @@
...
+        self.mask = np.swapaxes(self.mask, 1, 2) # Reorder to [vfat][strip][time]
+        self.maskReason = np.swapaxes(self.maskReason, 1, 2) # Reorder to [vfat][strip][time]
+        self.noise = np.swapaxes(self.noise, 1, 2) # Reorder to [vfat][strip][time]
+
+    def removeBadScans(self, maxMaskedChannelFraction = 0.07):
+        """Finds bad scans and removes them from the data.
+
+        Any scan matching one of the following criteria is considered bad:
+
+        * No channel was masked
```
bdorney: Why is this considered as bad.  A detector with no channel masked is a perfect detector and is what we are striving to build for LS2.  Justify or re-vise.

macros/timeHistoryAnalyzer.py
```diff
@@ -0,0 +1,457 @@
...
+    """
+
+    def __init__(self, inputDir):
+        """Creates a TimeSeriesData object by reading the files located in the
+        inputDir directory.
+
+        The input directory must contain the following files:
+
+        * gemPlotterOutput_mask_vs_scandate.root
+        * gemPlotterOutput_maskReason_vs_scandate.root
```
bdorney: introduce exception handling here so that the user will be given a meaningful error message if `inputDir` is wrong

anaInfo.py
```diff
@@ -61,6 +61,15 @@ class MaskReason:
     HighNoise   = 0x08
     HighEffPed  = 0x10
 
+    @staticmethod
+    def list():
```
bdorney: find a different name for this `list` is too close to the actual built in type, if you prefer `listOfTuples()` but this will cause confusion later on. e.g. `list()` vs `list[]` are too similar.

macros/timeHistoryAnalyzer.py
```diff
@@ -0,0 +1,457 @@
...
+                             maxSkip)
+
+    return list(filter(lambda r: np.count_nonzero(r.noise() < threshold) >= minBadScans,
+                       ranges))
+
+class TimeSeriesData(object):
+    """Holds information about time variation of scan results.
+
+    Each property is stored as a 3D Numpy array with indexes
+    [vfat][strip][time]. The time index -> scan date mapping is exposed in the
```
bdorney: in all cases change `strip` to `chanStripOrPanPin` since this could represent three different quantities and is not always "strip"

macros/timeHistoryAnalyzer.py
```diff
@@ -0,0 +1,457 @@
...
+    date attribute.
+
+    Attributes:
+        dates: Array of strings containing the scan dates.
+        mask: Data for the "mask" property,
+        maskReason: Data for the "maskReason" property,
```
bdorney: numpy array, `array` is something different.

macros/timeHistoryAnalyzer.py
```diff
@@ -0,0 +1,457 @@
...
+
+        for vfat in range(0,24):
+            dirname = 'VFAT%d' % vfat
+            dir_mask = file_mask.Get(dirname)
+            dir_maskReason = file_maskReason.Get(dirname)
+            dir_noise = file_noise.Get(dirname)
+
+            hist_mask = dir_mask.Get("h_ROBstr_vs_scandate_Obsmask_VFAT%d" % vfat)
+            hist_maskReason = dir_maskReason.Get("h_ROBstr_vs_scandate_ObsmaskReason_VFAT%d" % vfat)
+            hist_noise = dir_noise.Get("h_ROBstr_vs_scandate_Obsnoise_VFAT%d" % vfat)
```
bdorney: why bother creating a `TDirectory` object here?  You can get these `TObjects` directly from the `TFile` object without needing to have an additional set of objects. e.g.:

```
            hist_noise = file_noise.Get("VFAT%i/h_ROBstr_vs_scandate_Obsnoise_VFAT%i" %(vfat, vfat))
```

Additionally change `%d` to `%I`; `vfat` is an integer not a floating point number; this will ensure no funny business (e.g. `1.0` vs. `1` is inserted).

macros/timeHistoryAnalyzer.py
```diff
@@ -0,0 +1,457 @@
...
+        import ROOT as r
+        from root_numpy import hist2array
+
+        file_mask = r.TFile('%s/gemPlotterOutput_mask_vs_scandate.root' % inputDir, 'READ')
+        file_maskReason = r.TFile('%s/gemPlotterOutput_maskReason_vs_scandate.root' % inputDir, 'READ')
+        file_noise = r.TFile('%s/gemPlotterOutput_noise_vs_scandate.root' % inputDir, 'READ')
+
+        self.mask = [] # [vfat][time][strip]; warning: reordered after loading
+        self.maskReason = [] # [vfat][time][strip]; warning: reordered after loading
+        self.noise = [] # [vfat][time][strip]; warning: reordered after loading
```
bdorney: Make these structured `numpy` arrays from the start, don't do gymnastics later to change them (lines 314-321).  You can then access them via their type dtype name then.

macros/timeHistoryAnalyzer.py
```diff
@@ -0,0 +1,457 @@
...
+
+    Searches the data for the given vfat and channel for ranges of scans with
+    noise < threshold. During the search, at most maxSkip scans with mask not
+    set can be skipped. Only ranges with more than minBadScans are kept.
+
+    Args:
+        data: The TimeSeriesData object to pull data from
+        vfat: The VFAT to return ranges for
+        channel: The channel to return ranges for
+        threshold: The minimum value of noise for a scan to be "good"
```
bdorney: What is "good" defined as?  Additionally `threshold` is exactly the name of the `TBranch` which tracks scurve_mean value but here it is related to `scurve_sigma`.  Change to be something less confusing.  e.g. `scurveSigmaCut` or `noiseCut` or similar.

macros/timeHistoryAnalyzer.py
```diff
@@ -0,0 +1,457 @@
...
+        """Returns a maskReason bitmask that contains all the maskReaons in the
+        range but not in the first scan"""
+        return self.allMaskReasons() ^ self.initialMaskReason()
+
+    def noise(self):
+        """Returns a Numpy array containing the noise information for scans in
+        the range"""
+        return self._noise[self.start:self.end]
+
+def _findRangesMeta(data, vfat, channel, channelData, maxSkip):
```
bdorney: Why have a leading '_' here?  This is not a class method so having it follow syntax for private doesn't really make sense.

bdorney:
There should be a corresponding update to `README.md` that gives a description of input arguments, outputs, and usage case examples.  And that seems to be what's missing here in this description.  If an expert cannot understand this easily; a `USER` will not be able too.

macros/timeHistoryAnalyzer.py
lmoureaux: It's a common convention in Python for module-private functions

macros/timeHistoryAnalyzer.py
lmoureaux: [`%d` = `%i` = "Signed integer decimal"](https://docs.python.org/2/library/stdtypes.html#string-formatting-operations)

macros/timeHistoryAnalyzer.py
lmoureaux: > Why not make these structured `numpy` arrays from the start

Because then I can't use `hist2array` as easily. I have to do the gymnastics inside the loop or after.

> You can then access them via their type dtype name then

This isn't a list of structs, so I can't see how structured arrays would help (if that's what you're talking about).

lmoureaux:
### Quick answers to select initial comments on the PR description

> What is meant by a range?

A set of successive scans for a given VFAT and channel. I'm not happy with the wording either; please provide a better, preferably short, name if can you have one.

> What is this noise criterion supposed to be identifying? Why introduce a new term that is not present in maskReason?

That's supposed to be what was requested in #88 -- or am I mistaken on the purpose of the noise cuts? If there's not enough detail for me to understand what's asked for, please update the request. It's in particular relevant because of this sentence in #88:

> Caveats is that this needs to be modified to not be sensitive to transient effects

So the algo needs to be modified, which is what I'm trying to achieve here. But it's hard not to break it if I don't understand what it's meant to do.

> This table is what is being requested in #88, e.g. initial maskReason was the first instance that was assigned?

What actually happens is more complicated that channels dying over time (even though the end result is the same). They can appear as dead for some time, be recovered, and then die for good. Such cases exist. Every detected "good->bad" (1) transition is recorded in the table.

(1) What exactly "good" and "bad" mean here depends on the value passed to the `--ranges` option.

> Additionally in the table for this spot can channels have multiple initial maskReasons? If not why not?

When a transition happens because of several `maskReason`s, it gets an entry in each. It's not clear from the context info in #88 what information you want to extract from this table.

Channels with multiple maskReasons appear in multiple bins.

> For example I see VFAT3 has 7 channels marked as HotChannel, FitFailed, and DeadChannel. Are these the same 7 dead channels in each case?

To answer this question one should look at the per-VFAT lists. In this case there are 7 DeadChannel and 7 HotChannel,FitFailed. For info, all of them were already dead at the time of the first scan and none was recovered afterwards.

> you may find them "useless" but they improve readability.

It's a matter of taste. IMHO they add unnecessary visual clutter and in most cases they actually reduce readability (the only exception being deeply nested loops, which are to be avoided anyway).

macros/timeHistoryAnalyzer.py
lmoureaux: Done

lmoureaux:
Expanded the documentation in the first post; it will eventually be included in the repo

macros/timeHistoryAnalyzer.py
```diff
@@ -0,0 +1,467 @@
...
+
+        Args:
+            inputDir: The path to the input directory
+        """
+        import ROOT as r
+        from root_numpy import hist2array
+
+        file_mask = r.TFile('%s/gemPlotterOutput_mask_vs_scandate.root' % inputDir, 'READ')
+        if file_mask.IsZombie():
+            raise IOError('Could not open %s' % file_mask.GetPath())
```
bdorney: Explicitly tell the user with this error message they were supposed give the `$PWD` which is the output of `plotTimeSeries.py`.

lmoureaux:
[Link to the documentation](https://github.com/lmoureaux/gem-plotting-tools/tree/feature_timeHistoryAnalyzer#analyzing-the-time-evolution-of-channels-timehistoryanalyzerpy)

macros/timeHistoryAnalyzer.py
```diff
@@ -0,0 +1,158 @@
...
+        findRangesFct = findRangesMask
+    elif options.ranges == "maskReason":
+        findRangesFct = findRangesMaskReason
+    elif options.ranges == "zeroInputCap":
+        findRangesFct = findRangesZeroInputCap
+        findRangesKwArgs['minNoise'] = options.minNoise
+        findRangesKwArgs['maxNoise'] = options.maxNoise
+    else:
+        print("Error: Invalid argument for --ranges: %s " % options.ranges)
+        sys.exit(os.EX_USAGE)
```
bdorney: How hard would it be to make the default be "all of the above" (lines 56, 58, and 60) and then if the user specifies `--ranges` it's for a specific.  Is this too complex?  

If it is too complex how hard would it be to provide a script which this executes "all of the above" as three separate calls of `timeHistoryAnalyzer.py` similar to [gemPlotterAllChannels.sh](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/macros/gemPlotterAllChannels.sh)?

Is this a "less than a day change?" If yes let's include it in this PR.  If not let's make a future feature request on the issue page.

bdorney: Maybe the script is best, and it writes the output to a file since having this spew multiple tables may get confusion?  Please comment when you have a chance.

macros/timeHistoryAnalyzer.py
```diff
@@ -0,0 +1,158 @@
...
+
+    # Filter if needed
+    if options.onlyCurrent:
+        for vfat in range(24):
+            for stripOrChan in range(128):
+                ranges[vfat][stripOrChan] = list(filter(lambda r: r.end == data.numScans(),
+                                                        ranges[vfat][stripOrChan]))
+
+    # Initialize tables
+    rangesTables = [ [] for vfat in range(24) ]
```
bdorney: A function exists for this, see [getEmptyPerVFATList](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/anautilities.py#L89).  It is not necessary to change but you may find it useful.

requirements.txt
```diff
@@ -5,3 +5,7 @@ numpy>1.7,<1.15; python_version == '2.7'
 numpy>1.7; python_version >= '3.4'
 ## currently conflicts if numpy isn't installed, as it will download it's own version due to steup_requries
 root_numpy
+## 'tabulate' dropped explicit support for Python 2.6 in version 0.8.1
+## Version 0.8.0 isn't in the CHANGELOG
+tabulate>=0.4.4,<0.8; python_version == '2.6'
```
bdorney: In one case you restrict `tabulate` to not be higher than `0.7.7` but here it's less than `0.8`.  Shouldn't these match?

utils/anahistory.py
```diff
@@ -0,0 +1,370 @@
+"""
+Tools to analyze the time evolution of detectors.
+"""
+
+import numpy as _np
+
+class ChannelTimeRange(object):
+    """Represents a range of scans in TimeSeriesData, for a given VFAT and
```
bdorney: Is `range` here the same meaning as it is in the documentation? e.g. a collection of "bad" scans?

macros/timeHistoryAnalyzer.py
lmoureaux: Changing the code would take much less than one day, however it would require significant modifications of the examples.

The main problem is usability:

* The user would have to understand which table come from which `--ranges` value. Since the tables can be large, even a header would be easily missed.
* Printing all tables would quickly fill the terminal buffer.
* The user could wonder why there are overlapping ranges, while it's clear if s.he explicitly asks for another definition

I'm open to do the change.

utils/anahistory.py
lmoureaux: It's a set of consecutive scans, with no special meaning

macros/timeHistoryAnalyzer.py
bdorney: Let's not change the python code.  Instead I think a script `makeAllRanges.sh` or similar would be better that writes the output to a text file; e.g.:

```
timeHistoryAnalyzer.py <args for mask> | tee maskOutput.txt
timeHistoryAnalyzer.py <args for maskReason> | tee maskReasonOutput.txt
timeHistoryAnalyzer.py <args for zeroInputCap> | tee zeroInputCapOutput.txt
```

Or similar.

requirements.txt
lmoureaux: They do match (`next(0.7.7) == 0.8` for `tabulate`). I'll make it more obvious

macros/timeHistoryAnalyzer.py
jsturdy: > The main problem is usability:

Alternatively, why are you `print`ing tables, rather than just `write`ing them to a file?

lmoureaux: Writing to a file isn't necessary atm, >xxx is as easy as -o xxx. If needed later, I'd rather produce trees than writing Markdown tables

PR #114, opened 2018-06-13
Bug Fix: initial guess for scurve mean position was range limited 

bdorney:
## Description

Found that in some cases scurve fitting was hitting the number of max iterations and not converging on the true fit parameter values.  This was especially true when scurve mean position was at low values of charge.  This PR addresses this concern.

Additionally poisson error bars for scurves are now assigned in scurve fit process addressing [this](https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/61#issuecomment-376870516) comment of #61.

Finally additionally debugging info is now provided when `debug` is set to True.  Namely the printing of the converged values after the fitting procedure for a channel completes. 

### Types of changes

## Motivation and Context

Previously the scurve fitting procedure was failing to converge on the true fit parameters due to the initial guess never being close enough to the true value.  The limitation of `stepN < 15` was causing only half the possible values of `scurve_mean` to be scanned.  This was leading to the fit results poorly describing the histogram as shown:

screen shot 2018-06-13 at 10 50 22

Note here the central value (error bar) of the black points represents the scurve mean (sigma) value.  Clearly you can see a large number of scurves are not being fit correctly when the scurve mean position is low.

## How Has This Been Tested?

Re-analysis of scurve data.  After allowing `stepN` to proceed to 30 this enables the CAL_DAC/VCAL value to go from 0 to 256 (previously was limited to 0 to 128). 

After this change you can see the scurve histograms are more well described by the fits:

screen shot 2018-06-13 at 13 23 43

Again the black point's central value (error bar) represents the scurve mean (sigma) value.

## Checklist:

### DIFF
--- a/fitting/fitScanData.py
+++ b/fitting/fitScanData.py
@@ -25,6 +25,7 @@ class ScanDataFitter(DeadChannelFinder):
         self.Nev = ndict()
         self.scanFuncs  = ndict()
         self.scanHistos = ndict()
+        self.scanHistosChargeBins = ndict()
         self.scanCount  = ndict()
         self.scanFitResults   = ndict()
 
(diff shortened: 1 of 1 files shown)


bdorney:
# Issue w/Effective Pedestal

For commit `b0a77b8`. I noticed that the effective pedestal was being erroneously reported.  In some cases by a lot.

Here is distribution of all scurves as `TH1F` objects for `vfat0`:
scurve_pedoverestimated

You can see for all of the `TH1F` objects that at `CAL_DAC = 0 fC` the number of entries is 0.  So the effective pedestal for all these curves should be `0`.

However the fits for these `TH1F` objects look like this:

scurve_fit_func_pedoverestimated

You can see that many of them have `func.Eval(0)` being much larger than 0.  Evidently the fit is not describing well the distributions.  Explicitly the pedestals reported are:

scurve_eff_ped_overestimated

Which is indeed wrong if considering the `TH1F` objects above.

# Effective Pedestal Solution 

I changed the initial guess of the effective pedestal parameter in the `TF1` from being the same as the scurve mean (in reality it never should have been this...) to `0.0`.  Additionally I changed the limits on this fit parameter to a more appropriate `[0, Nev*2]` as the pedestal must be less than or equal to the number of charge injections.

Finally I change the parameter limits on parameter [3] (which is supposed to equal `Nev/2`) to be more constrained around this value.  

Checking these changes one at a time I saw that the evaluation of the effective pedestal improved considerably.  When both changes are applied the result is:

The fits:
scurve_fit_postfix

The effective pedestal distribution:
scurve_ped_postfix

There's still one or two scurves that report the effective pedestal incorrectly...however the situation is greatly improved.  I would be worried about over-constraining this parameter.  In any case `anaUltraScurve.py` has a `--maxEffPedPercent` argument which determines when the effective pedestal should be used to mask a channel.  

Perhaps if others have an additional idea on how to resolve this we can proceed.  But for now performance is greatly increased.

Issue #94, opened 2018-05-15
Docs migration


jsturdy:
## Brief summary of issue
Current README needs to be refactored into a proper `docs` style
README should have info about getting set up, links to contributing guidelines, links to full documentation

### Types of issue
- [x] Cleanup

## Possible Solution (for bugs)
* Create `docs` folder
* Migrate documentation portion of README into this, preferably using rst for compatibility with `sphinx`/`readthedocs` style documentation

mexanick:
Yes, let's use sphinx as a baseline. I'll look into sphinx setup for the reg_utils repo and will provide more detailed feedback and maybe a sort of template here as well. Basically, sphinx is using the ReStructured Text as doc format which is reasonably close to Markdown used in the existing readme, so it would not be a hard task to get the proper documentation with it. We will need a skeleton first and then distribute the existing content around it. Given the completeness of this project documentation in a README, I do not expect a need to write a lot of new stuff. 
After this is done and RPM installation procedure is polished, the README should shrink to a short instruction on getting the code installed and link to the full documentation as @jsturdy suggested.

bdorney:
Was this resolved by #115 and #116?  Or is something more needed?

lmoureaux:
> Was this resolved by #115 and #116? Or is something more needed?

A significant part of the README and many docstrings still have to be ported to `rst`

bdorney:
Okay, thanks for update.  We keep this issue open for now.

PR #115, opened 2018-06-13
Add infrastructure to create API documentation

lmoureaux:
## Description

Based on the `reg_utils` setup with some modifications:

* Added the `napoleon` plugin to parse [Google style](https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments) and Numpy style comments (they are more readable in-source)
* `make html` instead of `make doc` (using the target of `mfSphinx.mk` directly)
* Updated author list (from the contributors list on Github)

Generated documentation is sparse and all formatting is lost. It will be necessary to format it in a format sphinx can understand, but this is out of the scope of this PR.

The generated docs can be found [here](http://homepage.iihe.ac.be/~lmoureau/GEM/sphinx/html/index.html).

### Types of changes

## Motivation and Context

#94

## How Has This Been Tested?

Documentation was generated on `lxplus` inside a `virtualenv`. Waiting for the first Travis build to finish.

## Checklist:


doc/conf.py
```diff
@@ -0,0 +1,246 @@
...
+# If true, an OpenSearch description file will be output, and all pages will
+# contain a <link> tag referring to it.  The value of this option must be the
+# base URL from which the finished HTML is served.
+#html_use_opensearch = ''
+
+# This is the file name suffix for HTML files (e.g. ".xhtml").
+#html_file_suffix = None
+
+# Output file base name for HTML help builder.
+htmlhelp_basename = 'gem-plotting-tools-doc'
```
jsturdy: `gem-plotting-tools` should bechanged to `gemplotting`, reflecting the actual name of the `python` package

doc/conf.py
```diff
@@ -0,0 +1,246 @@
...
+#'pointsize': '10pt',
+
+# Additional stuff for the LaTeX preamble.
+#'preamble': '',
+}
+
+# Grouping the document tree into LaTeX files. List of tuples
+# (source start file, target name, title, author, documentclass [howto/manual]).
+latex_documents = [
+  ('index', 'gem-plotting-tools.tex', u'reg\\_utils Documentation',
```
jsturdy: * `gem-plotting-tools` should bechanged to `gemplotting`, reflecting the actual name of the `python` package
* residual `reg_utils` should be changed as well

doc/conf.py
```diff
@@ -0,0 +1,246 @@
...
+# If false, no module index is generated.
+#latex_domain_indices = True
+
+
+# -- Options for manual page output --------------------------------------------
+
+# One entry per manual page. List of tuples
+# (source start file, name, description, authors, manual section).
+man_pages = [
+    ('index', 'gem-plotting-tools', u'gem-plotting-tools Documentation',
```
jsturdy: `gem-plotting-tools` should bechanged to `gemplotting`, reflecting the actual name of the `python` package

doc/conf.py
```diff
@@ -0,0 +1,246 @@
...
+#man_show_urls = False
+
+
+# -- Options for Texinfo output ------------------------------------------------
+
+# Grouping the document tree into Texinfo files. List of tuples
+# (source start file, target name, title, author,
+#  dir menu entry, description, category)
+texinfo_documents = [
+  ('index', 'gem-plotting-tools', u'gem-plotting-tools Documentation',
```
jsturdy: `gem-plotting-tools` should bechanged to `gemplotting`, reflecting the actual name of the `python` package

doc/conf.py
```diff
@@ -0,0 +1,246 @@
...
+   u'Mykhailo Dalchenko, Evaldas Juska, RobertKing, Andrew Peck, Jared Sturdy', 'gem-plotting-tools', 'One line description of project.',
```
jsturdy: *`gem-plotting-tools` should bechanged to `gemplotting`, reflecting the actual name of the `python` package
* Contributors list here is unmodified from where it was copied

doc/conf.py
```diff
@@ -0,0 +1,246 @@
+# -*- coding: utf-8 -*-
+#
+# gem-plotting-tools documentation build configuration file, created by
```
jsturdy: `gem-plotting-tools` should bechanged to `gemplotting`, reflecting the actual name of the `python` package

doc/index.rst
```diff
@@ -0,0 +1,25 @@
+.. gem-plotting-tools documentation master file, created by
```
jsturdy: `gem-plotting-tools` should bechanged to `gemplotting`, reflecting the actual name of the `python` package

doc/index.rst
```diff
@@ -0,0 +1,25 @@
...
+   sphinx-quickstart on Wed May 23 14:52:11 2018.
+   You can adapt this file completely to your liking, but it should at least
+   contain the root `toctree` directive.
+
+gem-plotting-tools package
```
jsturdy: `gem-plotting-tools` should bechanged to `gemplotting`, reflecting the actual name of the `python` package

lmoureaux:
Despite Travis reporting success, building the documentation on Travis **failed** because of `napoleon` wasn't available. Under investigation.

lmoureaux:
* Travis build still not fixed; will need to do it in Docker image
* Documentation structure improved
* I'm not sure where `buildMapFiles` should appear. Currently it's not documented anywhere

mexanick:
@lmoureaux you don't need to build the docs in docker container unless you're importing some other GEM modules (like some CDLLs). Even in that case you can wrap in mock import the autodoc importing. From the warnings which travis provided it seems you only miss numpy, so installing it under your venv should help

jsturdy:
also, i suspect a typo in one of the requirements files, as the package `root-numpy` doesn't exist (it's `root_numpy`)

lmoureaux:
> you don't need to build the docs in docker container unless you're importing some other GEM modules

If any module has `import ROOT` or `import root_numpy`, I'll need to install it on the machine. Since it's already done in the Docker image, it's the easiest way (and additionally Debian packages for ROOT are terribly outdated)

mexanick:
Then just run smth similar to the command below in .travis.yml instead of your `make html`. You'll need to populate the command with relevant python packages though
```
    - docker exec -ti $DOCKER_CONTAINER_ID /bin/bash -ec "cd /home/daqbuild; virtualenv building -p python --no-site-packages; . building/bin/activate; cd ${REPO_NAME}; pip install pip==9.0.3; pip install --upgrade setuptools; pip install sphinx; pip install python/reg_interface/rpm/*.tar.gz; export REG_XML_PATH=$PWD; export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/opt/rwreg/lib:/opt/wiscrpcsvc/lib; make doc"
```

jsturdy:
For "complicated" commands running inside the `docker` container, as mentioned previously to @mexanick, it will probably be best to write a script that lives in `.ci`, as this has the added benefit that it would work in a migration to `gitlab-ci`
Not a prereq for this PR, but a consideration moving forward

lmoureaux:
Still failing :disappointed:

Can't install `root_numpy` because it doesn't find `root-config` in `$PATH` and `$ROOTSYS` is not set. Where is ROOT in the Docker image? It's called `devtoolsroot` so it must be there, right?

jsturdy:
yes, there are various versions of `root` installed in `/opt/root/<root version>`
* for `cc7` images see: https://gitlab.cern.ch/cms-gem-daq-project/gemdaq_ci_worker/-/jobs/1344875
* for `slc6` images see: https://gitlab.cern.ch/cms-gem-daq-project/gemdaq_ci_worker/-/jobs/1344871

lmoureaux:
Building the documentation on Docker requires installing the dependencies (`gempython`) there. I can `wget` the package similar to [`setup_gemdaq.sh`](https://github.com/cms-gem-daq-project/sw_utils/blob/develop/scripts/setup_gemdaq.sh), but I'd like a more robust solution.

Is it acceptable to leave the Docker build fail (ie as it is now) until proper `pip` or `rpm` repos are set up?

jsturdy:
> Is it acceptable to leave the Docker build fail (ie as it is now) until proper pip or rpm repos are set up?

Ideally, the docs will build *without* requiring any dependencies be installed, but can be mentioned in the docs somewhere that "version `X.Y.Z` of `<package>` (for which docs are being built) depends on API version `A.B.C` of `<required package>`" (@mexanick had set things up like this such that the doc building performs a basic test of the package, by trying all the `import`s, which is a useful minimal check of the package).
However, at some point even the unit testing will be done in `travis`/`gitlab-ci` and this will require their installation, so I don't know whether it will be beneficial to do things the "right" way... requires further thought.

All that to say, your `travis` isn't "failing" per se in any event (because the error is only in the `after_success`, so it is an acceptable solution for now (until we get docs publishing done through the CI job at least, because then we will need the docs to build properly :-) )

lmoureaux:
> Ideally, the docs will build without requiring any dependencies be installed

The problem is that Sphinx `import`s the packages to be documented, and therefore any failing `import` statement prevents the docs from being built. Since it's perfectly legitimate to have `import` statements at module scope, I don't think there's much we can do except installing the deps.

> it is an acceptable solution for now

Removing the WIP tag to not prevent merge.

PR #116, opened 2018-06-20
Docs migration (first round: mainly gemPlotter.py)

lmoureaux:
## Description

There are three main changes in this PR:

* Modified some scripts to include:
```
if __name__ == '__main__':
```
It's necessary for Sphinx to import them correctly.

* Added rudimentary docstrings to all scripts
* Ported the documentation of `gemPlotter.py` to a docstring. It can be built as part of the larger HTML documentation or as a stand-alone `man` page. In the future, the latter could be included in the `pip` and `rpm` packages.

The README will not be modified as long as the docs aren't hosted centrally.

I'm not yet sure about how to organize the documentation, but this can easily be changed later.

### Types of changes

## Motivation and Context

Use a proper tool for documentation, Sphinx.

## How Has This Been Tested?

Docs were built and published [here](http://homepage.iihe.ac.be/~lmoureau/GEM/sphinx/html/index.html). (Note that this can be ahead by a few commits.)

### Screenshots

`man` displaying the `man` page:
screenshot_20180620_172215

`lynx` displaying the same in HTML:
screenshot_20180620_172439

## Checklist:

### DIFF
--- a/anaSBitThresh.py
+++ b/anaSBitThresh.py
@@ -1,4 +1,10 @@
 #!/bin/env python
+
+"""
+anaSBitThresh
+=============
+"""
+
 import os
 import sys
 from optparse import OptionParser
--- a/anaSBitThresh.py
+++ b/anaSBitThresh.py
@@ -5,101 +5,102 @@ anaSBitThresh
 =============
 """
 
-import os
-import sys
-from optparse import OptionParser
-from gempython.utils.nesteddict import nesteddict as ndict
+if __name__ == '__main__':
+    import os
+    import sys
+    from optparse import OptionParser
+    from gempython.utils.nesteddict import nesteddict as ndict
 
-from gempython.gemplotting.utils.anaoptions import parser
+    from gempython.gemplotting.utils.anaoptions import parser
 
-parser.add_option("--maxNoiseRate", type="float", dest="maxNoiseRate", default=0,
-                  help="Max Noise Rate allowed in Hz", metavar="maxNoiseRate")
-parser.set_defaults(outfilename="SBitRatePlots.root")
+    parser.add_option("--maxNoiseRate", type="float", dest="maxNoiseRate", default=0,
+                    help="Max Noise Rate allowed in Hz", metavar="maxNoiseRate")
+    parser.set_defaults(outfilename="SBitRatePlots.root")
 
-(options, args) = parser.parse_args()
-filename = options.filename[:-5]
-os.system("mkdir " + filename)
+    (options, args) = parser.parse_args()
+    filename = options.filename[:-5]
+    os.system("mkdir " + filename)
 
-print filename
-outfilename = options.outfilename
+    print filename
+    outfilename = options.outfilename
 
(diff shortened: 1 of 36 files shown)


macros/gemPlotter.py
```diff
@@ -1,18 +1,288 @@
 #!/bin/env python
 
+r"""
```
bdorney: what does this stray `r` character do here?

bdorney:
Related too https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/94

macros/gemPlotter.py
lmoureaux: It's a marker for a "raw" string, just like "u" is for Unicode. Escape sequences in raw strings are not interpreted (so `r"\" == "\\"`)

bdorney: Could you add a comment just above describing this to make sure it's not overlooked and removed by someone who doesn't recognise it (e.g. someone like me :D )

jsturdy: I'm generally opposed to commenting things for the sole purpose of telling another developer what the correct syntax is...

anaSBitThresh.py
```diff
@@ -1,99 +1,106 @@
...
-            print 'vt1 for VFAT%i found'%vfat
-            vt1[vfat]=vRate[vfat].GetBinCenter(binX)
-            break
+
+"""
+anaSBitThresh
+=============
+"""
+
+if __name__ == '__main__':
```
jsturdy: are you migrating all the "executable" only scripts to this calling syntax?
if so, i'd wait for them all to be migrated before this PR (I only see this widespread change in a couple of files, though it could be that it has already been implemented in all other tools)

mexanick: yeah, this should be done for all the executables

lmoureaux: It's done in all files where it was missing (except if I missed some).

jsturdy: ok, i suppose sphinx would let you know in the event one was missed :-)

lmoureaux: > i suppose sphinx would let you know in the event one was missed :-)

Not if I missed it completely and didn't even add a documentation entry for it (Sphinx is only semi-automatic, the structure is defined by hand and the contents are extracted).

macros/gemPlotter.py
lmoureaux: It does deserve an entry in the yet-to-be-written developer docs, though

anaSBitThresh.py
jsturdy: OK, if you can do a quick verification, something like:
`find /path/to/scripts -type f -iname '*.py' -print0 -exec fgrep '__main__' {} \'`
`find /path/to/macros -type f -iname '*.py' -print0 -exec fgrep '__main__' {} \'`
and see if you miss anything.  if not, i'll sign off on it

lmoureaux: No script missing

PR #119, opened 2018-06-21
Bug Fix:  Bad histo range and superfluous warning message printing in anaUltraScurve.py

bdorney:
## Description

When not supplying the `--isVFAT3` option (e.g. for v2b electronics) the y-axis range of histograms made by `anaUltraScurve.py` was incorrect (e.g. min = max and max = min).  Additionally superfluous message printing during fitting due to poor choice of parameter limits is addressed.

### Types of changes

## Motivation and Context

Couldn't analyze v2b electronics.  Annoyed by meaningless message spam.

## How Has This Been Tested?

Re-analyzed old data.

scurve_burned_vfat2_by_vfatchannels

## Checklist:

### DIFF
--- a/anaUltraScurve.py
+++ b/anaUltraScurve.py
@@ -154,7 +154,7 @@ if __name__ == '__main__':
                       help="Physical filename of a custom, non-default, channel mapping (optional)", metavar="extChanMapping")
     parser.add_option("-f", "--fit", action="store_true", dest="performFit",
                       help="Fit scurves and save fit information to output TFile", metavar="performFit")
-    parser.add_option("--isVFAT3", action="store_true", dest="isVFAT3",
+    parser.add_option("--isVFAT3", action="store_true", dest="isVFAT3", default=False,
                       help="Provide this argument if input data was acquired from vfat3", metavar="isVFAT3")
     parser.add_option("--IsTrimmed", action="store_true", dest="IsTrimmed",
                       help="If the data is from a trimmed scan, plot the value it tried aligning to", metavar="IsTrimmed")
--- a/anaUltraScurve.py
+++ b/anaUltraScurve.py
@@ -215,17 +215,24 @@ if __name__ == '__main__':
 
     # Initialize distributions
     for vfat in range(0,24):
+        if options.isVFAT3:
+            yMin_Charge = calDAC2Q_Slope[vfat]*255.5+calDAC2Q_Intercept[vfat]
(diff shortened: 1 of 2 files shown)


anaUltraScurve.py
```diff
@@ -154,7 +154,7 @@ def plotAllSCurvesOnCanvas(vfatHistos, vfatHistosPanPin2=None, obsName="scurves"
                       help="Physical filename of a custom, non-default, channel mapping (optional)", metavar="extChanMapping")
     parser.add_option("-f", "--fit", action="store_true", dest="performFit",
                       help="Fit scurves and save fit information to output TFile", metavar="performFit")
-    parser.add_option("--isVFAT3", action="store_true", dest="isVFAT3",
+    parser.add_option("--isVFAT3", action="store_true", dest="isVFAT3", default=False,
```
jsturdy: Why set a default here? `store_true` defaults to `False` automatically

bdorney: It does not, it defaults to None.  This was found when passing it to the instance of ScanDataFitter and caused an exception to be generated due to a NoneType

jsturdy: very strange... (strange that `None` and `False` if used as boolean variables don't behave the same for you)
which line is the failure traced to?

bdorney: It shows up right here:

https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/a13112a626b7b532835a5f0b0dcc497b7f47e3ab/anaUltraScurve.py#L296-L300

If the user supplies `--isVFAT3` then `options.isVFAT3` here gives `True` to the constructor of `ScanDataFitter` however if the argument is not supplied and the default is not set `None` is given to the constructor which then throws an error if suplying the `--debug` option because:

https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/a13112a626b7b532835a5f0b0dcc497b7f47e3ab/fitting/fitScanData.py#L226-L241

bdorney: Setting the `default` to `False` resolved this issue.

jsturdy: ah, it's because you're using an `integer` string format for a `boolean` type (works for `True/False` but not, by design, `None`), formatting it as a `boolean` (`%r`) or casting the value to a `bool` in your current implementation would work in all cases

bdorney:
> ah, it's because you're using an integer string format for a boolean type (works for True/False but not, by design, None), formatting it as a boolean (%r) or casting the value to a bool in your current implementation would work in all cases

I'll resolve this in a later PR.  Thanks for the feedback.

Issue #138, opened 2018-07-17
Include man pages in the package


lmoureaux:
## Brief summary of issue

gem-plotting-tools has a nice collection of man pages that can be built using `make man`. However, they are not included in the `pip` package (didn't check the `rpm`). I propose to include them.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

## Expected Behavior

* Man pages are installed along with the software
* If needed, the setup script updates the `MANPATH`

## Current Behavior

Man pages are not distributed.

## Context (for feature requests)

It's common practice among Linux users to type `man command` when looking for information about `command`. We could support this if we installed our man pages.

jsturdy:
Agree 100%
If you can focus on setting up the `rpm` inclusion that would be great.
I don't think it makes sense to add them to a `pip` package as `man` pages are system installed, and this will be taken care of by the `rpm`
For `pip` packages, we'll rely on the rtd site

PR #139, opened 2018-07-19
Install the man pages

lmoureaux:
## Description

This PR modifies the Makefile and setup script to install the man pages. They are installed at `$BASE/share/man`, where `$BASE` is as follows:

* `$VIRTUAL_ENV` when installed within a virtualenv
* `/usr` when installed system-wide

Man pages in the virtualenv can be accessed by setting the correct `$MANPATH`.

[outdated]
*A side-effect of this change is that I figured out how to modify the default `setuptools` install location. All scripts and macros are now installed in `$BASE/bin`. This means that `virtualenv` can take care of updating the `$PATH` (system-wide installs still need a setup script).*
[/outdated]

Another side-effect is that I corrected the license information in the package metadata. It was saying "MIT" while the `LICENSE` file always contained the text of GPLv3 (there is no way to change back to MIT without contacting every single contributor).

The setup script and developer documentation will need to be updated to use the `$MANPATH`.

### Downsides

* The `man` pages produced when the package is not installed may not be completely accurate. This is a problem for `make rpm` since it doesn't work inside a `virtualenv`.
* [outdated]*Macros and scripts are present twice in the package: under `bin/` and under `gempython/...`. I couldn't figure out why.*[/outdated]
* CI is broken :cry:

As a result of these downsides and my inability to fix them shortly, this PR is provided more as a list of commits to be cherry-picked.

### Types of changes

## Motivation and Context

* #138
* [outdated] *Comment in `setup.py`:*
   ```
   would prefer to use this for executables, if one can control the install location
   to be user defined rather than /usr/bin
   ```
   *The preferred method (native support from `setuptools`) is now used.* [/outdated]

## How Has This Been Tested?

* The `pip` and `rpm` packages were built, the former from a `virtualenv` and the latter on `gem904daq01`.
* The `pip` package was installed; scripts and man pages are correctly installed and have the correct permissions.
* It was checked that the `rpm` package contains all files.
* The `man` pages built in each environment were (quickly) checked.

Since I don't have access to a `rpm`-based machine and I don't want to install one, I couldn't check that the `rpm` package installs cleanly.

## Checklist:

### DIFF
--- a/pkg/setup.cfg
+++ b/pkg/setup.cfg
@@ -4,7 +4,7 @@ name = gempython_gemplotting
 author = GEM Online Systems Group
 author_email = cms-gem-online-sw@cern.ch
 summary = __summary__
-license = MIT
+license = GNU General Public License v3 (GPLv3)
 description-file = gempython/gemplotting/README.md
 home-page = https://cms-gem-daq-project.github.io/gem-plotting-tools
 project_urls =
@@ -20,7 +20,8 @@ classifier =
     Topic :: Data Acquisition
     Topic :: Scientific
     Topic :: Utilities
-    License :: OSI Approved :: MIT
+    License :: OSI Approved :: GNU General Public License (GPL)
+    License :: OSI Approved :: GNU General Public License (GPLv3)
     Operating System :: POSIX
     Operating System :: Unix
     Programming Language :: Python [files]
--- a/pkg/setup.py
+++ b/pkg/setup.py
@@ -60,11 +60,11 @@ def getdatafiles():
     # Man files
     man_sections = {}
     for file in listdir(mandir):
-        print('Found man page: {}'.format(file))
+        print('Found man page: %s' % file)
         section = file.split('.')[-2]
         man_sections[section] = man_sections.get(section, []) + [join(mandir, file)]
     for section in man_sections:
-        data_files.append(('share/man/man{}'.format(section), man_sections[section]))
+        data_files.append(('share/man/man%s' % section, man_sections[section]))
 
     return data_files
 
--- a/pkg/setup.py
+++ b/pkg/setup.py
@@ -75,5 +75,5 @@ setup(name             = '__packagename__',
           'setuptools>=25.0'
       ],
       install_requires = getreqs(),
-      license          = 'MIT',
+      license          = 'GNU General Public License v3 (GPLv3)',
 )
--- a/pkg/setup.py
+++ b/pkg/setup.py
@@ -61,7 +61,7 @@ def getdatafiles():
     man_sections = {}
     for file in listdir(mandir):
         print('Found man page: %s' % file)
-        section = file.split('.')[-1]
+        section = file.split('.')[-2]
         man_sections[section] = man_sections.get(section, []) + [join(mandir, file)]
     for section in man_sections:
         data_files.append(('share/man/man%s' % section, man_sections[section]))
--- a/pkg/setup.py
+++ b/pkg/setup.py
@@ -7,6 +7,7 @@ from os.path import isfile,join
 scriptdir  = 'gempython/scripts'
 scriptpath = '/opt/cmsgemos/bin'
 scripts    = listdir(scriptdir)
+mandir     = 'man'
 
 def readme():
     with open('README.md') as f:
@@ -53,6 +54,20 @@ __builddate__='{6:s}'
 """.format(__version__,__release__,__buildtag__,__gitrev__,__gitver__,__packager__,__builddate__))
     return '{0:s}'.format(__version__)
 
+def getdatafiles():
+    data_files = []
+
+    # Man files
+    man_sections = {}
+    for file in listdir(mandir):
+        print('Found man page: %s' % file)
+        section = file.split('.')[-1]
+        man_sections[section] = man_sections.get(section, []) + [join(mandir, file)]
+    for section in man_sections:
+        data_files.append(('share/man/man%s' % section, man_sections[section]))
+
+    return data_files
+
 setup(name             = '__packagename__',
       version          = getVersion(),
       # use_scm_version  = True,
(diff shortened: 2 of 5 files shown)


pkg/installrpm.sh
```diff
@@ -23,8 +23,8 @@ cat INSTALLED_FILES.backup|egrep -v 'gempython/__init__.py*' > INSTALLED_FILES
 # set permissions
 cat <<EOF >>INSTALLED_FILES
 %attr(0755,root,root) /opt/cmsgemos/bin/*.py
-%attr(0755,root,root) /usr/lib/python*/site-packages/gempython/scripts/*.py
-%attr(0755,root,root) /usr/lib/python*/site-packages/gempython/gemplotting/macros/*.py
+%attr(0755,root,root) /opt/cmsgemos/lib/python*/site-packages/gempython/scripts/*.py
```
jsturdy: This was also explicitly avoided in the initial setup, as in future versions this will be replaced with system site-packages location macros provided by `rpm` (currently only works for `cc7`)

pkg/setup.cfg
```diff
@@ -4,7 +4,7 @@ name = gempython_gemplotting
 author = GEM Online Systems Group
 author_email = cms-gem-online-sw@cern.ch
 summary = __summary__
-license = MIT
+license = GNU General Public License v3 (GPLv3)
```
jsturdy, line 7: License changes *must* be discussed

pkg/setup.cfg
```diff
@@ -20,7 +20,8 @@ classifier =
     Topic :: Data Acquisition
     Topic :: Scientific
     Topic :: Utilities
-    License :: OSI Approved :: MIT
+    License :: OSI Approved :: GNU General Public License (GPL)
```
jsturdy, line 23: Same

pkg/setup.py
```diff
@@ -32,6 +38,20 @@ def getreqs():
         reqs = f.readlines()
         return [x.strip() for x in reqs]
 
+def getdatafiles():
+    data_files = []
+
+    # Man files
+    man_sections = {}
+    for file in listdir(mandir):
+        print('Found man page: %s' % file)
```
jsturdy, line 63: Also, prefer consistent string formatting (avoid `%` formatting in favour of `.format()` formatting) for all current and future contributions, `CONTRIBUTING.md` will be updated to reflect this and a few other changes

jsturdy (changes requested):
There are far too many changes in this PR that should have been discussed in #138 before being implemented

> A side-effect of this change is that I figured out how to modify the default setuptools install location. All scripts and macros are now installed in $BASE/bin. This means that virtualenv can take care of updating the $PATH (system-wide installs still need a setup script).

This was by design, avoiding putting things in `macros` into `/opt/cmsgemos/bin`

> Another side-effect is that I corrected the license information in the package metadata. It was saying "MIT" while the LICENSE file always contained the text of GPLv3 (there is no way to change back to MIT without contacting every single contributor).

This is probably the discussion that I mention inline.  I would have to go back in time to figure out why a given license was (probably arbitrarily) chosen for this project, and probably raise it as a discussion, (contacting contributors on this project is pretty trivial)

pkg/setup.py
bdorney: For my own ignorance, why is `.format()` viewed as superior?

pkg/setup.cfg
lmoureaux: This is not a license change, this is reflecting the fact that the current license *is* GPLv3. If you want to contact all copyright holders, fine; but until you get answers the license remains GPLv3.

pkg/installrpm.sh
lmoureaux: > system site-packages location macros provided by `rpm`

Could you provide more details about this? A link to the `rpm` doc about this would be great

lmoureaux:
> avoiding putting things in macros into `/opt/cmsgemos/bin`

May I ask what's the rationale for the difference between `scripts` and `macros` in this project?

pkg/setup.py
jsturdy: I don't know about *superior*... it is the standard that was introduced to eventually replace `%` formatting, though I suspect that `%` formatting will stick around for quite some time...
Both have a few advantages, though I think `.format()` has more.
The primary thing here (in this comment) is maintaining consistency

lmoureaux:
> I would have to go back in time to figure out why a given license was (probably arbitrarily) chosen for this project 

Commit 3d58f5d2 by... @jsturdy. The commit message ("Initial commit") isn't helpful in telling how the license was chosen.

FYI:
* CMSSW uses [Apache 2.0](http://www.apache.org/licenses/LICENSE-2.0).
* XDAQ is under a 3-clause [BSD-style license](https://svnweb.cern.ch/trac/cmsos/browser/trunk/LICENSE)

jsturdy:
> May I ask what's the rationale for the difference between scripts and macros in this project?

Mainly, it is to isolate the key executable parts (analyzing scans, which should go in `/opt/cmsgemos/bin`) from random plotting and style macros (can still be on the `PATH` for ease of use, provided `PATH` is modified to include the `macros` directory)

Further down the road, much of this will be replaced with `console_scripts` and `entry_points`, but that is a minor priority

lmoureaux:
Removed the controversial move to a more standard behavior using standard tools; the `rpm` package will now install `man` pages in `/usr/share`. Updated the first post accordingly.

lmoureaux:
Comments have been addressed and I'm waiting for another review.

jsturdy (approved):
RPM requirements:
```bash
rpm -q --requires -p rpm/gempython_gemplotting-1.0.0-0.0.6.30.dev5.087aeffgit.centos7.python2.7.noarch.rpm
/bin/env
/bin/sh
/usr/bin/env
cmsgemos_gempython >= 0.3.1
config(gempython_gemplotting) = 1.0.0-0.0.6.30.dev5.087aeffgit.centos7.python2.7
numpy >= 1.7
python >= 2.6
python(abi) = 2.7
root_numpy >= 4.7
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PartialHardlinkSets) <= 4.0.4-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
tabulate >= 0.4.4
rpmlib(PayloadIsXz) <= 5.2-1
```

* looks like some mixed shebangs are used, these should be cleaned up in the future
* also in the future, would like to be able to provide these to `bdist_rpm` based on the `python` version, but I believe this would be a significant redesign, moving the majority of the `setup.cfg` stuff into functions imported in `setup.py`

PR #147, opened 2018-08-08
Modification in fitting routine used in anaUltraScurve.py

brentstone:
The analyzer script ```anaUltraScurve.py``` fits the Scurves with a class ```ScanDataFitter```, defined in ```fitting/fitScanData.py```. A modification is made with regard to the initial guess of the parameter ```p1```, such that this guess will always fall within the prescribed bounds of the fit on ```p1```.

## Description

The choice of initial guess in the routine for ```p1``` is a the absolute value of the product of ```CAL_DACm``` and a Gaussian random variable with ```mean``` 10 and ```sigma``` 5, which, in some cases, fell below the lower bound of the fit for VFAT3, especially if ```CAL_DACb``` is large. In these cases, the code would then enter an infinite ```while``` loop. The routine is modified such that the initial guess is now the value of the lower bound plus the previously prescribed guess: ```p1_guess --> p1_low + p1_guess```.

### Types of changes

## Motivation and Context

When analyzing Scurves with v3 electronics, the fits on one vfat in particular would not converge and the code got caught in an infinite loop.

## How Has This Been Tested?

This code was tested on 3 sets of Scurves from the following chambers: ```X-S-CERN-0012```, ```X-S-CERN-0011```, (both ```v3``` chambers) and the ```v2b``` chamber ```VII-S-CERN-0001```. The fits converge for all vfats in each case.

The fit results for ```X-S-CERN-0012``` are located in ```/data/bigdisk/GEM-Data-Taking/GE11_QC8/GE11-X-S-CERN-0012/scurve/2018.08.08.09.35/SCurveData/```.

The fit results for ```X-S-CERN-0011``` are located in ```/data/bigdisk/GEM-Data-Taking/GE11_QC8/GE11-X-S-CERN-0011/scurve/2018.08.07.19.00/```.

The fit results for ```VII-S-CERN-0001``` are located in ```/afs/cern.ch/user/b/bstone/public/v2b_scurvefit_test/```.

The ```Summary.png``` and ```fitSummary.png``` files can also be viewed on the Mattermost here: https://mattermost.web.cern.ch/cms-gem-ops/pl/4cm7xex58fnttmfn174q5z84th

## Checklist:

### DIFF
--- a/fitting/fitScanData.py
+++ b/fitting/fitScanData.py
@@ -277,7 +277,10 @@ class ScanDataFitter(DeadChannelFinder):
 
                     # Provide an initial guess
                     init_guess_p0 = self.calDAC2Q_m[vfat]*(8+stepN*8)+self.calDAC2Q_b[vfat]
-                    init_guess_p1 = abs(self.calDAC2Q_m[vfat]*rand)  #self.calDAC2Q_m[vfat] might be negative (e.g. VFAT3 case)
+                    if self.isVFAT3:
+                        init_guess_p1 = self.calDAC2Q_m[vfat]*(256)+self.calDAC2Q_b[vfat]+abs(self.calDAC2Q_m[vfat]*rand)  #self.calDAC2Q_m[vfat] might be negative (e.g. VFAT3 case)
+		    else:
+			init_guess_p1 = abs(self.calDAC2Q_m[vfat]*rand)
                     init_guess_p2 = 0.
                     init_guess_p3 = self.Nev[vfat][ch]/2.
 
--- a/fitting/fitScanData.py
+++ b/fitting/fitScanData.py
@@ -276,11 +276,11 @@ class ScanDataFitter(DeadChannelFinder):
                     #if (self.calDAC2Q_m[vfat]*(rand)+self.calDAC2Q_b[vfat]) < 0: continue
 
                     # Provide an initial guess
-                    init_guess_p0 = self.calDAC2Q_m[vfat]*(8+stepN*8)+self.calDAC2Q_b[vfat]
+                    init_guess_p0 = self.calDAC2Q_m[vfat]*(8+stepN*8)+self.calDAC2Q_b[vfat] 
                     if self.isVFAT3:
                         init_guess_p1 = self.calDAC2Q_m[vfat]*(256)+self.calDAC2Q_b[vfat]+abs(self.calDAC2Q_m[vfat]*rand)  #self.calDAC2Q_m[vfat] might be negative (e.g. VFAT3 case)
-		    else:
-			init_guess_p1 = abs(self.calDAC2Q_m[vfat]*rand)
+                    else:
+                        init_guess_p1 = abs(self.calDAC2Q_m[vfat]*rand)
                     init_guess_p2 = 0.
                     init_guess_p3 = self.Nev[vfat][ch]/2.
 
@@ -338,12 +338,14 @@ class ScanDataFitter(DeadChannelFinder):
 
                     # Fit
                     fitResult = self.scanHistos[vfat][ch].Fit('myERF','SQ')
+                    print 'fitResult: ', fitResult
                     fitEmpty = fitResult.IsEmpty()
                     if fitEmpty:
                         fitTF1.SetLineColor(kOrange-2)
                         # Don't try to fit empty data again
                         break
                     fitValid = fitResult.IsValid()
+                    print 'fitValid? ', fitValid
                     if not fitValid:
                         continue
                     fitChi2 = fitTF1.GetChisquare()
--- a/fitting/fitScanData.py
+++ b/fitting/fitScanData.py
@@ -337,7 +337,7 @@ class ScanDataFitter(DeadChannelFinder):
                                     )
 
                     # Fit
-                    fitResult = self.scanHistos[vfat][ch].Fit('myERF','SQ') 
+                    fitResult = self.scanHistos[vfat][ch].Fit('myERF','SQ')
                     fitEmpty = fitResult.IsEmpty()
                     if fitEmpty:
                         fitTF1.SetLineColor(kOrange-2)
--- a/fitting/fitScanData.py
+++ b/fitting/fitScanData.py
@@ -337,15 +337,13 @@ class ScanDataFitter(DeadChannelFinder):
                                     )
 
                     # Fit
-                    fitResult = self.scanHistos[vfat][ch].Fit('myERF','SQ')
-                    print 'fitResult: ', fitResult
+                    fitResult = self.scanHistos[vfat][ch].Fit('myERF','SQ') 
                     fitEmpty = fitResult.IsEmpty()
                     if fitEmpty:
                         fitTF1.SetLineColor(kOrange-2)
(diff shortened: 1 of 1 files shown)


bdorney:
Can you upload the png files to mattermost and then provide the hyperlink to the post here?

We would find:

- Summary.png
- fitSummary.png

Of interest.

fitting/fitScanData.py
```diff
@@ -277,7 +277,10 @@ def fit(self, debug=False):
 
                     # Provide an initial guess
                     init_guess_p0 = self.calDAC2Q_m[vfat]*(8+stepN*8)+self.calDAC2Q_b[vfat]
-                    init_guess_p1 = abs(self.calDAC2Q_m[vfat]*rand)  #self.calDAC2Q_m[vfat] might be negative (e.g. VFAT3 case)
+                    if self.isVFAT3:
+                        init_guess_p1 = self.calDAC2Q_m[vfat]*(256)+self.calDAC2Q_b[vfat]+abs(self.calDAC2Q_m[vfat]*rand)  #self.calDAC2Q_m[vfat] might be negative (e.g. VFAT3 case)
+		    else:
+			init_guess_p1 = abs(self.calDAC2Q_m[vfat]*rand)
```
bdorney: Is the `else` in line 282 supposed to be paired with the `if` at line 280?  Seems you have a different indentation here.  Perhaps your editor is inserting `\t` characters when pressing the `tab` key instead of inserting four `\s` characters (e.g. space).

fitting/fitScanData.py
```diff
@@ -277,7 +277,10 @@ def fit(self, debug=False):
 
                     # Provide an initial guess
                     init_guess_p0 = self.calDAC2Q_m[vfat]*(8+stepN*8)+self.calDAC2Q_b[vfat]
-                    init_guess_p1 = abs(self.calDAC2Q_m[vfat]*rand)  #self.calDAC2Q_m[vfat] might be negative (e.g. VFAT3 case)
+                    if self.isVFAT3:
+                        init_guess_p1 = self.calDAC2Q_m[vfat]*(256)+self.calDAC2Q_b[vfat]+abs(self.calDAC2Q_m[vfat]*rand)  #self.calDAC2Q_m[vfat] might be negative (e.g. VFAT3 case)
```
bdorney: What's the rational here?

You're setting the initial guess for `[1]` which represents the scurve width (should be ENC).  This should be in units of fC.  

```python
(self.calDAC2Q_m[vfat]*(256)+self.calDAC2Q_b[vfat])+abs(self.calDAC2Q_m[vfat]*rand)
```

So the first term in paranthesis (I added just to make it easier to reference) is the minimum (maximum) charge in voltage step (current injection mode), this is not the ENC.  Then you're adding to it a slope term that is in fC which could more aptly represent the ENC.

So why for the initial guess have this first term?  While it might work in the few test cases I wonder if there is not a better approach.  Remind me the reason you identified for this failing?  It was performing an infinite loop but I don't remember the cause.

Maybe there is a better more physical way to solve this than adding what seems like an arbitrary charge amount, which may fail in the case of current injection mode (unsure...)

fitting/fitScanData.py
brentstone: thanks, it certainly is. i'll reformat

fitting/fitScanData.py
brentstone: As for the whole term you have put in parentheses, I took that from a couple lines down where this is the lower bound on the range for the fit: https://github.com/brentstone/gem-plotting-tools/blob/develop/fitting/fitScanData.py#L199.

Since this initial value is always positive, I simply added it to the lower bound to ensure that the first guess is within our set bounds. The reason for this failing is essentially that the lower bound is too high for the guess. The boolean ```fitValid``` is used to determine whether to enter a ```continue``` statement (enter when it is false), and so we enter the infinite loop since our starting point is not within the bounds.

brentstone: Admittedly I didn't second guess the choice of lower bound; I agree that we should maybe think of a better starting point for the fit. However, the converged values typically were closer to the lower bound than the upper, which this method is suitable for.

bdorney: Update the comment on this line to reflect what we have done, e.g. added the guess to the lower bound and then I'll approve this.

fitting/fitScanData.py
```diff
@@ -337,7 +337,7 @@ def fit(self, debug=False):
                                     )
 
                     # Fit
-                    fitResult = self.scanHistos[vfat][ch].Fit('myERF','SQ')
+                    fitResult = self.scanHistos[vfat][ch].Fit('myERF','SQ')
```
bdorney: trailing white space here, this will generate spurious `git diff`'s that will be hard to track down.  

I would suggest configuring your editor to how to strip trailing whitespaces on saving the file [example in vim](http://vim.wikia.com/wiki/Remove_unwanted_spaces).  Similar examples exist for `emacs`.

bdorney: See for example coding style guidelines that we have setup for the repo's: 

https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/.github/CONTRIBUTING.md

Note this is common to all gemdaq repo's.

fitting/fitScanData.py
```diff
@@ -276,11 +276,11 @@ def fit(self, debug=False):
...
 
                     # Provide an initial guess
-                    init_guess_p0 = self.calDAC2Q_m[vfat]*(8+stepN*8)+self.calDAC2Q_b[vfat]
+                    init_guess_p0 = self.calDAC2Q_m[vfat]*(8+stepN*8)+self.calDAC2Q_b[vfat] 
                     if self.isVFAT3:
                         init_guess_p1 = self.calDAC2Q_m[vfat]*(256)+self.calDAC2Q_b[vfat]+abs(self.calDAC2Q_m[vfat]*rand)  #self.calDAC2Q_m[vfat] might be negative (e.g. VFAT3 case)
-		    else:
-			init_guess_p1 = abs(self.calDAC2Q_m[vfat]*rand)
+                    else:
+                        init_guess_p1 = abs(self.calDAC2Q_m[vfat]*rand)
```
bdorney: It looks like you've inserted the `\t` character here?  Maybe git is just rendering it as a tab for some reason.

If it is a tab, it should be replaced by four `\s` (space) characters since different editors may render the `\t` character differently and in a whitespace sensitive language (e.g. python) this could cause conflicts (e.g. case statements not matching up).

To convert tabs to spaces see this [vim example ](http://vim.wikia.com/wiki/Converting_tabs_to_spaces)

bdorney (changes requested):
Subtle formatting changes needed and one comment.  Otherwise fine.

bdorney:
Also for historical reasons please provide the Mattermost links to the plots you uploaded either in your main message above or as a comment here illustrating the results of testing.

fitting/fitScanData.py
brentstone: thx

fitting/fitScanData.py
brentstone: they way it is now is correct, with no tabs and all spaces (had tabs before)

bdorney (approved):

PR #157, opened 2018-10-09
New Feature: adding new directories in DATA_PATH for new vfatqc tools

bdorney:
## Description

Added new subdirectories under `DATA_PATH` to cover automatic lookup of new `vfatqc` tools

### Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Motivation and Context

Adding new tools to `run_scans.py` and a planned (future) update of `ana_scans.py`.  Automatic parsing of subdirectories is helpful.

## How Has This Been Tested?

WIP

## Checklist:

### DIFF
--- a/utils/anautilities.py
+++ b/utils/anautilities.py
@@ -146,7 +146,7 @@ def getDirByAnaType(anaType, cName, ztrim=4):
     elif anaType == "trim":
         dirPath = "%s/%s/%s/z%f/"%(dataPath,cName,anaType,ztrim)
     elif anaType == "trimV3":
-        dirPath = "%s/%s/%s/"%(dataPath,cName,anaType)
+        dirPath = "%s/%s/trim/"%(dataPath,cName)
 
     return dirPath
 
--- a/utils/anautilities.py
+++ b/utils/anautilities.py
@@ -121,18 +121,32 @@ def getDirByAnaType(anaType, cName, ztrim=4):
     dataPath  = os.getenv('DATA_PATH')
 
     dirPath = ""
-    if anaType == "latency":
+    if anaType == "dacScanV3":
+        dirPath = "%s/%s"%(dataPath,anaType)
+    elif anaType == "latency":
         dirPath = "%s/%s/%s/trk/"%(dataPath,cName,anaType)
+    elif anaType == "sbitMonInt":
+        dirPath = "%s/%s/sbitMonitor/intTrig/"%(dataPath,cName)
+    elif anaType == "sbitMonRO":
+        dirPath = "%s/%s/sbitMonitor/readout/"%(dataPath,cName)
+    elif anaType == "sbitRatech":
+        dirPath = "%s/%s/sbitRate/perchannel/"%(dataPath,cName)
+    elif anaType == "sbitRateor":
+        dirPath = "%s/%s/sbitRate/channelOR/"%(dataPath,cName)
     elif anaType == "scurve":
         dirPath = "%s/%s/%s/"%(dataPath,cName,anaType)
+    elif anaType == "temperature":
+        dirPath = "%s/%s"%(dataPath,anaType)
     elif anaType == "thresholdch":
-        dirPath = "%s/%s/%s/channel/"%(dataPath,cName,"threshold")
+        dirPath = "%s/%s/threshold/channel/"%(dataPath,cName)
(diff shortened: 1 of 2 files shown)


utils/anautilities.py
```diff
@@ -121,18 +121,32 @@ def getDirByAnaType(anaType, cName, ztrim=4):
...
+    elif anaType == "sbitMonRO":
+        dirPath = "%s/%s/sbitMonitor/readout/"%(dataPath,cName)
+    elif anaType == "sbitRatech":
+        dirPath = "%s/%s/sbitRate/perchannel/"%(dataPath,cName)
+    elif anaType == "sbitRateor":
+        dirPath = "%s/%s/sbitRate/channelOR/"%(dataPath,cName)
     elif anaType == "scurve":
         dirPath = "%s/%s/%s/"%(dataPath,cName,anaType)
+    elif anaType == "temperature":
+        dirPath = "%s/%s"%(dataPath,anaType)
```
jsturdy, line 139: this monitoring isn't done per-chamber?

utils/anautilities.py
```diff
@@ -121,18 +121,32 @@ def getDirByAnaType(anaType, cName, ztrim=4):
...
     elif anaType == "thresholdvftrig":
-        dirPath = "%s/%s/%s/vfat/trig/"%(dataPath,cName,"threshold")
+        dirPath = "%s/%s/threshold/vfat/trig/"%(dataPath,cName)
     elif anaType == "thresholdvftrk":
-        dirPath = "%s/%s/%s/vfat/trk/"%(dataPath,cName,"threshold")
+        dirPath = "%s/%s/threshold/vfat/trk/"%(dataPath,cName)
     elif anaType == "trim":
         dirPath = "%s/%s/%s/z%f/"%(dataPath,cName,anaType,ztrim)
+    elif anaType == "trimV3":
+        dirPath = "%s/%s/trim/"%(dataPath,cName)
```
jsturdy, line 149: is there a possibility that trimming *could* be done different ways/to different points with the V3 tools?

utils/anautilities.py
bdorney: The output TFile produced by dacScanV3.py may contain multiple chambers worth of information depending on the ohMask used at runtime.  If this design choice is undesirable than I will try to come up with a way to get one-to-one correspondence.  But it may not be possible.  The analysis routine of course could break the results into one-to-one correspondence.

bdorney: The same situation exists for the temperature data.

utils/anautilities.py
bdorney: Can you describe what different ways/points you're thinking about?

utils/anautilities.py
jsturdy: OK

utils/anautilities.py
jsturdy: trying to understand if it would be at some point of interest to have "tagged" trim results, that could be toggled to when setting a configuration; in this case, having all results(dates) fall under the `trim/` subdirectory would be less than ideal.
in the vein of the V2, something a la `ztrim` or `vt1value`. maybe in the case of V3, the working point trimmed at, since it matters.
though if no such mechanism exists (or has reasonable expectation to exist, i.e., there will **only** ever be one way the trimming is done) the current structure would suffice

bdorney: I’ve been thinking about this for sometime. I think what might be interesting is from the CFG_THR_ARM_DAC calibration w can develop a set of trim configurations such that the threshold is set to 0.5, 1.0, 1.5, ... fC.  However with the current toolkit and reliance on the VFAT DAQ MONITOR this would be prohibitively long to perform.

However in this line of thinking the trim folder could be divided into sub folders which specify the trimmed threshold value. This would require a small change to the trim routine for the user to specify the trimmed threshold value, e.g. —thresh=0.5 for 0.5 fC.

It’s doable, but I would suggest a dedicated set of PR’s.  And for a multilink case would need to be done using the full data path for speed purposes.

jsturdy: OK, so for the current implementation, there's no need.  And for the future, when the scans are using tracking data, this structure may likely entirely disappear (run-based, rather than date+tag-based, database storage, etc.,) so it won't be worth implementing here

jsturdy (approved):

Issue #155, opened 2018-10-08
Feature Request: DAC Scan Analysis Routine (Good for newcomer)


bdorney:
## Brief summary of issue

We have a new DAC Scan routine for VFAT3 introduced in https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/203.

Now we need a tool to analyze it.  

One critical thing to keep in mind is that `dacScanV3.py` can scan multiple optohybrids per call, so the output `TFile` may contain info for multiple detectors.

## Timescale
Would need this by Monday October 15th for the 904 Camp.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [X] Feature request (request for change which adds functionality)

## Expected Behavior

Relevant pseudocode below:

```python
if __name__ == '__main__':
   setup an argparse.ArgumentParser instance
   get arguments from a user
   
   # Determine which links present suggest using root_numpy, something like:
   import root_numpy as rp
   import numpy as np
   bNames = ['link']
   ohArray = rp.tree2array(tree=calTree, branches=bNames)
   ohArray = np.unique(ohArray) # now ohArray will have one entry for each unmasked optohybrid

   # Determine which DAC was scanned and against which ADC
   nameX = ""
   nameY = ""
   for event in calTree:
      nameX = event.nameX
      nameY = event.nameY
      break # all entries will be the same

   # Get ADC0 or ADC1 calibration
   # nameY will either be ADC0 or ADC1, you'll need to load calibration for each link
   # use parseCalFile(...) from gem-plotting-tools/utils/anautilities.py
   # Note this calibration is unique to each (vfat,ohN) pairing
   # In future this is where a DB query based on vfatID would take place 

   # Make output plots, suggest TGraphErrors objects
   # I would suggest you use a nesteddict object (gempython/utils/nesteddict.py) to index these by [oh][vfat]
   for link in ohArray:
      for vfat in range(0,24):
           dict_dacScanResults[link][vfat] = r.TGraphErrors(...) # declare TGraphErrors
           You can get number of points in TGraphErrors by looking at maxVfat3DACSize of amc_user_functions_xhal.py (from cmsgemos PR#206)

   # Loop over events in the tree and fill plots
   for event in calTree:
       assign (dacValX, dacValY, dacValY_Err) to the tgraph for a given (vfat,oh) pair
       Here you need to use the ADCX (for X = 0,1) calibration to convert dacValY & dacValY_Err to physical units
[...]

PR #158, opened 2018-10-12
add DAC scan analysis

AndrewLevin:
This script analyzes DAC scans. It reads an input file containing DAC and ADC values, performs calibration, performs a fit, and reports the DAC values which should produce the nominal voltage or current as specified in the VFAT3 manual. 3 by 8 plots of the calibrated data and the fit are also produced.

## Description
Pseudo-code, from @bdorney, copied here for convenience from  https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/155:

```
if __name__ == '__main__':
   setup an argparse.ArgumentParser instance
   get arguments from a user
   
   # Determine which links present suggest using root_numpy, something like:
   import root_numpy as rp
   import numpy as np
   bNames = ['link']
   ohArray = rp.tree2array(tree=calTree, branches=bNames)
   ohArray = np.unique(ohArray) # now ohArray will have one entry for each unmasked optohybrid

   # Determine which DAC was scanned and against which ADC
   nameX = ""
   nameY = ""
   for event in calTree:
      nameX = event.nameX
      nameY = event.nameY
      break # all entries will be the same

   # Get ADC0 or ADC1 calibration
   # nameY will either be ADC0 or ADC1, you'll need to load calibration for each link
   # use parseCalFile(...) from gem-plotting-tools/utils/anautilities.py
   # Note this calibration is unique to each (vfat,ohN) pairing
   # In future this is where a DB query based on vfatID would take place 

   # Make output plots, suggest TGraphErrors objects
   # I would suggest you use a nesteddict object (gempython/utils/nesteddict.py) to index these by [oh][vfat]
   for link in ohArray:
      for vfat in range(0,24):
           dict_dacScanResults[link][vfat] = r.TGraphErrors(...) # declare TGraphErrors
           You can get number of points in TGraphErrors by looking at maxVfat3DACSize of amc_user_functions_xhal.py (from cmsgemos PR#206)

   # Loop over events in the tree and fill plots
   for event in calTree:
       assign (dacValX, dacValY, dacValY_Err) to the tgraph for a given (vfat,oh) pair
       Here you need to use the ADCX (for X = 0,1) calibration to convert dacValY & dacValY_Err to physical units
       You also need to identify if nameX corresponds to a bias voltage DAC or a bias current DAC, see VFAT3 manual Table 25: GBL CFG CTR 4 : Monitoring settings
       Note nameX is the register that was scanned during the DAC scan, but we want plots that are Y=nameX (reg scanned) and X=nameY (ADC value), see below 
 
   # Fit the TGraphErrors Objects
   for oh in ohArray
      for vfat in range(0,24):
          dict_dacScanResults[oh][vfat].Fit(...) # linear fitting may not be suitable, polynomial may be required, investigation needed.  Possible try several fits and choose one with best norm chi2
   
    # Determine DAC values to achieve recommended bias voltage and current settings
    See Table 9, Nominal Current column, and the VFAT3a specific note under table 9
    See Table 10, Nominal Voltage column
    Here for each (vfat,oh) you should use the fit function to do a look-up on the nominal value to get the DAC value that will achieve this, e.g.
    dacVal = myTF1.Eval(nominal voltage)
    This then should be printed to terminal as a table, an output file (in either tab-delimited or TTree format) and then as a TGraph object with DAC value vs. VFAT position

    # Make plots
    Make our standard 3x8 grid plot of the DAC scan results, see make3x8Canvas(...), one such grid plot should be made for each entry in ohArray
    Store plots in an output TFile (use TDirectories to organize logically)
    
    Exit
```

Help output:
```
[amlevin@gem904qc8daq gem-plotting-tools]$ ./anaDACScan.py -h
usage: anaDACScan.py [-h] [--calFileList calFileList] [-o outfilename]
                     [--assignXErrors]
                     infilename

Arguments to supply to anaDACScan.py

positional arguments:
  infilename            Filename from which input information is read

optional arguments:
  -h, --help            show this help message and exit
  --calFileList calFileList
                        File specifying which calFile to use for each OH.
                        Format: 0 /path/to/my/cal/file.txt<newline> 1
                        /path/to/my/cal/file.txt<newline>...
  -o outfilename, --outfilename outfilename
                        Filename to which output information is written
  --assignXErrors       dacValY_Err values are always used, but dacValX_Err
                        values are used if and only if this flag is set
```

calFileList example:
```
1 /home/amlevin/gem-plotting-tools/calibration_OH1.txt
```

Example execution:
```
[amlevin@gem904qc8daq gem-plotting-tools]$ ./anaDACScan.py dacScanV3.root  --calFileList calibration.txt 
Analyzing: 'dacScanV3.root'
Warning: the nominal DAC value is that was found from the fit is less than 0
Warning: the nominal DAC value is that was found from the fit is less than 0
Warning: the nominal DAC value is that was found from the fit is less than 0
Warning: the nominal DAC value is that was found from the fit is less than 0
Warning: the nominal DAC value is that was found from the fit is less than 0
Warning: the nominal DAC value is that was found from the fit is less than 0
Warning: the nominal DAC value is that was found from the fit is less than 0
Warning: the nominal DAC value is that was found from the fit is less than 0
Warning: the nominal DAC value is that was found from the fit is less than 0
Warning: the nominal DAC value is that was found from the fit is less than 0
Warning: the nominal DAC value is that was found from the fit is less than 0
| OH | vfatN | dacVal |
| :-: | :---: | :----: |
| 1 | 0  | 120 | 
| 1 | 1  | 120 | 
| 1 | 2  | 120 | 
| 1 | 3  | 120 | 
| 1 | 4  | -87 | 
| 1 | 5  | -82 | 
| 1 | 6  | -98 | 
| 1 | 7  | -75 | 
| 1 | 8  | 120 | 
| 1 | 9  | 120 | 
| 1 | 10  | 120 | 
| 1 | 11  | 120 | 
| 1 | 12  | -72 | 
| 1 | 13  | -100 | 
| 1 | 14  | -76 | 
| 1 | 15  | 120 | 
| 1 | 16  | 120 | 
| 1 | 17  | 120 | 
| 1 | 18  | 120 | 
| 1 | 19  | 120 | 
| 1 | 20  | -85 | 
| 1 | 21  | -97 | 
| 1 | 22  | -111 | 
| 1 | 23  | -104 | 
Info in <TCanvas::Print>: png file Summary.png has been created
```

### Types of changes

## Motivation and Context
This script was requested in https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/155

## How Has This Been Tested?
I have tested that the script runs on an input file provided by @bdorney: `dacScanV3.root`

Output root file structure:
connection to localhost 10152018-08 16 10

DACvsADC:
dacvsadc

RawADCvsDAC:
rawadcvsdac

NominalDacValVsVFATX:
nominaldacvalvsvfatx

Summary plot:
summary

## Checklist:

### DIFF
--- a/anaDACScan.py
+++ b/anaDACScan.py
@@ -40,6 +40,10 @@ Arguments
 
     Name of the output root file. Default is DACFitData.root.
 
+.. option:: --output_txtfile_filename
+
+    Name of output text file that will contain the nominal DAC values in tab-delimited format. Default is NominalDACValues.txt.
+
 Example
 -------
 
@@ -185,12 +189,12 @@ if __name__ == '__main__':
         #Use Ohm's law to convert the currents to voltages. The VFAT3 team told us that a 20k ohm resistor was used.
         if nominalDacValues[nameX][1][1] == "A":
 
+            calibrated_ADC_value = calibrated_ADC_value/20000.0
+            calibrated_ADC_error = calibrated_ADC_value/20000.0
+
             if nameX != 'CFG_IREF':
                 calibrated_ADC_value -= nominalDacValues['CFG_IREF'][0] 
             
-            calibrated_ADC_value = calibrated_ADC_value*20000.0
-            calibrated_ADC_error = calibrated_ADC_value*20000.0
-
         #the reversal of x and y is intended - we want to plot the nameX variable on the y-axis and the nameY variable on the x-axis
         #the nameX variable is the DAC register that is scanned, and we want to determine its nominal value
         if args.assignXErrors:
--- a/anaDACScan.py
+++ b/anaDACScan.py
@@ -79,7 +79,6 @@ if __name__ == '__main__':
     r.TH1.SetDefaultSumw2(False)
     r.gROOT.SetBatch(True)
     r.gStyle.SetOptStat(1111111)
-    r.gStyle.SetDrawOption
     
     dacScanFile = r.TFile(args.infilename)
 
--- a/anaDACScan.py
+++ b/anaDACScan.py
@@ -122,7 +122,7 @@ if __name__ == '__main__':
         print("Error: unexpected value of nameX: '%s'"%nameX)
         exit(1)
 
-    #the nominal reference current is 10 uA and it has a scaling    
+    #the nominal reference current is 10 uA and it has a scaling factor of 0.5   
     nominal_iref = 10/0.5
 
     nominal = nominalDacValues[nameX][0]/nominalDacScalingFactors[nameX]
--- a/anaDACScan.py
+++ b/anaDACScan.py
@@ -188,15 +188,15 @@ if __name__ == '__main__':
     for oh in ohArray:
         for vfat in range(0,24):
             dict_RawADCvsDAC_Graphs[oh][vfat] = r.TGraphErrors()
-            dict_RawADCvsDAC_Graphs[oh][vfat].GetXaxis().SetTitle(nameX)
-            dict_RawADCvsDAC_Graphs[oh][vfat].GetYaxis().SetTitle(nameY)
+            dict_RawADCvsDAC_Graphs[oh][vfat].GetXaxis().SetTitle(nameY)
+            dict_RawADCvsDAC_Graphs[oh][vfat].GetYaxis().SetTitle(nameX)
             dict_DACvsADC_Graphs[oh][vfat] = r.TGraphErrors()
             #the reversal of x and y is intended - we want to plot the nameX variable on the y-axis and the nameY variable on the x-axis
             if nominalDacValues[nameX][1][len(nominalDacValues[nameX][1])-1] == 'A':
-                dict_DACvsADC_Graphs[oh][vfat].GetXaxis().SetTitle(nameY + " (uA)")
+                dict_DACvsADC_Graphs[oh][vfat].GetXaxis().SetTitle(nameX + " (uA)")
             else:
-                dict_DACvsADC_Graphs[oh][vfat].GetXaxis().SetTitle(nameY + " (mV)")
-            dict_DACvsADC_Graphs[oh][vfat].GetYaxis().SetTitle(nameX)
+                dict_DACvsADC_Graphs[oh][vfat].GetXaxis().SetTitle(nameX + " (mV)")
+            dict_DACvsADC_Graphs[oh][vfat].GetYaxis().SetTitle(nameY)
 
     outputFiles = {}         
              
--- a/anaDACScan.py
+++ b/anaDACScan.py
@@ -1,8 +1,60 @@
 #!/usr/bin/env python
 
-"""
-anaDACScan
-==============
+r"""
+``anaDACScan`` -- analyzes DAC scan data
+========================================
+
+Synopsis
+--------
+
+**anaDACScan.py** [*OPTIONS*]
+
+Description
+-----------
+
+This script reads in calibration information and DAC scan data, performs a fit for each VFAT, computes the DAC value corresponding to the nominal current or voltage for each VFAT, and reports the results.
+
+
+Arguments
+---------
+
+.. program:: anaDACScan.py
+
+.. option:: infilename
+
+    Name of the input file root.
+
+.. option:: --assignXErrors  
+
+    Use the dacValX_Err values stored in the input file
+
+.. option:: --calFileList
+  
+    Provide a file containing a list of calFiles in the format: 
+    OH1 calFileOH1
+    OH2 calFileOH2
+    OH3 calFileOH3
+
+.. option:: --outfilename
+
+    Name of the output root file. Default is DACFitData.root.
+
+Example
+-------
+
+.. code-block:: bash
+
+    anaDACScan.py /afs/cern.ch/user/d/dorney/public/v3Hack/dacScanV3.root  --calFileList calibration.txt --assignXErrors
+
+
+Environment
+-----------
+
+
+Internals
+---------
+
+
 """
 
 if __name__ == '__main__':
--- a/anaDACScan.py
+++ b/anaDACScan.py
@@ -67,7 +67,7 @@ if __name__ == '__main__':
     parser.add_argument('infilename', type=str, help="Filename from which input information is read", metavar='infilename')
     parser.add_argument("--calFileList", type=str, help="File specifying which calFile to use for each OH. Format: 0 /path/to/my/cal/file.txt<newline> 1 /path/to/my/cal/file.txt<newline>...", metavar="calFileList")
     parser.add_argument('-o','--outfilename', dest='outfilename', type=str, default="DACFitData.root", help="Filename to which output information is written", metavar='outfilename')
-    parser.add_argument('-t','--output_txtfile_filename_base', dest='output_txtfile_filename', type=str, default="NominalDACValues.txt", help="Name of output text file that will contain the nominal DAC values in tab-delimited format", metavar='output_txtfile_filename')
+    parser.add_argument('-t','--output_txtfile_filename', dest='output_txtfile_filename', type=str, default="NominalDACValues.txt", help="Name of output text file that will contain the nominal DAC values in tab-delimited format", metavar='output_txtfile_filename')
     parser.add_argument('--assignXErrors', dest='assignXErrors', action='store_true', help="Whether to assign errors for the X variable (which is actually plotted on the y-axis)")
 
     args = parser.parse_args()
@@ -188,8 +188,8 @@ if __name__ == '__main__':
             if nameX != 'CFG_IREF':
                 calibrated_ADC_value -= nominalDacValues['CFG_IREF'][0] 
             
-            calibrated_ADC_value = calibrated_ADC_value/20000.0
-            calibrated_ADC_error = calibrated_ADC_value/20000.0
+            calibrated_ADC_value = calibrated_ADC_value*20000.0
+            calibrated_ADC_error = calibrated_ADC_value*20000.0
 
         #the reversal of x and y is intended - we want to plot the nameX variable on the y-axis and the nameY variable on the x-axis
         #the nameX variable is the DAC register that is scanned, and we want to determine its nominal value
--- a/anaDACScan.py
+++ b/anaDACScan.py
@@ -25,7 +25,7 @@ Arguments
 
 .. option:: --assignXErrors  
 
-    If this flag is set, the dacValX_Err values stored in the input file when constructing the plots. If this flag is not set, the error associated with the dacValX values is always taken to be 0. The dacValY_err values are always used.
+    If this flag is set, the dacValX_Err values stored in the input file are used when constructing the plots and performing the fits. If this flag is not set, the error associated with the dacValX values is always taken to be 0. Note that these errors are actually errors on the variable plotted on the y-axis in the DAC vs ADC plots. The dacValY_Err values, which would correspond to the errors on the x-axis in the DAC vs ADC plots, cannot be used because fits do not converge when they are used. 
 
 .. option:: --calFileList
   
@@ -231,13 +231,13 @@ if __name__ == '__main__':
         if args.assignXErrors:
             dict_DACvsADC_Graphs[oh][vfat].SetPoint(dict_DACvsADC_Graphs[oh][vfat].GetN(),calibrated_ADC_value,event.dacValX)
             dict_RawADCvsDAC_Graphs[oh][vfat].SetPoint(dict_RawADCvsDAC_Graphs[oh][vfat].GetN(),event.dacValY,event.dacValX)
-            dict_DACvsADC_Graphs[oh][vfat].SetPointError(dict_DACvsADC_Graphs[oh][vfat].GetN()-1,calibrated_ADC_error,event.dacValX_Err)
-            dict_RawADCvsDAC_Graphs[oh][vfat].SetPointError(dict_RawADCvsDAC_Graphs[oh][vfat].GetN()-1,event.dacValY_Err,event.dacValX_Err)
+            dict_DACvsADC_Graphs[oh][vfat].SetPointError(dict_DACvsADC_Graphs[oh][vfat].GetN()-1,0,event.dacValX_Err)
+            dict_RawADCvsDAC_Graphs[oh][vfat].SetPointError(dict_RawADCvsDAC_Graphs[oh][vfat].GetN()-1,0,event.dacValX_Err)
         else:
             dict_DACvsADC_Graphs[oh][vfat].SetPoint(dict_DACvsADC_Graphs[oh][vfat].GetN(),calibrated_ADC_value,event.dacValX)
             dict_RawADCvsDAC_Graphs[oh][vfat].SetPoint(dict_RawADCvsDAC_Graphs[oh][vfat].GetN(),event.dacValY,event.dacValX)
-            dict_DACvsADC_Graphs[oh][vfat].SetPointError(dict_DACvsADC_Graphs[oh][vfat].GetN()-1,calibrated_ADC_error,0)
-            dict_RawADCvsDAC_Graphs[oh][vfat].SetPointError(dict_RawADCvsDAC_Graphs[oh][vfat].GetN()-1,event.dacValY_Err,0)
+            dict_DACvsADC_Graphs[oh][vfat].SetPointError(dict_DACvsADC_Graphs[oh][vfat].GetN()-1,0,0)
+            dict_RawADCvsDAC_Graphs[oh][vfat].SetPointError(dict_RawADCvsDAC_Graphs[oh][vfat].GetN()-1,0,0)
 
     # Fit the TGraphErrors objects    
     for oh in ohArray:
--- a/anaDACScan.py
+++ b/anaDACScan.py
@@ -66,13 +66,15 @@ if __name__ == '__main__':
     from gempython.gemplotting.utils.anautilities import make3x8Canvas
     
     from gempython.gemplotting.mapping.chamberInfo import chamber_config
+
+    import os
     
     import argparse
     
     parser = argparse.ArgumentParser(description='Arguments to supply to anaDACScan.py')
 
     parser.add_argument('infilename', type=str, help="Filename from which input information is read", metavar='infilename')
-    parser.add_argument("calFileFile", type=str, help="File specifying which calFile to use for each OH. Format: OH1 filenameOH1 <newline> OH2 filenameOH2 <newline> etc", metavar="calFileFile")
+    parser.add_argument("--calFileList", type=str, help="File specifying which calFile to use for each OH. Format: OH1 filenameOH1 <newline> OH2 filenameOH2 <newline> etc", metavar="calFileList")
     parser.add_argument('-o','--outfilename', dest='outfilename', type=str, default="DACFitData.root", help="Filename to which output information is written", metavar='outfilename')
     parser.add_argument('--assignXErrors', dest='assignXErrors', action='store_true', help="Whether to assign errors for the X variable (which is actually plotted on the y-axis)")
 
@@ -109,45 +111,61 @@ if __name__ == '__main__':
 
     calInfo = {}
 
-    for line in open(args.calFileFile):
-        line = line.strip(' ').strip('\n')
-        first = line.split(' ')[0].strip(' ')
-        second = line.split(' ')[1].strip(' ')
-        tuple_calInfo = parseCalFile(second)
-        calInfo[int(first)] = {'slope' : tuple_calInfo[0], 'intercept' : tuple_calInfo[1]}
-        
+    if args.calFileList != None:
+        for line in open(args.calFileList):
+            line = line.strip(' ').strip('\n')
+            first = line.split(' ')[0].strip(' ')
+            second = line.split(' ')[1].strip(' ')
+            tuple_calInfo = parseCalFile(second)
+            calInfo[int(first)] = {'slope' : tuple_calInfo[0], 'intercept' : tuple_calInfo[1]}
+
+    dataPath = os.getenv("DATA_PATH") 
+
+    #for each OH, check if calibration files were provided, if not search for the calFile in the $DATA_PATH, if it is not there, then skip that OH for the rest of the script
+    for oh in ohArray:
+        if oh not in calInfo.keys():
+            calDacCalFile = "{0}/{1}/calFile_calDac_{1}.txt".format(dataPath,chamber_config[oh])
+            calDacCalFileExists = os.path.isfile(calDacCalFile)
+            if not calDacCalFileExists:
+                print("Skipping OH{0}, detector {1}, missing CFG_CAL_DAC Calibration file:\n\t{2}".format(
+                    oh,
+                    chamber_config[oh],
+                    calDacCalFile))
+                ohArray = np.delete(ohArray,(oh))
+
+
+    if len(ohArray.keys()) == 0:
+        print('No OHs with a calFile, exiting.')
+        exit(1)
+           
     # Initialize nested dictionaries
     dict_dacScanResults = nesteddict()
     dict_dacScanFuncs = nesteddict()
 
     # Initialize a TGraphErrors for each vfat
-    for link in ohArray:
+    for oh in ohArray:
         for vfat in range(0,24):
-             dict_dacScanResults[link][vfat] = r.TGraphErrors()
+             dict_dacScanResults[oh][vfat] = r.TGraphErrors()
 
     outputFiles = {}         
              
-    for link in ohArray:         
-        outputFiles[link] = r.TFile("$DATA_PATH/"+chamber_config[link]+"/dacScans/scandate/"+args.outfilename,'recreate')
+    for oh in ohArray:         
+        outputFiles[oh] = r.TFile("$DATA_PATH/"+chamber_config[oh]+"/dacScans/scandate/"+args.outfilename,'recreate')
              
     # Loop over events in the tree and fill plots
     for event in dacScanFile.dacScanTree:
-        link = event.link
+        oh = event.link
         vfat = event.vfatN
 
-        if link not in calInfo.keys():
-            print('Error: calibration file for link %i was not provided'%link)
-            exit(1)
-        
-        calibrated_DAC_value=calInfo[link]['slope'][vfat]*event.dacValX+calInfo[link]['intercept'][vfat]
-        calibrated_DAC_error=calInfo[link]['slope'][vfat]*event.dacValX_Err+calInfo[link]['intercept'][vfat]
+        calibrated_DAC_value=calInfo[oh]['slope'][vfat]*event.dacValX+calInfo[oh]['intercept'][vfat]
+        calibrated_DAC_error=calInfo[oh]['slope'][vfat]*event.dacValX_Err+calInfo[oh]['intercept'][vfat]
         
         if args.assignXErrors:
-            dict_dacScanResults[link][vfat].SetPoint(dict_dacScanResults[link][vfat].GetN(),event.dacValY,calibrated_DAC_value)
-            dict_dacScanResults[link][vfat].SetPointError(dict_dacScanResults[link][vfat].GetN()-1,event.dacValY_Err,calibrated_DAC_error)
+            dict_dacScanResults[oh][vfat].SetPoint(dict_dacScanResults[oh][vfat].GetN(),event.dacValY,calibrated_DAC_value)
+            dict_dacScanResults[oh][vfat].SetPointError(dict_dacScanResults[oh][vfat].GetN()-1,event.dacValY_Err,calibrated_DAC_error)
         else:    
-            dict_dacScanResults[link][vfat].SetPoint(dict_dacScanResults[link][vfat].GetN(),event.dacValY,calibrated_DAC_value)
-            dict_dacScanResults[link][vfat].SetPointError(dict_dacScanResults[link][vfat].GetN()-1,event.dacValY_Err,0)
+            dict_dacScanResults[oh][vfat].SetPoint(dict_dacScanResults[oh][vfat].GetN(),event.dacValY,calibrated_DAC_value)
+            dict_dacScanResults[oh][vfat].SetPointError(dict_dacScanResults[oh][vfat].GetN()-1,event.dacValY_Err,0)
 
     # Fit the TGraphErrors objects    
     for oh in ohArray:
--- a/anaDACScan.py
+++ b/anaDACScan.py
@@ -237,33 +237,36 @@ if __name__ == '__main__':
              dacVal_branch[0] = dict_dacVals[oh][vfat]
              tree_dacVals[oh].Fill() 
              
-    # Write out the dacVal results to both a file and the terminal         
+    # Write out the dacVal results to a root file, a text file, and the terminal
+    outputTxtFiles_dacVals = {}    
+            
+    if scandate == 'noscandate':
+        outputTxtFiles_dacVals[oh] = open(elogPath+"/"+chamber_config[oh]+"/"+args.output_txtfile_filename,'w')
+    else:    
+        outputTxtFiles_dacVals[oh] = open(dataPath+"/"+chamber_config[oh]+"/dacScans/"+scandate+"/"+args.output_txtfile_filename,'w')
+
     print( "| OH | vfatN | dacVal |")
     print( "| :-: | :---: | :----: |")        
     
     for oh in ohArray:
         outputFiles[oh].cd()
+        outputFiles[oh].mkdir("Summary")
+        outputFiles[oh].cd("Summary")
         graph_dacVals[oh].Write("dacVals")
         tree_dacVals[oh].Write("dacVals")
         for vfat in range(0,24):
+            outputFiles[oh].cd()
+            outputFiles[oh].mkdir("VFAT"+str(vfat))
+            outputFiles[oh].cd("VFAT"+str(vfat))
+            dict_DACvsADC_Graphs[oh][vfat].Write()
+            outputFiles[oh].cd("../")
+            outputTxtFiles_dacVals[oh].write(str(vfat)+"\t"+str(dict_dacVals[oh][vfat])+"\n")
             print("| {0} | {1}  | {2} | ".format(
                 oh,
                 vfat,
                 dict_dacVals[oh][vfat]
             ))
 
-    outputTxtFiles_dacVals = {}    
-            
-    # Write out the results to a txt file in tab-delimited format
-    if scandate == 'noscandate':
-        outputTxtFiles_dacVals[oh] = open(elogPath+"/"+chamber_config[oh]+"/"+args.output_txtfile_filename,'w')
-    else:    
-        outputTxtFiles_dacVals[oh] = open(dataPath+"/"+chamber_config[oh]+"/dacScans/"+scandate+"/"+args.output_txtfile_filename,'w')
-
-    for oh in ohArray:
-        for vfat in range(0,24):
-            outputTxtFiles_dacVals[oh].write(str(vfat)+"\t"+str(dict_dacVals[oh][vfat])+"\n")
-            
     # Make plots        
     for oh in ohArray:
         canv_Summary = make3x8Canvas('canv_Summary',dict_DACvsADC_Graphs[oh],'AP',dict_DACvsADC_Funcs[oh],'')
@@ -272,4 +275,5 @@ if __name__ == '__main__':
         else:
             canv_Summary.SaveAs(dataPath+"/"+chamber_config[oh]+"/dacScans/"+scandate+"/SummaryOH.png")
         outputFiles[oh].cd()
+        outputFiles[oh].cd("Summary")
         canv_Summary.Write()
--- a/anaDACScan.py
+++ b/anaDACScan.py
@@ -25,7 +25,7 @@ Arguments
 
 .. option:: --assignXErrors  
 
-    Use the dacValX_Err values stored in the input file
+    If this flag is set, the dacValX_Err values stored in the input file when constructing the plots. If this flag is not set, the error associated with the dacValX values is always taken to be 0. The dacValY_err values are always used.
 
 .. option:: --calFileList
   
@@ -69,7 +69,7 @@ if __name__ == '__main__':
     parser.add_argument('infilename', type=str, help="Filename from which input information is read", metavar='infilename')
     parser.add_argument("--calFileList", type=str, help="File specifying which calFile to use for each OH. Format: 0 /path/to/my/cal/file.txt<newline> 1 /path/to/my/cal/file.txt<newline>...", metavar="calFileList")
     parser.add_argument('-o','--outfilename', dest='outfilename', type=str, default="DACFitData.root", help="Filename to which output information is written", metavar='outfilename')
-    parser.add_argument('--assignXErrors', dest='assignXErrors', action='store_true', help="Whether to assign errors for the X variable (which is actually plotted on the y-axis)")
+    parser.add_argument('--assignXErrors', dest='assignXErrors', action='store_true', help="dacValY_Err values are always used, but dacValX_Err values are used if and only if this flag is set")
 
     args = parser.parse_args()
 
--- a/anaDACScan.py
+++ b/anaDACScan.py
@@ -192,7 +192,7 @@ if __name__ == '__main__':
             dict_RawADCvsDAC_Graphs[oh][vfat].GetYaxis().SetTitle(nameY)
             dict_DACvsADC_Graphs[oh][vfat] = r.TGraphErrors()
             #the reversal of x and y is intended - we want to plot the nameX variable on the y-axis and the nameY variable on the x-axis
-            if nominalDacValues['CFG_IREF'][1][1] == 'A':
+            if nominalDacValues[nameX][1][1] == 'A':
                 dict_DACvsADC_Graphs[oh][vfat].GetXaxis().SetTitle(nameY + " (uA)")
             else:
                 dict_DACvsADC_Graphs[oh][vfat].GetXaxis().SetTitle(nameY + " (mV)")
--- a/anaDACScan.py
+++ b/anaDACScan.py
@@ -46,7 +46,7 @@ Example
 
 .. code-block:: bash
 
-    anaDACScan.py /afs/cern.ch/user/d/dorney/public/v3Hack/dacScanV3.root  --calFileList calibration.txt --assignXErrors
+    anaDACScan.py /path/to/input.root  --calFileList calibration.txt --assignXErrors
 
 
 Environment
--- a/anaDACScan.py
+++ b/anaDACScan.py
@@ -91,25 +91,29 @@ if __name__ == '__main__':
 
     dacScanFile = r.TFile(args.infilename)
 
+    import numpy as np
+    import root_numpy as rp
+    
+    list_bNames = ['link']
+    ohArray = rp.tree2array(tree=dacScanFile.dacScanTree, branches=list_bNames)
+    ohArray = np.unique(ohArray['link'])
+    
+    dataPath = os.getenv("DATA_PATH") 
+    
     if len(args.infilename.split('/')) > 1 and len(args.infilename.split('/')[len(args.infilename.split('/')) - 2].split('.')) == 5:
         scandate = args.infilename.split('/')[len(args.infilename.split('/')) - 2]
     else:    
         scandate = 'noscandate'
-    
-    import numpy as np
-    import root_numpy as rp
-    
-    # Determine which links present
-    list_bNames = ['link']
-    ohArray = rp.tree2array(tree=dacScanFile.dacScanTree, branches=list_bNames)
-    ohArray = np.unique(ohArray['link'])
 
+    for oh in ohArray:
+        os.system("mkdir -p "+ dataPath+"/"+chamber_config[oh]+"/dacScans/"+scandate)
+            
     # Determine which DAC was scanned and against which ADC
     nameX = ""
     nameY = ""
     for event in dacScanFile.dacScanTree:
-        nameX = event.nameX.data()
-        nameY = event.nameY.data()
+        nameX = str(event.nameX.data())
+        nameY = str(event.nameY.data())
         break # all entries will be the same
 
     if nameY not in ['ADC0', 'ADC1']:
@@ -125,8 +129,6 @@ if __name__ == '__main__':
             second = line.split(' ')[1].strip(' ')
             tuple_calInfo = parseCalFile(second)
             calInfo[int(first)] = {'slope' : tuple_calInfo[0], 'intercept' : tuple_calInfo[1]}
-
-    dataPath = os.getenv("DATA_PATH") 
 
     #for each OH, check if calibration files were provided, if not search for the calFile in the $DATA_PATH, if it is not there, then skip that OH for the rest of the script
     for oh in ohArray:
@@ -153,6 +155,9 @@ if __name__ == '__main__':
     for oh in ohArray:
         for vfat in range(0,24):
              dict_dacScanResults[oh][vfat] = r.TGraphErrors()
+             #the reversal of x and y is intended - we want to plot the nameX variable on the y-axis and the nameY variable on the x-axis
+             dict_dacScanResults[oh][vfat].GetXaxis().SetTitle(nameY)
+             dict_dacScanResults[oh][vfat].GetYaxis().SetTitle(nameX)
 
     outputFiles = {}         
              
@@ -164,9 +169,10 @@ if __name__ == '__main__':
         oh = event.link
         vfat = event.vfatN
 
+        #the reversal of x and y is intended - we want to plot the nameX variable on the y-axis and the nameY variable on the x-axis
         calibrated_DAC_value=calInfo[oh]['slope'][vfat]*event.dacValX+calInfo[oh]['intercept'][vfat]
         calibrated_DAC_error=calInfo[oh]['slope'][vfat]*event.dacValX_Err+calInfo[oh]['intercept'][vfat]
-        
+
         if args.assignXErrors:
             dict_dacScanResults[oh][vfat].SetPoint(dict_dacScanResults[oh][vfat].GetN(),event.dacValY,calibrated_DAC_value)
             dict_dacScanResults[oh][vfat].SetPointError(dict_dacScanResults[oh][vfat].GetN()-1,event.dacValY_Err,calibrated_DAC_error)
--- a/anaDACScan.py
+++ b/anaDACScan.py
@@ -260,6 +260,7 @@ if __name__ == '__main__':
                 
                 if maxVfat3DACSize[dacSelect][1] == nameX:
                     maxDacValue = int(maxVfat3DACSize[dacSelect][0])
+                    break
                     
             #evaluate the fitted function at the nominal current or voltage value and convert to an integer
             nominalDacValue = int(dict_DACvsADC_Funcs[oh][vfat].Eval(nominal))
--- a/anaDACScan.py
+++ b/anaDACScan.py
@@ -1,0 +1,82 @@
+#!/bin/env python
+
+"""
+anaDACScan
+==============
+"""
+
+import argparse
+import numpy as np
+import root_numpy as rp
+import ROOT as r
+
+from gempython.gemplotting.utils.anautilities import parseCalFile
+
+from gempython.utils.nesteddict import nesteddict
+
+from gempython.gemplotting.utils.anautilities import make3x8Canvas
+
+if __name__ == '__main__':
+
+    parser = argparse.ArgumentParser(description='Arguments to supply to anaDACScan.py')
+
+    parser.add_argument('infilename', metavar='infilename', type=str, help="Filename from which input information is read")
+#    parser.add_argument("calFile", metavar="calFile", type=str,help="File specifying CAL_DAC/VCAL to fC equations per VFAT")
+    parser.add_argument('-o','--outfilename', metavar='outfilename', type=str, help="Filename to which output information is written", default="DACFitData.root")
+
+    args = parser.parse_args()
+
+    print("Analyzing: '%s'"%args.infilename)
+
+    # Set default histogram behavior
+    r.TH1.SetDefaultSumw2(False)
+    r.gROOT.SetBatch(True)
+    r.gStyle.SetOptStat(1111111)
+
+    dacScanFile = r.TFile(args.infilename)
+    
+    # Determine which links present
+    list_bNames = ['link']
+    ohArray = rp.tree2array(tree=dacScanFile.dacScanTree, branches=list_bNames)
+    ohArray = np.unique(ohArray['link'])
+
+    # Determine which DAC was scanned and against which ADC
+    nameX = ""
+    nameY = ""
+    for event in dacScanFile.dacScanTree:
+        nameX = event.nameX.data()
+        nameY = event.nameY.data()
+        break # all entries will be the same
+
+    # Initialize nested dictionaries
+    dict_dacScanResults = nesteddict()
+    dict_dacScanFuncs = nesteddict()
+
+    # Get ADC0 or ADC1 calibration
+    # tuple_calInfo = parseCalFile(args.calFile)
+    
+    # Initialize a TGraphErrors for each vfat
+    for link in ohArray:
+        for vfat in range(0,24):
+             dict_dacScanResults[link][vfat] = r.TGraphErrors()
+             
+    # Loop over events in the tree and fill plots         
+    for event in dacScanFile.dacScanTree:
+        link = event.link
+        vfat = event.vfatN
+
+        dict_dacScanResults[link][vfat].SetPoint(dict_dacScanResults[link][vfat].GetN(),event.dacValY,event.dacValX)
+        dict_dacScanResults[link][vfat].SetPointError(dict_dacScanResults[link][vfat].GetN()-1,event.dacValY_Err,event.dacValX_Err)
+
+    # Fit the TGraphErrors objects    
+    for oh in ohArray:
+        for vfat in range(0,24):
+            dict_dacScanFuncs[oh][vfat] = r.TF1("DAC Scan Function","[0]*x+[1]")
+            dict_dacScanResults[oh][vfat].Fit(dict_dacScanFuncs[oh][vfat]) 
+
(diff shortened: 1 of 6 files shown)


bdorney:
Is this a work in progress or is it ready for review?

AndrewLevin:
@bdorney, it is a work in progress

now, the only thing missing is the use of the correct nominal value for all of the dacSelect options

the tables in the vfat3 manual do not actually have rows for each of the dacSelect values that you wrote in https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/155

to help me with this, can you tell me the nominal value for CFG_BIAS_SH_I_BFCAS?

anaDACScan.py
```diff
@@ -0,0 +1,137 @@
+#!/usr/bin/env python
+
+"""
+anaDACScan
+==============
+"""
```
bdorney: You should add a help string here; this will help `sphinx` automatically build the html documentation for the repo, see for example:

https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/packageFiles4Docker.py#L3-L98

anaDACScan.py
```diff
@@ -0,0 +1,137 @@
...
+import ROOT as r
+
+from gempython.gemplotting.utils.anautilities import parseCalFile
+from gempython.utils.nesteddict import nesteddict
+from gempython.gemplotting.utils.anautilities import make3x8Canvas
+
+import argparse
```
bdorney: You have no other functions declared in this script, put these imports inside `main`

anaDACScan.py
```diff
@@ -0,0 +1,137 @@
...
+
+if __name__ == '__main__':
+
+    parser = argparse.ArgumentParser(description='Arguments to supply to anaDACScan.py')
+
+    parser.add_argument('infilename', type=str, help="Filename from which input information is read", metavar='infilename')
+    parser.add_argument("calFileFile", type=str, help="File specifying which calFile to use for each OH. Format: OH1 filenameOH1 <newline> OH2 filenameOH2 <newline> etc", metavar="calFileFile")
```
bdorney: per our discussion make this optional and if not provided have a look up in `DATA_PATH` per the example described [here](https://mattermost.web.cern.ch/cms-gem-daq/pl/o5aknmyn6tdgun6mzw1gmza44c).

bdorney: Additionally I would rename this to just `--calFileList`.  And in the help string I would specify the format of this file in a more detailed way (for example how it is done in [plot_eff.py](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/d7b78cef238894998fc350377c3e4ace1a0b4484/macros/plot_eff.py#L72-L101)).

anaDACScan.py
```diff
@@ -0,0 +1,142 @@
...
+        outputFiles[link] = r.TFile("$DATA_PATH/"+chamber_config[link]+"/dacScans/scandate/"+args.outfilename,'recreate')
+             
+    # Loop over events in the tree and fill plots
+    for event in dacScanFile.dacScanTree:
+        link = event.link
+        vfat = event.vfatN
+
+        if link not in calInfo.keys():
+            print('Error: calibration file for link %i was not provided'%link)
+            exit(1)
```
bdorney: Why `exit()`?  The other links may still be valid.  I would add either a `raw_input` prompt for the user to provide a calibration file for this link (adding it to `calInfo`), or allow the user to state "ignore" in which this link is always skipped.  If this is too challenging then a simple `continue` instead of `exit()`.

anaDACScan.py
AndrewLevin: done

anaDACScan.py
AndrewLevin: done

anaDACScan.py
AndrewLevin: This logic is unnecessary now, since I now remove OHs without a calFile from the OH array.

bdorney:
Commits here don’t seem to reflect your comment. Perhaps some commits where not pushed to your remote? Also rebase on top of develop branch, and when ready remove WIP.  I’ll review after WIP is removed.

You’ll need to contact VFAT3 team for question regarding BFCAS.

AndrewLevin:
@bdorney, I rebased it (although given that I only added a new file, I am not sure why that would matter)

> Commits here don’t seem to reflect your comment. Perhaps some commits where not pushed to your remote? 

Can you be more specific? All commits were pushed to the remote.

> You’ll need to contact VFAT3 team for question regarding BFCAS.

I just sent an e-mail to Paul and Henri with you in cc.

anaDACScan.py
```diff
@@ -0,0 +1,236 @@
...
+    
+    dataPath = os.getenv("DATA_PATH") 
+    
+    if len(args.infilename.split('/')) > 1 and len(args.infilename.split('/')[len(args.infilename.split('/')) - 2].split('.')) == 5:
+        scandate = args.infilename.split('/')[len(args.infilename.split('/')) - 2]
+    else:    
+        scandate = 'noscandate'
+
+    for oh in ohArray:
+        os.system("mkdir -p "+ dataPath+"/"+chamber_config[oh]+"/dacScans/"+scandate)
```
bdorney: if `scandate == noscandate` then I wouldn't call this but instead write to `ELOG_PATH/chamber_config[oh]`

anaDACScan.py
```diff
@@ -0,0 +1,236 @@
...
+            calDacCalFile = "{0}/{1}/calFile_calDac_{1}.txt".format(dataPath,chamber_config[oh])
+            calDacCalFileExists = os.path.isfile(calDacCalFile)
+            if not calDacCalFileExists:
+                print("Skipping OH{0}, detector {1}, missing CFG_CAL_DAC Calibration file:\n\t{2}".format(
+                    oh,
+                    chamber_config[oh],
+                    calDacCalFile))
+                ohArray = np.delete(ohArray,(oh))
+
+
```
bdorney: unnecessary white space

anaDACScan.py
```diff
@@ -0,0 +1,236 @@
...
+    dict_dacScanResults = nesteddict()
+    dict_dacScanFuncs = nesteddict()
+
+    # Initialize a TGraphErrors for each vfat
+    for oh in ohArray:
+        for vfat in range(0,24):
+             dict_dacScanResults[oh][vfat] = r.TGraphErrors()
+             #the reversal of x and y is intended - we want to plot the nameX variable on the y-axis and the nameY variable on the x-axis
+             dict_dacScanResults[oh][vfat].GetXaxis().SetTitle(nameY)
+             dict_dacScanResults[oh][vfat].GetYaxis().SetTitle(nameX)
```
bdorney: It would also be good to have the actual DAC scan plotted e.g. `ADCX vs <reg>`.  Looking at the raw data helps to understand if the further results should be considered valid.

bdorney: I would also clarify the comment in line 158; e.g. why is this reversal being done?  From here it's not illustrated tha the goal is to obtain nominal value of register `nameX`.

anaDACScan.py
```diff
@@ -0,0 +1,236 @@
...
+        outputFiles[oh] = r.TFile(dataPath+"/"+chamber_config[oh]+"/dacScans/"+scandate+"/"+args.outfilename,'recreate')
+             
+    # Loop over events in the tree and fill plots
+    for event in dacScanFile.dacScanTree:
+        oh = event.link
+        vfat = event.vfatN
+
+        #the reversal of x and y is intended - we want to plot the nameX variable on the y-axis and the nameY variable on the x-axis
+        calibrated_DAC_value=calInfo[oh]['slope'][vfat]*event.dacValX+calInfo[oh]['intercept'][vfat]
+        calibrated_DAC_error=calInfo[oh]['slope'][vfat]*event.dacValX_Err+calInfo[oh]['intercept'][vfat]
```
bdorney: This is incorrect.  The calibration information is for `ADCX` but `dacValX` and `dacValX_Err` are for the register the DAC scan is being performed on.  You're applying this incorrectly.

You need to first be plotting `ADCX vs. regName` where `ADCX` is in physical units, e.g. voltage or current units.  Then we need a function for `f(V or I)` to obtain the nominal DAC values from.  This would require three `TObject`s per `(vfat,oh)` pair:

- `ADCX vs DAC` where `ADCX` is terms of ADC counts or physical units
- `DAC vs. ADCX` where `ADCX` is in terms of physical units, this is the plot that's actually being fit
- `TF1` that is fit to `DAC vs. ADCX`

anaDACScan.py
```diff
@@ -0,0 +1,236 @@
...
+            dict_dacScanResults[oh][vfat].Fit(dict_dacScanFuncs[oh][vfat]) 
+
+    nominal = -1        
+
+    #from Tables 9 and 10 of the VFAT3 manual 
+    if nameX == "CFG_THR_ARM_DAC":
+        nominal = 3.2
+    else:
+        print("Error: unexpected value of nameX: '%s'"%nameX)
+        exit(1)
```
bdorney: Lines 189 to 196 probably just make this into a dictionary and store it in `utils/anaInfo.py`; the key value should be the register name.  Then you can just do a check if `nameX in dict.keys()` and then get the nominal value by passing `nameX` as the key.

anaDACScan.py
```diff
@@ -0,0 +1,236 @@
...
+        vfatN_branch = np.zeros(1, dtype=int)
+        dacVal_branch = np.zeros(1, dtype=int)
+        tree_dacVals[oh].Branch('vfatN',vfatN_branch,'vfatN/i')
+        tree_dacVals[oh].Branch('dacVal',dacVal_branch,'dacVal/i')
+        for vfat in range(0,24):
+             dict_dacVals[oh][vfat] = dict_dacScanResults[oh][vfat].Eval(nominal)
+             graph_dacVals[oh].SetPoint(graph_dacVals[oh].GetN(),vfat,dict_dacVals[oh][vfat])
+             vfatN_branch[0] = vfat
+             dacVal_branch[0] = dict_dacVals[oh][vfat]
+             tree_dacVals[oh].Fill()
```
bdorney: Prefer this as a delimited text file; this allows a wider use case for other code/hardware (e.g. `ROOT` is not installed on the CTP7, outside `gempython` `cmsgemos` doesn't use `ROOT`).  You can write this to a file at the same time you're writing the table to terminal (e.g. lines 216-229).

anaDACScan.py
```diff
@@ -0,0 +1,236 @@
...
+            dict_dacScanResults[oh][vfat].SetPointError(dict_dacScanResults[oh][vfat].GetN()-1,event.dacValY_Err,calibrated_DAC_error)
+        else:    
+            dict_dacScanResults[oh][vfat].SetPoint(dict_dacScanResults[oh][vfat].GetN(),event.dacValY,calibrated_DAC_value)
+            dict_dacScanResults[oh][vfat].SetPointError(dict_dacScanResults[oh][vfat].GetN()-1,event.dacValY_Err,0)
+
+    # Fit the TGraphErrors objects    
+    for oh in ohArray:
+        for vfat in range(0,24):
+            dict_dacScanFuncs[oh][vfat] = r.TF1("DAC Scan Function","[0]*x+[1]")
+            dict_dacScanResults[oh][vfat].Fit(dict_dacScanFuncs[oh][vfat])
```
bdorney: Use fit option `"Q"` to not spam terminal.

anaDACScan.py
```diff
@@ -0,0 +1,236 @@
...
+    for oh in ohArray:
+        outputFiles[oh].cd()
+        graph_dacVals[oh].Write("dacValsOH"+str(oh))
+        tree_dacVals[oh].Write("dacValsOH"+str(oh))
+        for vfat in range(0,24):
+            print("| {0} | {1}  | {2} | ".format(
+                oh,
+                vfat,
+                dict_dacVals[oh][vfat]
+            ))
```
bdorney, line 332: Typically `TFile`s produced by scripts in `gem-plotting-tools` are organized with subdirectories to be easier to find objects within them.  The typical subdirectory structure is:

```
VFAT0
VFAT1
...
VFAT22
VFAT23
Summary
```

Where Summary contains summary information (e.g. lines 231-236).

Please add this structure to the output `TFile` objects.

anaDACScan.py
```diff
@@ -0,0 +1,236 @@
...
+        for vfat in range(0,24):
+             dict_dacVals[oh][vfat] = dict_dacScanResults[oh][vfat].Eval(nominal)
+             graph_dacVals[oh].SetPoint(graph_dacVals[oh].GetN(),vfat,dict_dacVals[oh][vfat])
+             vfatN_branch[0] = vfat
+             dacVal_branch[0] = dict_dacVals[oh][vfat]
+             tree_dacVals[oh].Fill() 
+             
+    # Write out the dacVal results to both a file and the terminal         
+    print( "| OH | vfatN | dacVal |")
+    print( "| -- | ----- | ------ |")
```
bdorney: To make reading this easier if pasted into mattermost change line 218 to:

```python
print("| :-: | :---: | :----: |")
```

This will ensure the columns are centered, see for example [markdown table documentation](https://docs.mattermost.com/help/messaging/formatting-text.html#tables).

anaDACScan.py
```diff
@@ -0,0 +1,236 @@
...
+            print("| {0} | {1}  | {2} | ".format(
+                oh,
+                vfat,
+                dict_dacVals[oh][vfat]
+            ))
+        
+    # Make plots        
+    for oh in ohArray:
+        canv_Summary = make3x8Canvas('canv_Summary',dict_dacScanResults[oh],'AP',dict_dacScanFuncs[oh],'')
+        canv_Summary.SaveAs(dataPath+"/"+chamber_config[oh]+"/dacScans/"+scandate+"/SummaryOH"+str(oh)+".png")
```
bdorney: if `scandate='noscadate'` then write this to `ELOG_PATH`.

anaDACScan.py
bdorney:
```python
elogPath = os.getenv('ELOG_PATH')
```

anaDACScan.py
bdorney: The value stored in this dictionary could be a tuple which has the nominal value and a string for the units, e.g.:

```python
nominalDacValues = {
   "CFG_IREF":(10,"uA"),
    ...
}
```

anaDACScan.py
bdorney: Most importantly you have not considered the handling of physical units correctly.  The cal_m and cal_b will place the ADC into voltage units.  However many front-end registers provide bias currents, not voltages.  So you will need to look at the register being scanned and determine if it is a bias voltage or current and calculate the physical unit appropriately as described [here](https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/155#issuecomment-428559337).  Please also note in Table 25 of the VFAT manual that bias currents that are not `CFG_IREF` need to have the `CFG_IREF` subtracted from the ADC reading.  This we can for the moment hardcode to `10uA` because it is assumed that the `CFG_IREF` will be set correctly for all other measurements.  

You can act on the physical unit stored in the `nominalDacValue` dictionary to switch between the "current unit" and "voltage unit" behavior by checking the unit in the stored tuple, see my comment [here](https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/158/files#r224981209).

anaDACScan.py
```diff
@@ -0,0 +1,236 @@
...
+
+    Use the dacValX_Err values stored in the input file
+
+.. option:: --calFileList
+  
+    Provide a file containing a list of calFiles in the following format. Example:: 
+
+    OH1 calFileOH1
+    OH2 calFileOH2
+    OH3 calFileOH3
```
bdorney: It's not clear from this example that the user should be entering an integer and then an absolute filename.  Change too:

```
0 /path/to/my/cal/file.txt
1 /path/to/my/cal/file.txt
...
...
```

anaDACScan.py
```diff
@@ -0,0 +1,236 @@
...
+--------
+
+**anaDACScan.py** [*OPTIONS*]
+
+Description
+-----------
+
+This script reads in calibration information and DAC scan data, performs a fit for each VFAT, computes the DAC value corresponding to the nominal current or voltage for each VFAT, and reports the results.
+
+
```
bdorney: unnecessary white space.

anaDACScan.py
```diff
@@ -0,0 +1,236 @@
...
+.. option:: --outfilename
+
+    Name of the output root file. Default is DACFitData.root.
+
+Example
+-------
+
+.. code-block:: bash
+
+    anaDACScan.py /afs/cern.ch/user/d/dorney/public/v3Hack/dacScanV3.root  --calFileList calibration.txt --assignXErrors
```
bdorney: Change filepath to:

```
/path/to/input.root
```

Never show someone's afs area on github.

anaDACScan.py
```diff
@@ -0,0 +1,236 @@
...
+
+Environment
+-----------
+
+
+Internals
+---------
+
+
+"""
```
bdorney, line 49: unnecessary white space, lines 50-59.

anaDACScan.py
```diff
@@ -0,0 +1,236 @@
...
+    from gempython.gemplotting.mapping.chamberInfo import chamber_config
+
+    import os
+    
+    import argparse
+    
+    parser = argparse.ArgumentParser(description='Arguments to supply to anaDACScan.py')
+
+    parser.add_argument('infilename', type=str, help="Filename from which input information is read", metavar='infilename')
+    parser.add_argument("--calFileList", type=str, help="File specifying which calFile to use for each OH. Format: OH1 filenameOH1 <newline> OH2 filenameOH2 <newline> etc", metavar="calFileList")
```
bdorney: Again it's not clear that `OH1` should be an integer from this `helpstring`.

anaDACScan.py
```diff
@@ -0,0 +1,236 @@
...
+
+    calInfo = {}
+
+    if args.calFileList != None:
+        for line in open(args.calFileList):
+            line = line.strip(' ').strip('\n')
+            first = line.split(' ')[0].strip(' ')
+            second = line.split(' ')[1].strip(' ')
+            tuple_calInfo = parseCalFile(second)
+            calInfo[int(first)] = {'slope' : tuple_calInfo[0], 'intercept' : tuple_calInfo[1]}
```
bdorney, line 163: Add an if statement that identifies the `#` character as a comment and to skip the line, e.g.:

```python
        if line[0] == "#":
            continue
```

bdorney:
When you clean this up please attach screenshots to the PR so we can see what plots have been produced.

anaDACScan.py
AndrewLevin: done

anaDACScan.py
AndrewLevin: done

anaDACScan.py
AndrewLevin: fixed

anaDACScan.py
AndrewLevin: done

anaDACScan.py
AndrewLevin: done

anaDACScan.py
AndrewLevin: done

anaDACScan.py
AndrewLevin: done

anaDACScan.py
AndrewLevin: done

anaDACScan.py
AndrewLevin: done

anaDACScan.py
AndrewLevin: fixed

anaDACScan.py
AndrewLevin: fixed

anaDACScan.py
AndrewLevin: done

anaDACScan.py
AndrewLevin: done

anaDACScan.py
AndrewLevin: @bdorney, how should the nominal DAC values be stored in the output root file? There is one nominal DAC value per VFAT. Should I store a single Float_t in each of the VFATX sub-directories?

AndrewLevin:
@bdorney, two questions:

1) In https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/203, CFG_THR_ARM_DAC and CFG_THR_ZCC_DAC are repeated:

14      CFG_THR_ARM_DAC
15      CFG_THR_ZCC_DAC
35      CFG_THR_ARM_DAC
36      CFG_THR_ZCC_DAC

Why? 

2) The calibrated ADC values are in what units (e.g. uA or nA)? I guess this depends on the calFile, which I don't have an real example of.

bdorney:
For question 1 this is illustrated in the VFAT3 manual, table 25.

For question 2 the calibration of the ADC is always in volts.  In the case of voltage the calibration equation is used, in the case of current you need to convert to volts, e.g. it's the current across a 20kOhm resistor.  If a monitoring current is being measured that is *not* the IREF than the current being monitored is the current of interest plus the IREF.  As described in table 25 of the VFAT3 manual.

While you *don't* need a calibration file for this exercise (e.g. set m=1 and b=0) here are two examples for VFAT3B:

```
% more vfat3CalInfo/vfat3b/vfat3_0061/calibration_data.dat

Adjusting the global reference current.
- Iref adjusted.
Register value: 28 .

Internal ADCs calibrated. Values:
ADC0: 2.035943 + -325.388101
ADC1: 2.301137 + -450.231461

CAL_DAC conversion completed.
CAL_DAC to fC: 0.215792 + -2.028758

Scan duration: 0.017219 min
```

And for VFAT3A:

```
% more vfat3CalInfo/vfat3a/hbrdnp011/calPars.reg
adc0M/D:adc0B/D:adc1M/D:adc1B/D:cal_dac_fcM/D:cal_dac_fcB/D:Iref/I
2.059961	-318.730307	0.823305	334.545018	0.275427	-2.187759	27
```

Note this is *unique* to each VFAT and there 5000+ so I don't have each value nor do I have a "characteristic" or "typical" value.

anaDACScan.py
bdorney: Why would you store this in the output root file?  You have the TF1 in the output root file, that's all that's needed.  Moreover the DAC value is an integer, not a float.

bdorney:
[VFAT3 manual](https://espace.cern.ch/cms-project-GEMElectronics/VFAT3/VFAT3%20User%20Manual%20v2.2.pdf)

anaDACScan.py
AndrewLevin: When I evaluate the TF1 at the nominal current or voltage, I get a float. That is the value I was planning to store in the output root file. It is written to the output text file and the terminal, but currently it is not stored in the root file. Furthermore, how should it be converted to an int? Just use floor?

AndrewLevin:
@bdorney, for the registers that are not CFG_IREF, it is not clear to me if the nominal value of CFG_IREF should be added or subtract. Which of the these is correct for current registers that are not CFG_IREF:

1) calibrate, then subtract the nominal current for CFG_IREF (10 uA), then convert to volts
2) calibrate, then add the nominal current for CFG_IREF (10 uA), then convert to volts.

anaDACScan.py
bdorney: But what is the purpose of storing it in the output `TFile`?  You have the `TF1` already which is much more useful.  Don't store it in the output `TFile`, you should be printing this information to a separate text file.

And as for converting just use `int()` which will handle the conversion and round appropriately.

AndrewLevin: done

anaDACScan.py
bdorney: still appear here, e.g. 178 & 179.  Perhaps change was not committed/pushed?

AndrewLevin:
@bdorney, in the example root file that you provided, the dacValYs are between 100.0 and 500.0, so calibrating them using a slope of 1 and an intercept of 0, and then dividing by 20000 gives values between 0.05 and 0.025. The nominal current for CFG_THR_ARM_DAC from the VFAT3 manual is 3.2 uA. Are we sure about the value of the resistance?

anaDACScan.py
AndrewLevin: sorry, fixed now

bdorney:
> @bdorney, in the example root file that you provided, the dacValYs are between 100.0 and 500.0, so calibrating them using a slope of 1 and an intercept of 0, and then dividing by 20000 gives values between 0.05 and 0.025. The nominal current for CFG_THR_ARM_DAC from the VFAT3 manual is 3.2 uA. Are we sure about the value of the resistance?

If you're using a dummy value of slope=1 and intercept of 0 then you're not in volts, but DAC units.

anaDACScan.py
```diff
@@ -0,0 +1,289 @@
...
+    import os
+    
+    import argparse
+    
+    parser = argparse.ArgumentParser(description='Arguments to supply to anaDACScan.py')
+
+    parser.add_argument('infilename', type=str, help="Filename from which input information is read", metavar='infilename')
+    parser.add_argument("--calFileList", type=str, help="File specifying which calFile to use for each OH. Format: 0 /path/to/my/cal/file.txt<newline> 1 /path/to/my/cal/file.txt<newline>...", metavar="calFileList")
+    parser.add_argument('-o','--outfilename', dest='outfilename', type=str, default="DACFitData.root", help="Filename to which output information is written", metavar='outfilename')
+    parser.add_argument('-t','--output_txtfile_filename', dest='output_txtfile_filename', type=str, default="NominalDACValues.txt", help="Name of output text file that will contain the nominal DAC values in tab-delimited format", metavar='output_txtfile_filename')
```
bdorney: I wouldn't have this option (`--output_txtfile_filename`), it can just be hardcoded like `chConfig.txt` and `vfatConfig.txt` are; how you have it, this file is written to a subdirectory either in `DATA_PATH` or `ELOG_PATH` that has the string `chamber_config[ohN]` in it.

bdorney:
Additionally I'd wager from Fig 29 in VFAT3 manual that the ADC calibration comes out in mV.

anaDACScan.py
```diff
@@ -0,0 +1,288 @@
...
+    ...
+    ...
+
+.. option:: --outfilename
+
+    Name of the output root file. Default is DACFitData.root.
+
+.. option:: --output_txtfile_filename
+
+    Name of output text file that will contain the nominal DAC values in tab-delimited format. Default is NominalDACValues.txt.
```
bdorney: drop lines 43-45

anaDACScan.py
AndrewLevin: the output text-file filename is hardcoded

AndrewLevin:
ok, 

> Additionally I'd wager from Fig 29 in VFAT3 manual that the ADC calibration comes out in mV.

I have taken this into account now

anaDACScan.py
```diff
@@ -0,0 +1,288 @@
...
+        else:    
+            outputFiles[oh] = r.TFile(dataPath+"/"+chamber_config[oh]+"/dacScans/"+scandate+"/"+args.outfilename,'recreate')
+
+    # Loop over events in the tree and fill plots
+    for event in dacScanFile.dacScanTree:
+        oh = event.link
+        vfat = event.vfatN
+
+        calibrated_ADC_value=calInfo[oh]['slope'][vfat]*event.dacValY+calInfo[oh]['intercept'][vfat]
+        calibrated_ADC_error=calInfo[oh]['slope'][vfat]*event.dacValY_Err+calInfo[oh]['intercept'][vfat]
```
bdorney: For calculating `dacValY_Err` in terms of voltage instead of ADC we don't want to use the intercept.  The error here is always the least sensitive bit (LSB), so this is an ADC value of 1, e.g. the ADC bin width. Applying the intercept will not give that (indeed using the sample calibration coef's I provided you get a hugely negative number).

anaDACScan.py
AndrewLevin: done

anaDACScan.py
AndrewLevin: fixed

anaDACScan.py
```diff
@@ -0,0 +1,288 @@
...
+
+        #Use Ohm's law to convert the currents to voltages. The VFAT3 team told us that a 20k ohm resistor was used.
+        if nominalDacValues[nameX][1][1] == "A":
+
+            calibrated_ADC_value = calibrated_ADC_value/20000.0
+            calibrated_ADC_error = calibrated_ADC_error/20000.0
+
+            if nameX != 'CFG_IREF':
+                calibrated_ADC_value -= nominal_iref
```
bdorney, line 236: I think you have a unit mismatch here.  You're assuming `calibrated_ADC_*` will be in volts (the calibration coef's I suspect are in mV) and so you're taking `mV / 20kOhm` and  assigning this to a current.  But then you're subtracting `nominal_iref` from it but the units on `nominal_iref` are not guaranteed to match the units on `calibrated_ADC_*`.  I think you need to double check this and implement a check in the code to ensure it comes out correct.  

This should also be around line 250 where you set `dict_dacVals`.  You should check the requested DAC value doesn't go outside the maximum range (otherwise it will roll over, e.g. 256 in an 8-bit register will be written as `0x1`.

I think you can check against [maxVfat3DacSize](https://github.com/cms-gem-daq-project/cmsgemos/blob/61538ef0b357202cef902edfd4169f81d1f25bc9/gempython/tools/amc_user_functions_xhal.py#L14).

If greater than the register max is returned it should at the bare minimum throw a warning of some kind.  And we could consider printing the register max instead (to prevent rollover).

utils/anaInfo.py
```diff
@@ -12,6 +12,25 @@
...
+        "CFG_BIAS_SD_I_BDIFF":(28,"uA"),
+        "CFG_BIAS_SD_I_BFCAS":(27,"uA"),
+        "CFG_BIAS_SD_I_BSF":(30,"uA"),
+        "CFG_BIAS_CFD_DAC_1":(20,"uA"),
+        "CFG_BIAS_CFD_DAC_2":(20,"uA"),
+        "CFG_HYST":(100,"nA"),
+        "CFG_THR_ARM_DAC":(3.2,"uA"),
+        "CFG_THR_ZCC_DAC":(275,"nA"),
+        "CFG_BIAS_PRE_VREF":(430,'mV'),
+        "CFG_ADC_VREF":(400,'mV')
```
bdorney: I think we've tried to guess at a few of these values and have paired them improperly.  For example `CFG_ADC_VREF` should be as close to 1.0V as possible, see VFAT3 manual, section 9.1.4.

Additionally these values appear to match for VFAT3b, in the future (not now) we should think how to handle VFAT3a.  Certainly this week we will need to...but hopefully we can get some input from the VFAT3 designers.

anaDACScan.py
AndrewLevin: ok, the conversion to V from mV was done in commit d52bfa5b505d2ec8a1cacde0d980e9f777d97f29

I would need to update my gempython to get the function maxVfat3DacSize. Do you recommend I do that? Or I can just use 255 as the maximum value for each register.

utils/anaInfo.py
AndrewLevin: yes, this is based on the pairing in your e-mail to me, I cannot verify that it is correct

bdorney: Yes I guessed as much, but please review the table I sent for accuracy.  For example you've paired `CFG_ADC_VREF`, nominal value 1.0V as mentioned, incorrectly to the nominal value of `SLVS_Vref` e.g. 400mV.  This should be resolved and a double check of the rest of the table just to make sure.

anaDACScan.py
bdorney: > ok, the conversion to V from mV was done in commit d52bfa5

That wasn't what I was really getting at.  Say someone is scanning `CFG_BIAS_PRE_I_BIT`.  The current here will be this bias current plus the IREF current.  But looking at lines 124-148 you've correctly enabled `nominal_iref` to be changed from mA, uA, or nA.  But the ADC will *always* read in the voltage unit it's designed to deliver in.  So you could get in a situation where `nominal_iref` is in `nA` but the reading coming from the ADC is converted into `uA` then say 10000 nA subtracted from 32 uA will in this case result in a number that is -9968.  

I think it's better to just stick with `uA` and `mV` inform the user these are the units expected, and convert all values to these (especially below in `nominalDacValues`).

> I would need to update my gempython to get the function maxVfat3DacSize. Do you recommend I do that? Or I can just use 255 as the maximum value for each register.

Don't do this, not every register is an 8 bit register!!  This will cause exactly what we are trying to avoid, rollover.

Build [cmsgemos_gempython](https://mattermost.web.cern.ch/cms-gem-daq/pl/3woxpxxsfpb4tk571bt3x5bw1h)

AndrewLevin:
@bdorney, for the two registers

CFG_THR_ARM_DAC
CFG_THR_ZCC_DAC

isn't it ambiguous if they correspond to a voltage or a current? Possibly the dacSelect number needs to be stored also?

bdorney:
> @bdorney, for the two registers
CFG_THR_ARM_DAC
CFG_THR_ZCC_DAC
isn't it ambiguous if they correspond to a voltage or a current? Possibly the dacSelect number needs to be stored also?

Yup that's a good point.  For now just consider the voltage case.  Maybe this week if you have free time in the camp you can open a GitHub issue for [dacScanV3.py](https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/rpc-playground/dacScanV3.py) to store the `dacSelect` in the output `TTree` (and even make the PR!).

anaDACScan.py
AndrewLevin: I have implemented an upper and a lower check of the nominal DAC value now.

anaDACScan.py
```diff
@@ -0,0 +1,311 @@
...
+    for event in dacScanFile.dacScanTree:
+        oh = event.link
+        vfat = event.vfatN
+
+        calibrated_ADC_value=calInfo[oh]['slope'][vfat]*event.dacValY+calInfo[oh]['intercept'][vfat]
+        calibrated_ADC_error=calInfo[oh]['slope'][vfat]*event.dacValY_Err
+
+        #from Table 29 of the VFAT3 manual, we are guessing the calibrated voltage is in mV
+        calibrated_ADC_value /= 1000.0
+        calibrated_ADC_error /= 1000.0
```
bdorney: I believe the `calInfo[oh]['slope']` and `calInfo[oh]['intercept']` are in `mV/ADC` and `mV` respectively.  Can you cross-check with Henri tomorrow?

anaDACScan.py
```diff
@@ -0,0 +1,311 @@
...
+            
+            for dacSelect in maxVfat3DACSize.keys():
+
+                #the registers CFG_THR_ARM_DAC and CFG_THR_ZCC_DAC could correspond to voltages or currents
+                #we will use voltages until a way of distinguishing the two cases is implemented 
+                if dacSelect == 14 or dacSelect == 15:
+                    continue
+                
+                if maxVfat3DACSize[dacSelect][1] == nameX:
+                    maxDacValue = int(maxVfat3DACSize[dacSelect][0])
```
bdorney, line 285: there should be a `break` after this; e.g. once this is found why continue to loop?

AndrewLevin: I added a break

anaDACScan.py
bdorney: > I have implemented an upper and a lower check of the nominal DAC value now.

I see that, but you didn't really address my main concern.  The units here may mismatch and you haven't made any check against that.  There should be some check which *prevents* subtract a number in `nA` from a number in `uA`.

anaDACScan.py
AndrewLevin: yes, I will ask him tomorrow if I get a chance or send him an e-mail

anaDACScan.py
bdorney: The easiest way to address this would be:

> I think it's better to just stick with uA and mV inform the user these are the units expected, and convert all values to these (especially below in nominalDacValues).

You could inform them in the help menu.

anaDACScan.py
```diff
@@ -0,0 +1,311 @@
...
+    import os
+    
+    import argparse
+    
+    parser = argparse.ArgumentParser(description='Arguments to supply to anaDACScan.py')
+
+    parser.add_argument('infilename', type=str, help="Filename from which input information is read", metavar='infilename')
+    parser.add_argument("--calFileList", type=str, help="File specifying which calFile to use for each OH. Format: 0 /path/to/my/cal/file.txt<newline> 1 /path/to/my/cal/file.txt<newline>...", metavar="calFileList")
+    parser.add_argument('-o','--outfilename', dest='outfilename', type=str, default="DACFitData.root", help="Filename to which output information is written", metavar='outfilename')
+    parser.add_argument('--assignXErrors', dest='assignXErrors', action='store_true', help="Whether to assign errors for the X variable (which is actually plotted on the y-axis)")
```
bdorney: You're making two sets of plots now `e.g. `ADC vs DAC` and `DAC vs. ADC` this help menu for `--asignXErrors` is a bit confusing.  What are you trying to tell the user?  Should improve clarity, any change here should also go in the `helpstring` above.

bdorney (changes requested):
Still a concern over the units issue.  Fix to `uA` and `mV` and do not allow other units.  Also please post screen shots of all the produced plots, even those obtained with dummy calibration coefficients.

anaDACScan.py
bdorney: Great, thanks, I think this will help resolve the units issue.

anaDACScan.py
AndrewLevin: I understand what you mean. I have corrected both help messages.

anaDACScan.py
AndrewLevin: sorry, I don't understand what the problem is or what you would like me to do

According to the VFAT3 register the nominal current of the current reference register is 10 uA, and this is independent of anything related to the other registers. Is there something I am missing about Table 9 of the VFAT3 manual?

> So you could get in a situation where `nominal_iref` is in `nA` but the reading coming from the ADC is converted into `uA` then say 10000 nA subtracted from 32 uA will in this case result in a number that is -9968.

32 uA - 10000 nA = 22 uA, and this is correctly done in my script, I convert everything to amperes before doing the subtraction

In the VFAFT3 manual, there are two registers for which the nominal current is in nA, one register for which the nominal voltage is in V. The rest are in uA or mV. 

I can just remove those three registers from the table or exit(1) if the user tries to use the script with one of those three registers. That seems to be what is implied from your comments. Could you please clarify?

bdorney: > I convert everything to amperes before doing the subtraction

Okay it was not clear to me, but this is good...guess I've been starting at a screen for too long today.

bdorney:
Once [this](
https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/158#discussion_r225006402) comment is resolved I'll force merge this PR and build the rpm's for the camp.  I understood Henri is coming tomorrow so we should be in time.  Otherwise try to reach him by email before 11h00.

bdorney:
Thanks for working through the weekend on this!

bdorney:
Given Henri's recent response we expect `mV` or `uA` please ensure all output plots are in these units.

AndrewLevin:
@bdorney, Henri only confirmed the units of the voltage calibration constants, should I also ask him about the units of the current calibration constants?

Also, I think I may be doing things in the wrong order:

https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/0abcc751eb7ed5adeaee10157ae46a607cb7393f/anaDACScan.py#L219-L223

The correct order for a current register is:

1) calibrate
2) subtract the nominal current
3) use Ohm's law to convert to volts,

right?

bdorney:
> @bdorney, Henri only confirmed the units of the voltage calibration constants, should I also ask him about the units of the current calibration constants?

There are no current calibration coefficients, and ADC *always* reads voltage.  A bias current is converted to a voltage by the use of a resistor (e.g. the 20kOhm resistor).

After everything is said and done bias voltage DAC scans should have their physical unit by `mV`.  Bias current DAC scans should have their physical unit by `uA`.

>The correct order for a current register is:
calibrate
subtract the nominal current
use Ohm's law to convert to volts,
right?

No, the ADC is always in voltage units.

1. Convert ADC counts to milli volts,
2. Convert milli volts to micro amps,
3. Subtract IREF from those bias currents that are *not* IREF.

AndrewLevin:
@bdorney, ok, now all of the plots are in mV and uA

AndrewLevin:
The errors on the dacValY variable, which is plotted on the x-axis in the DAC vs ADC plots, have been removed now. ROOT's fitting method does not converge when these errors are used, regardless of what they are set to. I tried a variety of tweaks to the fitting method and the initial values of the polynomial coefficients, but nothing helps at all. In any case, we have such a large number of points and these errors are small enough that their effect would be negligible, at least in the cases I have considered. The fits with the fifth degree polynomial are extremely good when using 0 for the error on the dacValY variable (see the screenshots).

bdorney:
Please add units to access labels where appropriate (e.g. `mV` or `#muA`).

Additionally there's no labels on the axes of `NominalDacValVsVFATX`.  Please add them, with units.  Additionally restrict this plot's y-axis range to something physical (e.g. negative numbers should not be displayed).

Drawing option `APE1` is strongly preferred.  Lines should not be drawn between points, only fit curves should be draw.

For `DACvsADC` increase the point size (or decrease the `TF1` width) so we can see how the points and the fit are matching.  Perhaps use a dotted line style for the fit.

Generally the over-arching question when developing a tool for general use is: "can someone with little experience understand what is being plotted?"

anaDACScan.py
```diff
@@ -0,0 +1,318 @@
...
+            line = line.strip(' ').strip('\n')
+            first = line.split(' ')[0].strip(' ')
+            second = line.split(' ')[1].strip(' ')
+            tuple_calInfo = parseCalFile(second)
+            calInfo[int(first)] = {'slope' : tuple_calInfo[0], 'intercept' : tuple_calInfo[1]}
+
+    #for each OH, check if calibration files were provided, if not search for the calFile in the $DATA_PATH, if it is not there, then skip that OH for the rest of the script
+    for oh in ohArray:
+        if oh not in calInfo.keys():
+            calDacCalFile = "{0}/{1}/calFile_calDac_{1}.txt".format(dataPath,chamber_config[oh])
```
bdorney: why `calDAC`?  The cal file is not for the `CFG_CAL_DAC` register.  We are calibrating the ADC, e.g.:

```python
`{0}/{1}/calFile_{2}_{1}.txt".format(dataPath,chamber_config[ohN],nameY)
```

anaDACScan.py
```diff
@@ -0,0 +1,318 @@
...
+            calDacCalFileExists = os.path.isfile(calDacCalFile)
+            if not calDacCalFileExists:
+                print("Skipping OH{0}, detector {1}, missing CFG_CAL_DAC Calibration file:\n\t{2}".format(
```
bdorney: again not `CFG_CAL_DAC`

bdorney:
We can also resolve the issue about not knowing if `CFG_THR_ARM_DAC` is a bias current or voltage by co-resolving: https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/206

lmoureaux:
You need to add `doc/man/anaDACScan.rst` and the corresponding line to `doc/man-index.rst` in order for your documentation to be built correctly

AndrewLevin:
@lmoureaux, the two files are added now

doc/man-index.rst
```diff
@@ -56,6 +56,7 @@ documented elsewhere.
     * :program:`ana_scans.py`
     * :program:`anaSBitTresh.py`
     * :program:`anaUltraLatency.py`
+    * :program:`anaDACScan.py`
```
lmoureaux: This will add `anaDACScan.py` to the list of undocumented tools. Since you wrote some documentation (great!), I'd add to "Analysis tools" toctree (as `man/anaDACScan`). I case you wonder, you can find a rendering of the corresponding page [here](http://homepage.iihe.ac.be/~lmoureau/GEM/sphinx/html/man-index.html) (unfortunately I don't update it very often).

AndrewLevin: done

lmoureaux:
Issues noticed when installing in a `virtualenv`:

* The `man` page is not generated because `doc/man/anaDACscan.rst` should be named `doc/man/anaDACScan.rst` instead (uppercase `S`, the same as in `doc/man-index.rst`
* An entry should be added to `doc/conf.py` for `man` page around line 254 
* For your script to be included in the packages, an additional line is needed in the `Makefile` around line 56

AndrewLevin:
@lmoureaux, fixed

AndrewLevin:
@bdorney, the scaling factors in Table 12 and Table 13 of the VFAT3 manual are now applied

anaDACScan.py
```diff
@@ -0,0 +1,321 @@
...
+            line = line.strip(' ').strip('\n')
+            first = line.split(' ')[0].strip(' ')
+            second = line.split(' ')[1].strip(' ')
+            tuple_calInfo = parseCalFile(second)
+            calInfo[int(first)] = {'slope' : tuple_calInfo[0], 'intercept' : tuple_calInfo[1]}
+
+    #for each OH, check if calibration files were provided, if not search for the calFile in the $DATA_PATH, if it is not there, then skip that OH for the rest of the script
+    for oh in ohArray:
+        if oh not in calInfo.keys():
+            calAdcCalFile = "{0}/{1}/calFile_calAdc_{1}.txt".format(dataPath,chamber_config[oh])
```
bdorney: This doesn't match the previous comment, this should explicitly be what was shown [here](https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/158#discussion_r225479887).

anaDACScan.py
bdorney: Still needs to be changed...

Additionally please review *all* comments before pushing commits to the remote as pushing new changes will *hide* potential other comments.

Additionally don't push to the remote after *every* commit...this creates a lot of spam emails and it's hard to filter what was *actually changed*.  Review all comments, make all changes, test, then push to the remote.

bdorney: you can include `nameY` in the `format(...)` argument to replace `CFG_CAL_DAC` here.

anaDACScan.py
```diff
@@ -0,0 +1,321 @@
...
+        if args.assignXErrors:
+            dict_DACvsADC_Graphs[oh][vfat].SetPoint(dict_DACvsADC_Graphs[oh][vfat].GetN(),calibrated_ADC_value,event.dacValX)
+            dict_RawADCvsDAC_Graphs[oh][vfat].SetPoint(dict_RawADCvsDAC_Graphs[oh][vfat].GetN(),event.dacValX,event.dacValY)
+            dict_DACvsADC_Graphs[oh][vfat].SetPointError(dict_DACvsADC_Graphs[oh][vfat].GetN()-1,0,event.dacValX_Err)
+            dict_RawADCvsDAC_Graphs[oh][vfat].SetPointError(dict_RawADCvsDAC_Graphs[oh][vfat].GetN()-1,event.dacValX_Err,0)
+        else:
+            dict_DACvsADC_Graphs[oh][vfat].SetPoint(dict_DACvsADC_Graphs[oh][vfat].GetN(),calibrated_ADC_value,event.dacValX)
+            dict_RawADCvsDAC_Graphs[oh][vfat].SetPoint(dict_RawADCvsDAC_Graphs[oh][vfat].GetN(),event.dacValX,event.dacValY)
+            dict_DACvsADC_Graphs[oh][vfat].SetPointError(dict_DACvsADC_Graphs[oh][vfat].GetN()-1,0,0)
+            dict_RawADCvsDAC_Graphs[oh][vfat].SetPointError(dict_RawADCvsDAC_Graphs[oh][vfat].GetN()-1,0,0)
```
bdorney, line 251: lines 229-238, you should always be applying `calibrated_ADC_error` here.  Why is this not applied?

anaDACScan.py
```diff
@@ -0,0 +1,321 @@
...
+
+.. program:: anaDACScan.py
+
+.. option:: infilename
+
+    Name of the input file root.
+
+.. option:: --assignXErrors  
+
+    If this flag is set, the dacValX_Err values stored in the input file are used when constructing the plots and performing the fits. If this flag is not set, the error associated with the dacValX values is always taken to be 0. Note that these errors are actually errors on the variable plotted on the y-axis in the DAC vs ADC plots. The dacValY_Err values, which would correspond to the errors on the x-axis in the DAC vs ADC plots, cannot be used because fits do not converge when they are used.
```
bdorney: Line 28 is too confusing.  Re-write:

```
If this flag is set then an uncertain on the DAC register value is assumed, otherwise the DAC register value is assumed to be a fixed unchanging value (almost always the case). 
```

anaDACScan.py
```diff
@@ -0,0 +1,321 @@
...
+    import os
+    
+    import argparse
+    
+    parser = argparse.ArgumentParser(description='Arguments to supply to anaDACScan.py')
+
+    parser.add_argument('infilename', type=str, help="Filename from which input information is read", metavar='infilename')
+    parser.add_argument("--calFileList", type=str, help="File specifying which calFile to use for each OH. Format: 0 /path/to/my/cal/file.txt<newline> 1 /path/to/my/cal/file.txt<newline>...", metavar="calFileList")
+    parser.add_argument('-o','--outfilename', dest='outfilename', type=str, default="DACFitData.root", help="Filename to which output information is written", metavar='outfilename')
+    parser.add_argument('--assignXErrors', dest='assignXErrors', action='store_true', help="dacValY_Err values are always used, but dacValX_Err values are used if and only if this flag is set")
```
bdorney: This doesn't match line 28, change to match my comment for line 28.

anaDACScan.py
AndrewLevin: please see: https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/158#issuecomment-430004251

bdorney: Make a dummy plot if necessary for fitting, the errors *have physical meaning* and should be shown.

bdorney: Additionally if the user applies a `--stepSize` argument and chooses something other than 1 (e.g. 10, or 25) the number of points will drastically decrease.

anaDACScan.py
AndrewLevin: done

anaDACScan.py
AndrewLevin: done

anaDACScan.py
AndrewLevin: done

anaDACScan.py
AndrewLevin: OK, so the procedure will be:

1) try to fit with the dacValY errors
2) check if the fit converges
3) if not, fit without the dacValY errors

In all cases, the output plots will show the dacValY errors.

bdorney: Sounds good.

You can also use the fit option “M” to have the minimized in Minuit try “more” to look for the solution.

AndrewLevin: @bdorney, I am now just using option 'EX0' in the fits

AndrewLevin:
@bdorney, I am now masking vfats that have a constant dacValY that is equal to 0

anaDACScan.py
```diff
@@ -0,0 +1,345 @@
...
+
+    #list of (oh,vfat) pairs to mask because they have a constant dacValyY value of 0, indicating they are not connecting
+    mask = []
+    
+    for oh in ohArray:
+        for vfat in range(0,24):
+            array = rp.tree2array(tree=dacScanFile.dacScanTree, branches=['dacValY'],selection='link == '+str(oh)+' && vfatN == '+str(vfat))
+            array = np.unique(array['dacValY'])
+            if len(array) == 1 and array[0] == 0:
+                mask.append((oh,vfat))
```
bdorney: Lines 88 - 99.

This is a bit inefficient, you're looping N * 24 times (where N could be up to 12) for a total of 288 times.  Each time you're performing an I/O operation (plus the original I/O operation at line 88).  This can be done in a single IO operation:

```python
list_bNames = ['vfatN', 'link', 'dacValY']
vfatMasks = rp.tree2array(tree=dacScanFile.dacScanTree,branches=list_bNames)
vfatMasks = np.unique(vfatMasks)
vfatMasks['dacValY' == 0] # this should return a numpy array which contains (oh,vfat) pairs for which `dacValY` is always zero 
ohArray = np.unique(ohArray['link']) # this is your ohArray above
```

No loops, only one I/O operation.

anaDACScan.py
```diff
@@ -0,0 +1,345 @@
...
+    dict_DACvsADC_Graphs = nesteddict()
+    dict_RawADCvsDAC_Graphs = nesteddict()
+    dict_DACvsADC_Funcs = nesteddict()
+
+    # Initialize a TGraphErrors and a TF1 for each vfat
+    for oh in ohArray:
+        for vfat in range(0,24):
+            dict_RawADCvsDAC_Graphs[oh][vfat] = r.TGraphErrors()
+            dict_RawADCvsDAC_Graphs[oh][vfat].GetXaxis().SetTitle(nameX)
+            dict_RawADCvsDAC_Graphs[oh][vfat].GetYaxis().SetTitle(nameY)
```
bdorney, line 192: to assist with reading the 3x8 grid plots (especially for people who are not familiar with the chip layout), please add:

```python
dict_RawADCvsDAC_Graphs[oh][vfat].SetTitle("VFAT{}".format(vfat))
```

For all plots.

anaDACScan.py
AndrewLevin: @bdorney, vfatMasks['dacValY' == 0] is always empty when I use this code

are you sure this is right?

bdorney: I did it off the top of my head; so probably some debugging is required.  But you see the idea.  You can get *all* the data in a single I/O transaction, and then manipulate the array you have.

anaDACScan.py
AndrewLevin: done

anaDACScan.py
AndrewLevin: @bdorney, I cannot find a numpy operation that does what we would like 'dacValY' == 0 to do

I just loaded the ttree into memory using LoadBaskets

bdorney: > @bdorney, I cannot find a numpy operation that does what we would like 'dacValY' == 0 to do

What's a scandate you're using?  I'll try.

anaDACScan.py
bdorney: doesn't seem like the change is reflected here.

AndrewLevin: the 3x8 grid is for the calibrated, not raw, plots

see this commit: https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/158/commits/8974a78b647e25177e7e5461beead27bfbcca53e

anaDACScan.py
AndrewLevin: 2018.10.17.11.03 on gem904daq01

bdorney: My original syntax was incorrect, but here is a solution that does not require the inner loop and still only reads from the `TTree` once:

```python
import root_numpy as rp
import numpy as np
import ROOT as r
dacFile = r.TFile("/data/bigdisk/GEM-Data-Taking/GE11_QC8/dacScanV3/2018.10.17.11.03/dacScanV3.root")
list_bNames = ['vfatN','link','dacValY']
vfatArray = rp.tree2array(tree=dacFile.dacScanTree,branches=list_bNames)
ohArray = np.unique(vfatArray['link'])
dict_nonzeroVFATs = {}
for ohN in ohArray:
    dict_nonzeroVFATs[ohN] = np.unique(vfatArray[vfatArray[vfatArray['dacValY'] > 0]['link'] == ohN]['vfatN'])
```

Then `ohArray` is:

```python
In [23]: ohArray
Out[23]: array([0, 1], dtype=int32)
```

And nonzero VFATs are:

```python
In [26]: dict_nonzeroVFATs[0]
Out[26]: array([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12], dtype=int32)

In [27]: dict_nonzeroVFATs[1]
Out[27]: array([ 0,  1,  2,  3, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23], dtype=int32)
```

And this can be used with:

```python
In [31]: for vfatN in range(0,24):
    if vfatN in dict_nonzeroVFATs[1]:
        print("vfatN {} is nonzero".format(vfatN))
    else:
        print("skipping vfatN {}, put a 'continue' for this case".format(vfatN))
```

Which returns:

```
vfatN 0 is nonzero
vfatN 1 is nonzero
vfatN 2 is nonzero
vfatN 3 is nonzero
skipping vfatN 4, put a 'continue' for this case
skipping vfatN 5, put a 'continue' for this case
skipping vfatN 6, put a 'continue' for this case
skipping vfatN 7, put a 'continue' for this case
skipping vfatN 8, put a 'continue' for this case
skipping vfatN 9, put a 'continue' for this case
skipping vfatN 10, put a 'continue' for this case
skipping vfatN 11, put a 'continue' for this case
vfatN 12 is nonzero
vfatN 13 is nonzero
vfatN 14 is nonzero
vfatN 15 is nonzero
vfatN 16 is nonzero
vfatN 17 is nonzero
vfatN 18 is nonzero
vfatN 19 is nonzero
vfatN 20 is nonzero
vfatN 21 is nonzero
vfatN 22 is nonzero
vfatN 23 is nonzero
```

anaDACScan.py
bdorney:
```python
                print("Skipping OH{0}, detector {1}, missing CFG_CAL_DAC Calibration file:\n\t{2}".format(
```

Still incorrect.  `CFG_CAL_DAC` is not the register calibration info that is needed here.

anaDACScan.py
bdorney: yes I see, I thought a 3x8 plot was made for each distribution

anaDACScan.py
AndrewLevin: @bdorney, I got 

```
Traceback (most recent call last):
  File "./anaDACScan.py", line 93, in <module>
    dict_nonzeroVFATs[ohN] = np.unique(vfatArray[vfatArray[vfatArray['dacValY'] > 0]['link'] == ohN]['vfatN'])
IndexError: boolean index did not match indexed array along dimension 0; dimension is 12288 but corresponding boolean dimension is 7382
```
when I tried to use that. I think you are reducing the size of the array and then evaluating the original array. I used numpy.logical_and to make it work.

anaDACScan.py
AndrewLevin: fixed

anaDACScan.py
bdorney: May be due to different versions of numpy. Anyway you got the point. Thanks for implementing the change.

bdorney (approved):

PR #159, opened 2018-10-17
WIP: Fixed issue w/vfatID only storing 16 bit numbers

bdorney:
## Description

Fixed issue w/vfatID only storing 16 bit numbers.
Due to the vfat bit slip problem you can get a chip ID that has a number that is larger than 16 bits.

### Types of changes

## Motivation and Context

Allow VFAT ID to take up to 32 bit numbers 

## How Has This Been Tested?

Change was trivial.

## Checklist:

### DIFF
--- a/anaUltraScurve.py
+++ b/anaUltraScurve.py
@@ -483,7 +483,7 @@ if __name__ == '__main__':
             myT.Branch( 'trimRange', trimRange, 'trimRange/I' )
         vfatCH = array( 'i', [ 0 ] )
         myT.Branch( 'vfatCH', vfatCH, 'vfatCH/I' )
-        vfatID = array( 'i', [-1] )
+        vfatID = array( 'I', [-1] )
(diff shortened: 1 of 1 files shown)


jsturdy (commented):
Do you understand why this is becoming an issue/not noticed before?
Going from `i` to `I` your guarantee is still the same (minimum of 2 bytes), just you get all 16 of those bits for a value with the `unsigned` type.
I would instead choose a type that has a guaranteed minimum size as large as you are expecting (and that matches what you're telling `ROOT` the size should be

bdorney:
The issue is due to the bit slip problem on the chop ID. And why an additional two bits help I would guess is due to the chance to have it in the first set is more likely.

In any case, I’ll change it to a 32 bit container integer, and then have ROOT write it as a 32 bit integer.  Really was just expecting a number between 0 and 5000.

bdorney:
superseded by https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/160

PR #161, opened 2018-10-22
WIP: Create anaSbitReadout.py

fraivone:
This script unpacks and plots data acquired with sbitReadout.py. The outputs are saved in `inputfolder/sbitReadOut/`
## IMPORTANT
1. the function getMapping in `gempython/gemplotting/utils/anautilities` needs to be modified. At the moment it works with VFAT2 map, in which channel number starts from 1. In order to use the script, I modified getMapping in my virtual environment as suggested here https://mattermost.web.cern.ch/cms-gem-daq/pl/y6r6cqn1rbfa7m38iwn1hb9t5e
5. in the folder `gempython/gemplotting/mapping/` there must be located the files named `shortChannelMap_VFAT3-HV3b-V1_VFAT3-HV3b-V2.txt` `longChannelMap_VFAT3-HV3b-V1_VFAT3-HV3b-V2.txt`, which are the outputs of Louis's pull request https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/134/files#diff-5a657d72aa6d1b867b7000684423a4b1
### USAGE EXAMPLE
```bash
anaSbitReadout -i folder_containing_sbitReadout_outputs/ -t short
```
### FURTHER DEVELOPMENTS
1. Default GEBType is long, so the user may run the script without specifying it  while working with a short GEB. This will cause a bad string <-> channel mapping
2. The path to the .txt file with the channel <-> string mapping could be provided as input; it would be easier to use different mappings
3. An option to mask noisy channel or whole vfat could be useful
### KNOWN ISSUES
* While loading the .dat file, ROOT returns this warning: 
```
Warning in <TTree::ReadStream>: Couldn t read formatted data in "evtNum/i:sbitClusterData0/i:sbitClusterData1/i:sbitClusterData2/i:sbitClusterData3/i:sbitClusterData4/i:sbitClusterData5/i:sbitClusterData6/i:sbitClusterData7/i" for branch evtNum on line 1; ignoring line
``` 
I think it's cause ROOT doesn't expect to find headers in the .dat files opened with the ReadFile function.  At the moment I don't know which is the simplest way to solve this (changing sbitReadout.py? changing the .dat files in this script?).
* the parser is imported from `gempython/gemplotting/utils/anaoptions`, so it contains more options than actually implemented. This is a problem for the consistency of the help menu. At the moment I don't know how to show only the used options in the help menu

### Types of changes

Example of output in case of a Cd109 source placed underneath VFAT7
immagine

## Checklist:


bdorney:
This file won't be built by the package right now since `anaSbitReadout.py` isn't in the list of files in the `Makefile` but `anaSBit*.py` is.  Rename this file via:

```bash
git mv anaSbitReadout.py anaSBitReadout.py
```

this will then be automatically picked up by the `Makefile`.

anaSbitReadout.py
```diff
@@ -0,0 +1,376 @@
+#!/bin/env python
+"""
+    Script to analyze sbitreadout data
+    By: Caterina Aruta (caterina.aruta@cern.ch) and Francesco Ivone (francesco.ivone@cern.ch)
+    """
```
bdorney: This needs to be redone such that it is compatible `sphinx` so that automatic documentation is made.

See examples:

- [gemPlotter.py](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/macros/gemPlotter.py),
- [gemPlotter.rst](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/doc/man/gemPlotter.rst), and
- [man-index.rst](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/doc/man-index.rst)

anaSbitReadout.py
```diff
@@ -0,0 +1,376 @@
...
+if __name__ == '__main__':
+    import ROOT as r
+    import numpy as np
+    import root_numpy as rp
+    import glob
+    import sys
+    import os
+    import time
+    from optparse import OptionParser
```
bdorney: Switch to using `ArgumentParser` from `argparse`.  Then define relevant arguments by hand instead of the blanket import in `anaoptions` (most of which are not suitable here).

anaSbitReadout.py
```diff
@@ -0,0 +1,376 @@
...
+    filename = options.filename+"/sbitReadOut"
+    outfilename = options.outfilename
+    print("Analyzing: '%s'" % path)
+    os.system("mkdir "+filename)
+
+    """
+    At the moment the output of the sbitReadout.py are .dat files with headers that will automatically fill the ROOT TTree. 
+    However ther is one last ':' in the first line that shouldn't be there; consequently the ReadFile function is not able to understand the header. So before reading these .dat files, one has to be sure to remove the ':'
+    To achieve this the following command is needed (removes the last : in the header of all .dat files in the input path)
+    """
```
bdorney: Lines 45-47 please address @jsturdy's comment [here](
https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/208#discussion_r226916707)

anaSbitReadout.py
```diff
@@ -0,0 +1,376 @@
...
+
+    if size not in ('long', 'short'):
+        raise AssertionError("Invalid value of GEBtype")
+    else:
+        pass
+
+    filename = options.filename+"/sbitReadOut"
+    outfilename = options.outfilename
+    print("Analyzing: '%s'" % path)
+    os.system("mkdir "+filename)
```
bdorney: Consider using [runCommand](https://github.com/cms-gem-daq-project/cmsgemos/blob/b44bef4d8cc328b8646474c4cbffced6d9d44197/gempython/utils/wrappers.py#L2) for all instances like this.

Also set permissions to include `g+rw` on all output directories that are made here.

anaSbitReadout.py
```diff
@@ -0,0 +1,376 @@
...
+    r.TH1.SetDefaultSumw2(False)
+    r.gROOT.SetBatch(True)
+    r.gStyle.SetOptStat(1111111)
+
+    mappath = pkg_resources.resource_filename(
+        'gempython.gemplotting', 'mapping/')
+    # Loading the dictionary with the mapping
+    vfat_ch_strips = getMapping(
+        mappath+size+"ChannelMap_VFAT3-HV3b-V1_VFAT3-HV3b-V2.txt", False)
+    print("\nVFAT channels to strips "+size+":\t MAP loaded")
```
bdorney: Typically this should allow for the user to supply a mapping file, if the mapping file argument is `None` then it should default to a mapping file that is present.  Additionally the file you're referencing here *isn't* in the repository so this if installed on a DAQ machine at P5 or 904 will cause a crash.

anaSbitReadout.py
```diff
@@ -0,0 +1,376 @@
...
+
+    # Loading and reversing the dictionary with (eta , phi) <-> vfatN
+    etaphi_to_vfat = ndict()
+    for i in range(1, 9):
+        etaphi_to_vfat[i] = {v: k for k,
+                             v in chamber_iEta2VFATPos[i].iteritems()}
```
bdorney: Seems like you're re-inventing the wheel here.  Why not just use the dictionaries that already exist:

https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/44030d02e019c065e0e9cec140f77740b65c9692/mapping/chamberInfo.py#L68-L87

anaSbitReadout.py
```diff
@@ -0,0 +1,376 @@
+#!/bin/env python
+"""
+    Script to analyze sbitreadout data
+    By: Caterina Aruta (caterina.aruta@cern.ch) and Francesco Ivone (francesco.ivone@cern.ch)
+    """
+if __name__ == '__main__':
+    import ROOT as r
```
bdorney: Please address [this comment](
https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/208#discussion_r226918729) from @jsturdy

anaSbitReadout.py
```diff
@@ -0,0 +1,376 @@
...
+    # Creating the output File and TTree
+    outF = r.TFile(filename+'/'+outfilename, 'recreate')
+    inT = r.TTree('Packed', 'Tree Holding packed raw data')
+
+    # searching for all the files with this format and adding them to the TTree
+    start_time = time.time()
+    print ("\nReading .dat files from the folder '%s'" % path)
+    for idx, file in enumerate(glob.glob(path+'/sbitReadOut_run*.dat')):
+        inT.ReadFile(file)
+        # inT.Fill()
```
bdorney: Please address [this comment](
https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/208#discussion_r226921101) by @jsturdy

anaSbitReadout.py
```diff
@@ -0,0 +1,376 @@
...
+    for i in range(1, 9):
+        etaphi_to_vfat[i] = {v: k for k,
+                             v in chamber_iEta2VFATPos[i].iteritems()}
+
+    """
+        Now it's time to load all the input files and to merge them into one root TTree
+        The TTree file it's going to auto extend each time a new file is found
+        A TFile it's going to hold this TTree
+
+        """
```
bdorney: clean up this spacing.

anaSbitReadout.py
```diff
@@ -0,0 +1,376 @@
...
+        ieta_h_strip[i].SetXTitle("Strip num")
+        ieta_h_ch[i].SetXTitle("Chan num")
+
+    # initializing 2Dhisto
+    hh_ieta_strip = ndict()
+    hh_ieta_ch = ndict()
+    hh_ieta_strip[0] = r.TH2I(
+        'ieta_strip', 'Strips summary        (i#phi = 1,2,3);strip number;i#eta', 384, 0, 384, 8, 0.5, 8.5)
+    hh_ieta_ch[0] = r.TH2I(
+        'ieta_ch', 'Channels summary        (i#phi = 1,2,3);chan number;i#eta', 384, 0, 384, 8, 0.5, 8.5)
```
bdorney: Will sum of square weights behave correctly if these are `TH2I`? e.g. how do the error bars behave? If you have two entries in a bin is the error correctly `1.4...`?

anaSbitReadout.py
```diff
@@ -0,0 +1,376 @@
...
+            The following loop takes care of this possibility
+            """
+            for i in range(2, chHitPerCluster[0]):
+                # filling the adjacent channel
+                vfatCH[i] = vfatCH[i-1] + 1
+                # if the new channel exceeds the total VFAT channels, increase phi and move to the first cannel of the next VFAT
+                if vfatCH[i] >= 128 and phi < 3:
+                    phi = phi + 1
+                    vfatCH[i] = 0
+                    vfatN[0] = etaphi_to_vfat[eta][phi]
```
bdorney: this is redundant; you already know `vfatN[0]` from line 194 (i.e. the for loop at line 236 is inside the loop at 187 so `vfatN[0]` is still the same value.  This also means the dictionary `etaphi_to_vfat` is not needed (since this is the only place it is used).

anaSbitReadout.py
bdorney: Remove lines 63-67, see comment [here](
https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/161/files#r226972715)

anaSbitReadout.py
```diff
@@ -0,0 +1,376 @@
...
+        pass
+
+    if size not in ('long', 'short'):
+        raise AssertionError("Invalid value of GEBtype")
+    else:
+        pass
+
+    filename = options.filename+"/sbitReadOut"
+    outfilename = options.outfilename
+    print("Analyzing: '%s'" % path)
```
bdorney: All print statements should appear like line 41 so that they are `python3` compatible.  You have some print statements that are not `python3` compatible.

anaSbitReadout.py
```diff
@@ -0,0 +1,376 @@
...
+            phi = vfat_to_etaphi[vfatN[0]][1]
+            # SBIT always includes doublet of adjacent channels
+            vfatCH[0] = 2*(sbitAddr % 64)
+            vfatCH[1] = vfatCH[0] + 1
+            strip[0] = vfat_ch_strips[vfatN[0]]['Strip'][vfatCH[0]]
+            strip[1] = vfat_ch_strips[vfatN[0]]['Strip'][vfatCH[1]]
+            # In case of wrong mapping adjacent channels may not be adjacent strips, which is physically inconsistent
+            if(np.abs(strip[0] - strip[1]) > 1):
+                print 'WARNING: not adjacent strips'
+                time.sleep(3)
```
bdorney: Make the `print` statement python3 compatible and remove the `sleep` command.  If you want this to stand out in a log file then colour the message using the [colors class](https://github.com/cms-gem-daq-project/cmsgemos/blob/b44bef4d8cc328b8646474c4cbffced6d9d44197/gempython/utils/gemlogger.py#L4), e.g.:

```python
print("{}{}{}".format(colors.RED,text,colors.ENDC))
```

anaSbitReadout.py
```diff
@@ -0,0 +1,376 @@
...
+                elif vfatCH[i] >= 128 and phi >= 3:
+                    # ERROR
+                    print "ERROR"
+                    print "word", word
+                    print "VFATN:", vfatN[0]
+                    print "VFATCH:", vfatCH[0]
+                    print "Cluster Sz:", sbitSize[0]
+                    print "Eta: ", eta
+                    print "phi: ", phi
+                    break
```
bdorney: This error message is not telling the user why it's an error; moreover how could `vfatCH` and `phi` be respectively greater than 128 and 3?

Suggest to redo this block (`python3` compatible) as a single line with `format(...)` used and a descriptive message to the user.

bdorney (changes requested):
Missing `sphinx` documentation, along with some code changes needed.

anaSbitReadout.py
bdorney: This can be fixed by editing this loop here:

https://github.com/cms-gem-daq-project/xhal/blob/645b5649c0f87ba3d82258082fbf7a4dff5aa853/xhalcore/src/common/rpc_manager/amc.cc#L113-L117

Add an if-else statement that checks if `cluster == 7`, if so then drop the `:` from the printed text, otherwise (`else`) print what is shown.

anaSbitReadout.py
fraivone: Added the option to specify mapping file

anaSbitReadout.py
fraivone: I've adapted the gemPlotter.py to make an auto documentation. However I've left the first 4 lines as they were (my forgetfulness). Let me know if it needs to be changed

anaSbitReadout.py
fraivone: Done

anaSbitReadout.py
fraivone: Fixed as suggested in the comment

anaSbitReadout.py
fraivone: Done

anaSbitReadout.py
fraivone: Yes, I've tested it by reading both entries and error. And it results:
error = sqrt (entries)

anaSbitReadout.py
fraivone: In this loop the **vfatCH** is incremented based on sbitsize. It's possible that, at some point, **vfatCH** exceeds 128, which means that there are hits on the next VFAT. That's why at each loop cycle **vfatN** has been re evaluated

anaSbitReadout.py
fraivone: Done

anaSbitReadout.py
fraivone: Done

anaSbitReadout.py
fraivone: - vfatCH >128 see [here](https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/161#discussion_r234820867)
- phi > 3 it may be due to wrong mapping file or something else

anaSbitReadout.py
bdorney: I'm a bit confused by what you've written here.  You state:

> I've adapted the gemPlotter.py...left the first 4 lines as they were...

But `gemPlotter.py` isn't in the list of changes...nor is it in the scope of this PR.  I don't understand what you mean here.

bdorney:
Why have you modified basically every function in `anautilties.py`?  Additionally many of these changes are needless and unnecessary spacing changes?  You're PR should *only* be modifying those files/functions that are actually related to the proposed feature implementation.

bdorney:
Superceded by https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/185

PR #162, opened 2018-10-27
improvements to anaDACScan.py: replace out-of-range DAC values, improve handling of whitespace/newlines in calibration file

AndrewLevin:
These are two unrelated but simple improvements to the anaDACScan.py script.

## Description
Currently, anaDACScan.py will just warn the user when the fitted nominal DAC values are below 0 or above the maximum range of the register. This PR replaces the out-of-range nominal DAC values with 0 or the maximum value, in addition to printing a warning.

Currently, the handling of the calibration file would fail if a tab or multiple spaces separated the columns. This can be fixed very easily.

### Types of changes

## Motivation and Context
The replacement of out-of-range nominal DAC values was requested by @bdorney in a gemctp7user PR https://github.com/cms-gem-daq-project/gemctp7user/pull/37#pullrequestreview-168729078. 

The issue with the calibration file handling was obvious.

## How Has This Been Tested?
I have run the updated script successfully on DAC scan data on gem904daq01.

## Checklist:

### DIFF
--- a/anaDACScan.py
+++ b/anaDACScan.py
@@ -288,14 +288,10 @@ if __name__ == '__main__':
             #evaluate the fitted function at the nominal current or voltage value and convert to an integer
             nominalDacValue = int(dict_DACvsADC_Funcs[oh][vfat].Eval(nominal))
             
-            if nominalDacValue > maxDacValue:
-                print('Warning: Fitted DAC value > the maximum value of the register. It will be replaced by the maximum value of the register.')
-                nominalDacValue = maxDacValue
+            if nominalDacValue < 0 or nominalDacValue > maxDacValue:
+                print('Warning: The fitted DAC value, '+str(nominalDacValue)+', is outside of the range that the register can hold: [0,'+str(maxDacValue)+']. It will be replaced by '+str(max(0,min(maxDacValue,nominalDacValue)))+'.')
+                nominalDacValue = max(0,min(maxDacValue,nominalDacValue))
 
-            if nominalDacValue < 0:
-                print('Warning: Fitted DAC value < 0. It will be replaced by 0.')
-                nominalDacValue = 0
-            
             dict_dacVals[oh][vfat] = nominalDacValue
             graph_dacVals[oh].SetPoint(graph_dacVals[oh].GetN(),vfat,dict_dacVals[oh][vfat])
              
--- a/anaDACScan.py
+++ b/anaDACScan.py
@@ -156,9 +156,9 @@ if __name__ == '__main__':
         for line in open(args.calFileList):
             if line[0] == "#":
                 continue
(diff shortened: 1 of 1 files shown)


jsturdy (commented):
Possibly cleaner implementation would be to simply take:
```python
max(0,min(maxDacValue,nominalDacValue))
```
or, in the case that `0` is *not* the minimum allowed DAC value
```python
max(minDacValue,min(maxDacValue,nominalDacValue))
```

AndrewLevin:
changed

anaDACScan.py
```diff
@@ -288,13 +288,11 @@
             #evaluate the fitted function at the nominal current or voltage value and convert to an integer
             nominalDacValue = int(dict_DACvsADC_Funcs[oh][vfat].Eval(nominal))
             
-            if nominalDacValue > maxDacValue:
-                print('Warning: the nominal DAC value that was found from the fit is larger than the maximum value that the DAC register will hold')
+            if nominalDacValue < 0 or nominalDacValue > maxDacValue:
+                print('Warning: The fitted DAC value, '+str(nominalDacValue)+', is outside of the range that the register can hold: [0,'+str(maxDacValue)+']. It will be replaced by '+str(max(0,min(maxDacValue,nominalDacValue)))+'.')
+                nominalDacValue = max(0,min(maxDacValue,nominalDacValue))
```
jsturdy: If you reverse L293 and L292, and use string formatting (rather than concatenation), the long term maintainability will improve

AndrewLevin: ok, changed

@bdorney, can you confirm that 0 is always the minimum DAC value?

bdorney: For the DAC's that are considered the minimum is 0, see:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/4d82a717400fd96cb6f18104b58b93a1e8751036/src/calibration_routines.cpp#L1372-L1401

jsturdy (approved):
OK by me

anaDACScan.py
```diff
@@ -156,9 +156,9 @@
         for line in open(args.calFileList):
             if line[0] == "#":
                 continue
-            line = line.strip(' ').strip('\n')
-            first = line.split(' ')[0].strip(' ')
-            second = line.split(' ')[1].strip(' ')
+            line = line.strip()
+            first = line.split()[0]
+            second = line.split()[1]
```
bdorney, line 161: While it won't matter for this application you should get in the habit of not calling subsequent actions like this.  If `line` is a large string you'll be spending time parsing it twice.  It would be better to parse it once:

```python
line = line.strip().split()
first = line[0]
second = line[1]
```

Fine for now; but pay closer attention to this in the future.

bdorney (approved):

PR #178, opened 2019-01-31
Bug Fix: Resolved a type mismatch error

bdorney:
## Description

Changed `mapVFATPos2VFATSN` from `int32` to `uint32` data type.  This resolves a type mismatch in older versions of `numpy` and is also more correct.  The `vfatID` is treated as an unsigned 32 bit integer due to the possibility of bit flips in the chipID register. 

### Types of changes

## Motivation and Context

Resolves type issue shown [here](https://mattermost.web.cern.ch/cms-gem-daq/pl/8sy34q8cn78zbm4fp6gcuqxdge).

## How Has This Been Tested?

Ran analysis without having the problem in older version of `numpy`.

## Checklist:


bdorney:
I think I'm just gonna make VFAT serial number a float until reed muller comes along.

Issue #180, opened 2019-02-05
Discussion: Expanding chamberInfo.py to multi-AMC's/crates (Short Term)


bdorney:
### Types of issue

- [ ] Bug report (report an issue with the code)
- [ ] Feature request (request for change which adds functionality)
- [x] Discussion

## Some Proposals

I need a short term hack to be able to handle multiple AMC's in the same crate, and possibly multiple crates by the same DAQ machine.  Right now the following dictionaries in [chamberInfo.py](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/mapping/chamberInfo.py) are index by `ohN`:

- `chamber_config`,
- `GEBtype`,
- `chamber_vfatMask`, and
- `chamber_vfatDACSettings`

With the advantage of being human readable but has some drawback that it requires the link number to be common across all dictionaries (room for sysadmin to make a mistake).

Couple ideas come to mind to expand this to a multi AMC/crate environment:

### Idea 1: Dictionary Index Is A Mathematical Expression 

All dictionaries are kept but we extend index by following:

```
idx = (crate -1) * 144 + (slot - 1) * 12 + ohN
```

 - Pros: probably requires minimal refactoring for all dependant scripts/functions in `gemplotting` and `vfatqc`
 - Cons: probably the hardest to maintain...however I expect this to be a "short term solution" and be replaced by a test stand's DB or the production DB (@ P5).

This would also retire the `cardName` option and probable force the `(shelf,slot,ohN)` ordered triplet.  This is an advantage as it would match cms address table expectations. 

### Idea 2: Dictionary Keys Are Now Tuples 

All dictionaries are kept but their keys are transformed into tuples of`(shelf,slot,ohN)` to access value of interest.

   - Pros: Still minimal refactoring, no complicated index rule to worry about
   - Cons: Some refactoring of `TTree`s in `treeStructures.py` to contain an `AMC` branch (backwards compatibility for analyzing old data, but this is solvable), still links a set of common indices across multiple dictionaries.  Again expected to be a "short term" solution.

This would also retire the `cardName` option and probable force the (`shelf`,`slot`,`ohN`) ordered triplet.  This is an advantage as it would match cms address table expectations. 

### Idea 3: Compress Dictionaries into Subset of Nested Dictionaries

The following dictionaries:

- `chamber_config`,
- `GEBtype`, and
- `chamber_vfatMask`.

[...]

PR #189, opened 2019-03-07
Addresses #180, Implementing Idea #2

bdorney:
## Description

Addresses #180 by implementing idea number 2.

I've completely removed the `chamber_vfatMask` dictionary.  This was just one more thing that had to be edited and tracked by hand.  Software tools exist now to get the VFAT mask on the fly so there's no point in tracking it in a config location.  Removes one more potential spot for error.

### Types of changes

## Motivation and Context

Need to be able to handle multiple uTCA crates and AMC's.

## How Has This Been Tested?

Took a new DAC scan and then analyzed data successfully.

## Checklist:

### DIFF
--- a/anaDACScan.py
+++ b/anaDACScan.py
@@ -29,10 +29,10 @@ Arguments
 
 .. option:: --calFileList
   
-    Provide a file containing a list of calFiles.  The space character is used as a delimiter.  The first column is the link number.  The second is the calibration file (given with full physical filename) for that link number. Example::
+    Provide a file containing a list of calFiles.  The space character is used as a delimiter.  The first three columns are respectively shelf, slot and link number.  The fourth column is the calibration file (given with full physical filename) for that link number. Example::
 
-    0 /path/to/my/cal/file.txt
-    1 /path/to/my/cal/file.txt
+    1 4 0 /path/to/my/cal/file.txt
+    1 4 1 /path/to/my/cal/file.txt
     ...
     ...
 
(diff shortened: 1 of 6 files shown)


jsturdy (commented):
As long as you're completely reworking the original `chamberInfo` hacks, may as well simply have the loading modules pull them from a location specified by an environment variable, rather than some fixed location within the package (would play into your QoL style requests)

bdorney:
> As long as you're completely reworking the original chamberInfo hacks, may as well simply have the loading modules pull them from a location specified by an environment variable, rather than some fixed location within the package (would play into your QoL style requests)

I'm not sure I understand what you mean by "loading modules" do you mean that we should change `chamberInfo.py` to load some text file that then fills the dicts and that it should look in some `$ENV` variable?

jsturdy:
That's probably less of a change than finding every place that does ("loading modules"):
```python
from chamberInfo import thing
```
and having it look in some variable dependent location.

What would be best was if it tries to load from the variable location, if it fails, then continues using some default that is provided in the package, printing a message about using defaults (or it could fail, if that is the desired behaviour)

Something like the following (here `user_constants.py` is in some directory that has been added to the `PYTHONPATH` variable); the commented lines may be uncommented if one prefers to manipulate the system path on the fly (not the best practice in general), and in that case the environment variable `CONSTANTS_PATH` should point to the location of `user_constants.py`):
```python
# import sys,os

try:
    # sys.path.append(os.getenv("CONSTANTS_PATH"))
    from user_constants import GBT
except Exception as e:
    print(e,"Using package default")
    GBT = {
        "VAR1": 1,
        "VAR2": 2,
        "VAR3": 3,
    }
```

The contents of `user_constants.py` in this example is:
```python
GBT = {
    "VAR1": 5,
    "VAR2": 4,
    "VAR3": 7,
}
```

bdorney:
okay; I would prefer to put this in a different PR.

Can you make a new issue and reference your comment above?

jsturdy:
95% of the work is already done (and the other 5% is just copy/paste/modify from my previous comment), **and** you've now added several setup specific configurations, which is the whole point of having a user specifiable location.
Better to just "get it done right"...

bdorney:
@jsturdy was the most recent commit what you intended?

jsturdy (approved):
I guess it's close enough (would have preferred if *all* specific information was replaced with simply generic placeholders)

mexanick (approved):

PR #191, opened 2019-03-11
Maintenance: Remove setup directory and associated instructions

bdorney:
## Description

Removes the `setup` directory and associated scripts.  These scripts are no longer supported and the package should be used either in a `virtualenv` or as a system installed `rpm`.

### Types of changes

- [x] Maintenance

## Motivation and Context

We expect a new set of users and developers will begin with the package; setup instructions should reflect the actual setup rather than outdated instructions.

## Checklist:


bdorney:
@BenjaminRS I hope it will be clearer now.

mexanick (approved):

Issue #190, opened 2019-03-11
Change Dead Channel Cut Defaults to v3 Compatible Values


bdorney:
## Brief summary of issue

Right now the default dead channel cuts, which identify a channel as dead, are for the VFAT2 ASIC.  We need to change these defaults to be for VFAT3 (case of no cooling circuit present), using medium VFAT3 gain and comparator in CFD mode (see [General Overview of VFAT3](https://github.com/cms-gem-daq-project/sw_utils/blob/develop/v3ElectronicsUserGuide.md#general-overview-of-vfat3)).

### Types of issue

- [x] Maintenance

## Expected Behavior

The defaults shown below:

https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/78c88741fbe292f37476c79ffad108a137c1aa27/anaUltraScurve.py#L178-L183

Should be taken to be for VFAT3 (case of no cooling circuit attached, medium VFAT3 gain and comparator in CFD mode) rather than for VFAT2.

## Current Behavior

Defaults reflect the VFAT2 case.

## Context (for feature requests)

Defaults reflect VFAT2 case and is not the same charge for VFAT3.

## Your Environment

* Version used: 78c88741fbe292f37476c79ffad108a137c1aa27

bdorney:
Closed by #192

PR #192, opened 2019-03-11
Update anaUltraScurve.py to no longer use vfat2 dead chan cut as default; Addresses Issue 190

fsimone91:
Set the isVFAT3 option as default option
Updated the deadChanCutHigh and deadChanCutLow for the v3 electronics

Addresses #190 

## Description

Default option now is the isVFAT3 option.
The default values for deadChanCutHigh and deadChanCutLow are different in the isVFAT3 and isVFAT2 cases, so:
- the default argument for the --deadChanCutLow and --deadChanCutHigh options have been removed
- if the user did not set --deadChanCutLow or --deadChanCutHigh, the two variables are `None`, in this case the default values (for v2 ad v3 electronics) are set

Sigma values for v3 electronics:
[plot of sigma at zero added capacitance for each VFAT position, at ARM_DAC 100 and Medium Gain](https://mattermost.web.cern.ch/files/f6p11q45mtnypx6f47bfneo7bh/public?h=gfy2loTeP4klfJlDXXos5LNOStrdAC-uo9ot1PQgQ5U)
[summary plot for all VFATs](https://mattermost.web.cern.ch/files/mdxt1gpzx7dfj8m38sicod4kdy/public?h=55D_o6gK29-uh1wxShCbS1qgNknxj1vAHczGHLVUVSQ)

### Types of changes

## Motivation and Context

Open Issue: https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/190

## Checklist:


anaUltraScurve.py
```diff
@@ -304,8 +320,7 @@ def plotAllSCurvesOnCanvas(vfatHistos, vfatHistosPanPin2=None, obsName="scurves"
     if options.performFit:
         fitter = ScanDataFitter(
                 calDAC2Q_m=calDAC2Q_Slope, 
-                calDAC2Q_b=calDAC2Q_Intercept,
-                isVFAT3=options.isVFAT3
+                calDAC2Q_b=calDAC2Q_Intercept
```
bdorney: ah we should keep link 308 but change it to:

```python
isVFAT3=isVFAT3
```

Otherwise the `ScanDataFitter` class instance will not behave correctly.

bdorney (changes requested):
ScanDataFitter should still be given the `isVFAT3` parameter.

bdorney:
Also you'll need to update your local develop branch.  

You'll need to do something like:

```bash
git fetch -p
git pull cms-gem-daq-project develop
```

Where `cms-gem-daq-project` is the name of the central repo that you have in your `.git/config` file; if you cloned directly from the central repository this will be called `origin`.

In the future it's a good idea to keep your local `develop` branch sync'd with the remote (cms-gem-daq-project) `develop` branch and then branch off of it when you want to change the code/create a new feature.  For example:

```bash
git fetch -p
git checkout develop
git pull cms-gem-daq-project #this makes sure you're even with the remote repository
git checkout -b myCoolNewFeatureBranch
# make your code changes and commit them
git push fsimone91 myCoolNewFeatureBranch
```

anaUltraScurve.py
```diff
@@ -175,17 +175,33 @@ def plotAllSCurvesOnCanvas(vfatHistos, vfatHistosPanPin2=None, obsName="scurves"
...
 
     parser.set_defaults(outfilename="SCurveFitData.root")
     (options, args) = parser.parse_args()
     
+    #Default isVFAT2=false, isVFAT3=true
+    isVFAT3 = (not options.isVFAT2)
+    
+    #if isVFAT2, deadChanCut are set to default for v2 electronics
+    if options.isVFAT2:
+        if options.deadChanCutLow is None:
```
bdorney, line 194: Just a thought, will this default to none if with `type="float"`? Or will it default to `0.0`?  I would suggest for each of these two parameters to have `default=None` for these two parameters at lines 178 & 181 respectively.

bdorney:
You can try to test your changes by building the package, installing it your python virtual environment and re-analyzing old data.

Build & installation instructions are available [here](https://github.com/cms-gem-daq-project/sw_utils/blob/develop/v3ElectronicsUserGuide.md#gem-plotting-tools).  Take note of the info about [setting up the gembuild repo](https://github.com/cms-gem-daq-project/sw_utils/blob/develop/v3ElectronicsUserGuide.md#build-prerequisites-the-gembuild-repo).

anaUltraScurve.py
fsimone91: Reading some examples from https://docs.python.org/3/library/optparse.html looks like when the option is not set, options.name is `none` independently on the type. Nevertheless, this is not explicitly stated in the text so I will follow your suggestion.

bdorney: Thanks for checking this! Good to know. Keep in mind on cc7 python version defaults to 2.7; I doubt there’s a difference but sometimes functionality changes. 

In either case I think we’ve resolved this nicely.

bdorney (approved):

mexanick (approved):

PR #193, opened 2019-03-27
[Bug Fix]: in DAC scan analysis, link should be compared with a single number

AndrewLevin:
The `link` branch should be compared with a single number instead of a triplet.

## Description
`ohKey` is a triplet while `vfatArray['link']` is a single integer, so a comparison of them will always be false. In this pull request, `ohKey` is replaced by `entry['link']` in the comparison.

### Types of changes

## Motivation and Context
This is causing the DAC scan analysis to produce empty results.

## How Has This Been Tested?
Yes, it has been tested on the coffin setup.

## Checklist:

### DIFF
--- a/utils/anautilities.py
+++ b/utils/anautilities.py
@@ -55,7 +55,10 @@ def dacAnalysis(args, dacScanTree, chamber_config, scandate='noscandate'):
     dict_nonzeroVFATs = {}
     for entry in crateMap:
         ohKey = (entry['shelf'],entry['slot'],entry['link'])
-        dict_nonzeroVFATs[ohKey] = np.unique(vfatArray[np.logical_and(vfatArray['dacValY'] > 0,vfatArray['link'] == ohKey)]['vfatN'])
+        arrayMask = np.logical_and(vfatArray['dacValY'] > 0, vfatArray['link'] == entry['link'])
+        arrayMask = np.logical_and(arrayMask, vfatArray['slot'] == entry['slot'])
+        arrayMask = np.logical_and(arrayMask, vfatArray['shelf'] == entry['shelf'])
+        dict_nonzeroVFATs[ohKey] = np.unique(vfatArray[arrayMask]['vfatN'])
 
     from gempython.utils.wrappers import envCheck
     envCheck("DATA_PATH")


bdorney:
You haven’t setup your chamberInfo.py correctly, see example:

https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/mapping/chamberInfo.py

A tuple key value is expected.

AndrewLevin:
Sorry, I don't understand. `vfatArray['link']` is a branch in `dacScanTree`

```
vfatArray = rp.tree2array(tree=dacScanTree,branches=list_bNames)
```

It should be a single number, and so it should not be compared with a triplet.

bdorney:
Yes I see.  My mistake.

bdorney (approved):

utils/anautilities.py
```diff
@@ -55,7 +55,7 @@ def dacAnalysis(args, dacScanTree, chamber_config, scandate='noscandate'):
     dict_nonzeroVFATs = {}
     for entry in crateMap:
         ohKey = (entry['shelf'],entry['slot'],entry['link'])
-        dict_nonzeroVFATs[ohKey] = np.unique(vfatArray[np.logical_and(vfatArray['dacValY'] > 0,vfatArray['link'] == ohKey)]['vfatN'])
+        dict_nonzeroVFATs[ohKey] = np.unique(vfatArray[np.logical_and(vfatArray['dacValY'] > 0,vfatArray['link'] == entry['link'])]['vfatN'])
```
bdorney: this should then probably include a check on shelf and slot as well to ensure the correct array is paired with the correct `ohKey`.

bdorney (changes requested):
ensure unambiguous comparison when deriving `dict_nonzeroVFATs[ohKey]`

utils/anautilities.py
bdorney: ~~Probably the following would work:~~

It doesn't see below.

bdorney: The actual solution looks like:

```python
    # get nonzero VFATs
    dict_nonzeroVFATs = {}
    for entry in crateMap:
        ohKey = (entry['shelf'],entry['slot'],entry['link'])
        arrayMask = np.logical_and(vfatArray['rate'] > 0, vfatArray['link'] == entry['link'])                                                                                                                                    
        arrayMask = np.logical_and(arrayMask, vfatArray['slot'] == entry['slot'])
        arrayMask = np.logical_and(arrayMask, vfatArray['shelf'] == entry['shelf'])
        dict_nonzeroVFATs[ohKey] = np.unique(vfatArray[arrayMask]['vfatN'])
```

`np.logical_and(...)` doesn't seem smart enough to take multiple arrays.  One could do it in two lines by taking the logical and of two logical ands but that I feel would degrade readability.

bdorney: @AndrewLevin ?

AndrewLevin: I just pushed your solution, but haven't tested it (`testConnectivity.py` does not seem to save the DAC scan data, so I need to run another one).

bdorney:
Please squash your git commit history as well.

AndrewLevin:
squashed into one commit

bdorney (commented):
Typo from my solution (applied to another branch)

E.g. “rate” doesn’t exist, should be dacValY

AndrewLevin:
fixed

jsturdy:
@bdorney is this the PR you wanted reviewed?
It is still blocked by your previous change request

jsturdy (approved):
cursory

PR #199, opened 2019-04-05
updating macro for dacScan comparison

adeiorio:
### Types of changes

## Checklist:

### DIFF
--- a/dacScanComparison2.py
+++ b/dacScanComparison2.py
@@ -1,0 +1,77 @@
+#
+#
+# This macro takes the name of the chamber and all the scandates from the scandates list file and produces tables for the twelve parameter of the VFAT electronics
+#
+#
+
+def getcolumn(filename, column, vfat):
+    f=open(filename,"r")
+    lines=f.readlines()
+    values=[]
+    if(vfat==-1):
+        for line in lines:
+            values.append(line.strip().split(' ')[column])
+    else:
+        values = "N.A."
+        for line in lines:
+            if(line.strip().split('\t')[0]==vfat):
+                values = line.strip().split('\t')[column]
+    f.close()
+    #print values
+    return values
+
+import argparse
+import os,commands
+import sys
+usage = 'python dacScanComparison.py python dacScanComparison.py /path/to/DACscans/ --sdlist scandatelist.txt --vfat all -o /path/for/the/results/'
+parser = argparse.ArgumentParser(usage)
+parser.add_argument('filepath', type=str, help="File path from which input information is read until the folder with the chamber name", metavar='filepath')
+parser.add_argument('--vfat', dest='vfat', type=str, default = 'none', help='')
+parser.add_argument('--sdlist', dest='sdlist', type=str, default = '', help='Names of the scandate file to analyse')
+parser.add_argument('-o','--outpath', dest='outpath', type=str, default = './', help='Path on the folder where the summary table will be written')
+parser.add_argument('--popup', dest='popup', default=False, action='store_true') 
+args = parser.parse_args()
+
+print("Analyzing: '%s'" %args.filepath)
+print("scandatelistfile: '%s'" %args.sdlist)
+
+chamber = getcolumn(args.sdlist,0,-1)[0]
+scandates = getcolumn(args.sdlist,1,-1)
+vfats = []
+
+if args.vfat=="none":
+    print("Please pass the name of the vfat(s) you want to analyse with the option --vfat")
+    sys.exit()
+elif args.vfat=="all":
+    for x in range(0,24):
+        vfats.append(str(x))
+else:
+    for vf in (args.vfat).split(","):
+        vfats.append(vf)
+#print vfats
+
+parameters = ["BIAS_CFD_DAC_1", "BIAS_CFD_DAC_2", "BIAS_PRE_I_BIT", "BIAS_PRE_I_BLCC", "BIAS_PRE_I_BSF", "BIAS_PRE_VREF", "BIAS_SD_I_BDIFF", "BIAS_SD_I_BFCAS", "BIAS_SD_I_BSF", "BIAS_SH_I_BDIFF", "BIAS_SH_I_BFCAS", "HYST"]
+
+for vfat in vfats:
+    table = []
+    table.append(parameters)
+    outpath = args.outpath+"/"+str(chamber)
+    if not(os.path.exists(outpath)):
+        os.makedirs(outpath)
+    outfile = outpath+"/SummaryTable_VFAT"+str(vfat)+".txt"
+    of = open(outfile, "w")
+    #print str(scandate)
+    for scandate in scandates:
+        par_path = args.filepath+"/"+str(chamber)+"/dacScans/"+str(scandate)+"/"
+        scanparam = []
+        for par in parameters:
+            scanparam.append(getcolumn(par_path+"NominalValues-CFG_"+par+".txt", 1, vfat))
+        table.append(scanparam)
+    of.write("VFAT "+vfat+"\n")
+    if(args.popup):
+        print("DAC_PARAM\t"+"\t".join([str(scandate) for scandate in scandates]))
+    of.write("DAC_PARAM\t"+"\t".join([str(scandate) for scandate in scandates])+"\n")
+    if(args.popup):
+        print('\n'.join(' '.join(map(str,sl)) for sl in zip(*table)))
+    of.write('\n'.join('\t'.join(map(str,sl)) for sl in zip(*table)))
+    of.close()


bdorney:
@adeiorio one thing you'll need to do here would be to fill out the template form when submitting a PR.  You need to fill out the text in each area and not leave it blank.

Additionally you should not make use excessively of `subprocess.call` (which is what you're doing by not encapsulating things inside a `main` block/function definitions/etc...).  Try to clean this up a bit.

I think a better location for this to go would be the `macros/` subdirectory.  For this you'll need to do a `git mv`, e.g.:

```bash
git mv  dacScanComparison2.py macros/ dacScanComp.py
```

This also would need documentation written in the [sphinx](http://www.sphinx-doc.org/en/master/) style.  See for example [lines 3-53](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/4f264b10acddcf846fa87faf9f4d87bb45ce9c77/anaDACScan.py#L3-L53) of `anaDACScan.py`.  For this you'll also need to edit: 

- [doc/man-index.rst](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/doc/man-index.rst), I would place it under misc.

And you'll need to create a [man page](https://github.com/cms-gem-daq-project/gem-plotting-tools/tree/develop/doc/man) for it, see for example:

- [doc/man/anaDACScan.rst](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/doc/man/anaDACScan.rst)

These are needed to ensure that the doc will build automatically.

Finally, this would need a rebase as it's not up-to-date with the `develop` branch.

bdorney (changes requested):
please address changes [here](https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/199#issuecomment-480209843)

dacScanComparison2.py
```diff
@@ -0,0 +1,77 @@
...
+parser.add_argument('--sdlist', dest='sdlist', type=str, default = '', help='Names of the scandate file to analyse')
+parser.add_argument('-o','--outpath', dest='outpath', type=str, default = './', help='Path on the folder where the summary table will be written')
+parser.add_argument('--popup', dest='popup', default=False, action='store_true') 
+args = parser.parse_args()
+
+print("Analyzing: '%s'" %args.filepath)
+print("scandatelistfile: '%s'" %args.sdlist)
+
+chamber = getcolumn(args.sdlist,0,-1)[0]
+scandates = getcolumn(args.sdlist,1,-1)
```
bdorney, line 39: It seems like your parsing some `scandelist.txt` file; I'm not sure how you expect this file to be formatted but instead of inventing your own parsing tool why not use what's in the API?  Specifically [parseListOfScanDatesFile](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/4f264b10acddcf846fa87faf9f4d87bb45ce9c77/utils/anautilities.py#L878)

dacScanComparison2.py
```diff
@@ -0,0 +1,77 @@
...
+                values = line.strip().split('\t')[column]
+    f.close()
+    #print values
+    return values
+
+import argparse
+import os,commands
+import sys
+usage = 'python dacScanComparison.py python dacScanComparison.py /path/to/DACscans/ --sdlist scandatelist.txt --vfat all -o /path/for/the/results/'
+parser = argparse.ArgumentParser(usage)
```
bdorney, line 27: This should not be a usage string, and since you have not defined a keyword this is interpreted as the `prog` argument of the `ArgumentParser` constructor, see:

https://docs.python.org/2/library/argparse.html#argparse.ArgumentParser

Additionally it's not necessary to provide a usage string since this will automatically be determined by `ArgumentParser` based on the arguments you define with `add_argument` or other class methods.

Typically this should be a `description`.

dacScanComparison2.py
```diff
@@ -0,0 +1,77 @@
...
+        values = "N.A."
+        for line in lines:
+            if(line.strip().split('\t')[0]==vfat):
+                values = line.strip().split('\t')[column]
+    f.close()
+    #print values
+    return values
+
+import argparse
+import os,commands
```
bdorney, line 24: why import something you are not using? e.g. `commands`?

dacScanComparison2.py
```diff
@@ -0,0 +1,77 @@
...
+print("Analyzing: '%s'" %args.filepath)
+print("scandatelistfile: '%s'" %args.sdlist)
+
+chamber = getcolumn(args.sdlist,0,-1)[0]
+scandates = getcolumn(args.sdlist,1,-1)
+vfats = []
+
+if args.vfat=="none":
+    print("Please pass the name of the vfat(s) you want to analyse with the option --vfat")
+    sys.exit()
```
bdorney, line 44: A better case here would be to exit with an error code (e.g. `os.EX_USAGE` see https://docs.python.org/2/library/os.html#process-management) or to raise an exception like `IOError` https://docs.python.org/2/library/exceptions.html.

dacScanComparison2.py
```diff
@@ -0,0 +1,77 @@
...
+elif args.vfat=="all":
+    for x in range(0,24):
+        vfats.append(str(x))
+else:
+    for vf in (args.vfat).split(","):
+        vfats.append(vf)
+#print vfats
```
bdorney, line 51: I would have this do all vfats by default; there's probably no case where we would be interested in just *one* VFAT.  The DAC scan application doesn't even allow for a single VFAT scan.

dacScanComparison2.py
```diff
@@ -0,0 +1,77 @@
...
+
+parameters = ["BIAS_CFD_DAC_1", "BIAS_CFD_DAC_2", "BIAS_PRE_I_BIT", "BIAS_PRE_I_BLCC", "BIAS_PRE_I_BSF", "BIAS_PRE_VREF", "BIAS_SD_I_BDIFF", "BIAS_SD_I_BFCAS", "BIAS_SD_I_BSF", "BIAS_SH_I_BDIFF", "BIAS_SH_I_BFCAS", "HYST"]
```
bdorney, line 53: Use what's in the API rather than defining an additional hard coded list, e.g. the keys of `nominalDACValues`:

https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/4f264b10acddcf846fa87faf9f4d87bb45ce9c77/utils/anaInfo.py#L17-L34

dacScanComparison2.py
```diff
@@ -0,0 +1,77 @@
...
+
+for vfat in vfats:
+    table = []
+    table.append(parameters)
```
bdorney, line 57: It seems like you're trying to make a table...why not use the `tabulate` package?  This is already being used in the repo.

dacScanComparison2.py
```diff
@@ -0,0 +1,77 @@
...
+    outpath = args.outpath+"/"+str(chamber)
+    if not(os.path.exists(outpath)):
+        os.makedirs(outpath)
+    outfile = outpath+"/SummaryTable_VFAT"+str(vfat)+".txt"
```
bdorney, line 61: I think a better thing would be to by default place this in `$ELOG_PATH` which is a `$ENV` variable on all 904 and P5 machines:

You can do something like:

```python
from gempython.utils.wrappers import envCheck, runCommand
envCheck("ELOG_PATH")
outPath = "{0}/{1}".format(os.getenv("ELOG_PATH"),chamberName)

runCommand( ["mkdir","-p",outPath] )
```

dacScanComparison2.py
adeiorio: I thought the macro to perform a quick check on a single VFAT that shows problems during the different scandates. It is flexible to perform in one time also all the VFATs, by creating different files for each VFAT.

dacScanComparison2.py
adeiorio: It refers to a previous version of the code. Removed!

dacScanComparison2.py
adeiorio: In some scandates the last 4 DacValues are not present. By importing all of them will give an error. Are all these values present in future?

dacScanComparison2.py
adeiorio: Where is this package? For sure I will use it, it can simplify a lot the code

dacScanComparison2.py
bdorney: Yes I understand; but I'm asking to expand the scope to something with a broad scope of validity.

dacScanComparison2.py
bdorney: Importing the dictionary will not cause an error; how you use the dictionary is the key.  Just don't use an entry in the dictionary that is not in the outputs.

dacScanComparison2.py
bdorney: https://pypi.org/project/tabulate/

This is what I found after a quick google search.  There might be a better reference guide of the tabulate API elsewhere.

dacScanComparison2.py
adeiorio: I will set the default to all and leave also the possibility to check a single VFAT. Is it fine?

dacScanComparison2.py
adeiorio: OK

dacScanComparison2.py
bdorney: Sounds great! Thanks

dacScanComparison2.py
adeiorio: Done

bdorney:
@adeiorio you'll need to rebase this branch since it's out of date with `develop`.

If you are unsure how to do that I'd be happy to show you.

adeiorio:
> 
> 
> @adeiorio you'll need to rebase this branch since it's out of date with `develop`.
> 
> If you are unsure how to do that I'd be happy to show you.

@bdorney I'd like to have more detailed instruction on this just to be sure to run the right command.

mexanick:
any progress on this? @bdorney @adeiorio

mexanick:
Since there was no activity in this PR for a long time, it is considered as stale and subjected to be closed.

PR #206, opened 2019-05-21
in calibrateThrDac.py, remove points with scurveMeanError/scurveMean < 0.001

AndrewLevin:
Add protection against failed Gaussian fits to the scurveMean distribution.

## Description
The scurve mean and the error on the scurve mean are determined by a Gaussian fit to a histogram (actually a TGraph) of the scurveMean. The Gaussian fit can fail in a way that produces a reasonable scurveMean but an scurveMeanError that is unreasonably small. This can cause the scurveMean vs THR_DAC fit to fail. So, this pull request removes points in which the scurveMeanError is less than 0.1% of the scurveMean error.

### Types of changes

## Motivation and Context
In one case, VFAT18 below, the highest THR_DAC point gets assigned an scurve mean error of 6*10^-7, which biases the THR_DAC vs scurveMean fit to such an extent that it fails.

canvScurveMeanVsThrDac_Summary 11 old

That point is removed with this pull request, as shown below.

canvScurveMeanVsThrDac_Summary 11:

## How Has This Been Tested?
Yes. Among the 33 list of scandates files that I tested, there are 7 points, shown below, that are removed by this algorithm (the Gaussian fit fails in a way such that the Gaussian is not even rendered in the plot). I think these plots indicate that there may be a related issue with the binning of the histogram in the scurve analysis, but I think that should be fixed in the scurve analysis script, and this script should not try to recover from it.

GE11-X-L-CERN-0002_2019 03 04 21 45_4_25 0
GE11-X-L-GHENT-0002_2019 01 16 18 35_11_30 0
GE11-X-L-GHENT-0002_2019 01 16 18 35_23_30 0
GE11-X-S-BARI-0012_2018 12 05 22 06_23_40 0
GE11-X-S-BARI-0012_2018 12 05 22 38_3_60 0
GE11-X-S-CERN-0012_2018 08 10 18 47_18_150 0
GEB3c-l1-w_2018 11 05 12 04_21_150 0

## Checklist:


bdorney:
Recall the instructions from @jsturdy in the SW dev meeting regarding our workflow.  We should first open an issue on github, explain the planned changes, and a PR should be made only after the community discusses and agrees on the planned changes.

Could you please open an issue here and let's open the discussion in that way.

AndrewLevin:
here is the issue: https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/208

PR #211, opened 2019-05-27
[Bugfix] forgot to import os in armDac calibration parsing function

AndrewLevin:
`import os` statement was missing

## Description
One line change.

### Types of changes

## Motivation and Context
Would crash if it ever reaches this line, although it seems PyROOT is not actually throwing IOErrors in the current version. Instead it prints:

```
Error in <TTree::ReadFile>: Cannot open file: ~/test.txt
```

## How Has This Been Tested?
Change is trivial.

## Checklist:

### DIFF
--- a/utils/anautilities.py
+++ b/utils/anautilities.py
@@ -1136,6 +1136,7 @@ def parseArmDacCalFile(filename):
         calTree.ReadFile(filename)
     except IOError:
         printRed("file {0} is not readable or does not exist, please cross-check".format(filename))
+        import os
         exit(os.EX_IOERR)
     array_CalData = rp.tree2array(tree=calTree, branches=list_bNames)
     


mexanick:
can you please also check other parts of the code. I believe the `import os` is missing in some other places too

jsturdy (approved):

AndrewLevin:
> can you please also check other parts of the code. I believe the `import os` is missing in some other places too

In `anautilities.py`, I looked at all of the `os.*` lines, and am pretty sure this is the only place where the import is missing. I also looked around for other files in `gem-plotting-tools` where it might be missing but didn't find any.

mexanick (approved):
I just checked - looks like the problem I'm talking about didn't propagate to GitHub yet and only resides in some private branch of @bdorney

Issue #208, opened 2019-05-22
in calibrateThrDac.py, Gaussian fits can fail


AndrewLevin:
The Gaussian fit to the scurveMean distribution can fail, and there is currently no protection against this.

## Brief summary of issue
In calibrateThrDac.py, a Gaussian is fitted to the per-vfat scurveMean distribution in order to determine the mean of the scurveMean distribution and the width of the scurveMean distribution. The Gaussian fit can fail, and it fails visibly in at least the 7 cases below.
![GE11-X-L-CERN-0002_2019 03 04 21 45_4_25 0](https://user-images.githubusercontent.com/3329216/58163048-968e9f00-7c83-11e9-8999-9d128ad387a8.png)
![GE11-X-L-GHENT-0002_2019 01 16 18 35_11_30 0](https://user-images.githubusercontent.com/3329216/58163050-968e9f00-7c83-11e9-948e-ed84fee0b9b9.png)
![GE11-X-L-GHENT-0002_2019 01 16 18 35_23_30 0](https://user-images.githubusercontent.com/3329216/58163051-97273580-7c83-11e9-8702-ff64fa975722.png)
![GE11-X-S-BARI-0012_2018 12 05 22 06_23_40 0](https://user-images.githubusercontent.com/3329216/58163053-97273580-7c83-11e9-9261-8079a423cd6e.png)
![GE11-X-S-BARI-0012_2018 12 05 22 38_3_60 0](https://user-images.githubusercontent.com/3329216/58163054-97273580-7c83-11e9-865f-672c00718258.png)
![GE11-X-S-CERN-0012_2018 08 10 18 47_18_150 0](https://user-images.githubusercontent.com/3329216/58163058-97273580-7c83-11e9-8645-d6a138f149b6.png)
![GEB3c-l1-w_2018 11 05 12 04_21_150 0](https://user-images.githubusercontent.com/3329216/58163060-97bfcc00-7c83-11e9-87b8-fcc2c1ac01d3.png)

### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

## Possible Solution
I would propose that, in calibrateThrDac.py, we remove points that satisfy scurveMeanError/scurveMean < 0.001. Also, we may want to look into how these histograms/tgraphs are binned in the scurve analysis script.

## Context (for feature requests)
[...]

PR #212, opened 2019-05-27
in calibrateThrDac.py, remove points with low scurveMeanError and restrict fit range

AndrewLevin:
Add protection against failed Gaussian fits to the scurveMean distribution and restrict the fit range automatically.

## Description
The scurve mean and the error on the scurve mean are determined by a Gaussian fit to a histogram (actually a TGraph) of the scurveMean. The Gaussian fit can fail in a way that produces a reasonable scurveMean but an scurveMeanError that is unreasonably small. This can cause the scurveMean vs THR_DAC fit to fail. So, this pull request removes points in which the scurveMeanError is less than 0.1% of the scurveMean error.

The new fit range restriction will not affect the fit, but will affect how the fitted function is plotted. The fit range is determined on a per vfat basis based on which points are included in the fit.

### Types of changes

## Motivation and Context
This address https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/208 including the comment https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/208#issuecomment-495305202.

## How Has This Been Tested?
Yes. As stated in the issue, among the 33 list of scandates files that I tested, there are 7 points that are removed by this algorithm. The scurveMean distributions shown there all indicate that it is reasonable to remove them. And the problematic scurveMean vs thrDac fit is improved in the case that motivated the issue (notice VFAT18):

Before:

58165543-45cd7500-7c88-11e9-8013-2ce581058492

With the new point removal algorithm:

canvScurveMeanVsThrDac_Summary 11

With the new point removal algorithm and the new restriction on the fit range:

## Checklist:

### DIFF
--- a/macros/calibrateThrDac.py
+++ b/macros/calibrateThrDac.py
@@ -494,7 +494,8 @@ if __name__ == '__main__':
             dict_ScurveMeanVsThrDac[vfat].GetPoint(thrDacIndexPairs[i][1],thrDacVal,scurveMean)
             scurveSigmaError = dict_ScurveSigmaVsThrDac[vfat].GetErrorY(thrDacIndexPairs[i][1])
             scurveMeanError = dict_ScurveMeanVsThrDac[vfat].GetErrorY(thrDacIndexPairs[i][1])
-            if scurveMean < 0.1 or scurveMeanError/scurveMean < 0.001:
+            from gempython.gemplotting.utils.anaInfo import scurveMeanMin, scurveMeanFracErrMin
+            if scurveMean < scurveMeanMin or scurveMeanError/scurveMean < scurveMeanFracErrMin:
                 continue
             if not setLastUnremovedScurveMean:
                 lastUnremovedScurveMean = scurveMean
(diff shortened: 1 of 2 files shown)


macros/calibrateThrDac.py
```diff
@@ -494,7 +494,7 @@
             dict_ScurveMeanVsThrDac[vfat].GetPoint(thrDacIndexPairs[i][1],thrDacVal,scurveMean)
             scurveSigmaError = dict_ScurveSigmaVsThrDac[vfat].GetErrorY(thrDacIndexPairs[i][1])
             scurveMeanError = dict_ScurveMeanVsThrDac[vfat].GetErrorY(thrDacIndexPairs[i][1])
-            if scurveMean < 0.1:
+            if scurveMean < 0.1 or scurveMeanError/scurveMean < 0.001:
```
jsturdy: I would put these values somewhere that they can be defined as either configuration parameters, or well-defined, documented constants

AndrewLevin: they are moved to anaInfo.py now

mexanick (changes requested):
please rebase on top of the current `develop`

AndrewLevin:
rebased

mexanick (approved):

jsturdy (approved):

Issue #63, opened 2018-01-18
Bug Report: gemSCurveAnaToolkit.py generates crash with ROOT v6.X or on cc7?


bdorney:
## Brief summary of issue

`gemSCurveAnaToolkit.py` generates a crash when running on cc7 with ROOT v6.X.  Suspect it's a difference between ROOT 5.34 and 6.X, unclear why.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [ ] Feature request (request for change which adds functionality)

### On `gem904daq03`
Works
```
[dorney@gem904daq03]~/scratch0/CMS_GEM/CMS_GEM_DAQ/gem-plotting-tools% gemSCurveAnaToolkit.py -i listOfScanDates_Scurve.txt -c -s30 -v1 --anaType=scurveAna --drawLeg
TClass::TClass:0: RuntimeWarning: no dictionary for class TF1Parameters is available
TStreamerInfo::BuildOld:0: RuntimeWarning: Cannot convert TF1::fParErrors from type:vector<double> to type:Double_t*, skip element
TStreamerInfo::BuildOld:0: RuntimeWarning: Cannot convert TF1::fParMin from type:vector<double> to type:Double_t*, skip element
TStreamerInfo::BuildOld:0: RuntimeWarning: Cannot convert TF1::fParMax from type:vector<double> to type:Double_t*, skip element
TStreamerInfo::BuildOld:0: RuntimeWarning: Cannot convert TF1::fSave from type:vector<double> to type:Double_t*, skip element
TStreamerInfo::BuildOld:0: RuntimeWarning: Cannot convert TF1::fParams from type:TF1Parameters* to type:Double_t*, skip element
Info in <TCanvas::Print>: png file canv_SCurveFitOverlay_VFAT1_Chan30.png has been created
eog canv_SCurveFitOverlay_GE11-X-S-CERN-0003_2018.01.17.14.29_VFAT1_Chan30.png

Your plot is stored in a TFile, to open it execute:
root /afs/cern.ch/user/d/dorney/scratch0/CMS_GEM/CMS_GEM_DAQ/data/gemelog/gemSCurveAnaToolkit.root
```

ROOT is 5.34/36
python is Python 2.6.6

### On `gem904qc8daq`
```
[...]

PR #213, opened 2019-06-03
in anaUltraScurve.py, TF1 associated with TTree branch should be copied into

AndrewLevin:
Bug was causing scurve fit branch to be corrupted in output tree.

## Description
Currently, we are correctly handling the `scurve_h` branch but not the `scurve_fit` branch:

https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/aacb46da4c6ebe952996ec8695584cd9bc926e9c/anaUltraScurve.py#L581-L583

This pull request uses the same logic involving a holder to handle the `scurve_fit` branch as the `scurve_h` branch.

### Types of changes

## Motivation and Context
Currently, the `scurve_fit` branch stored in the output tree does not contain the scurve fit, and it will induce a crash if it is read from. 

A similar issue was reported https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/63, although it is not clear to me if that is actually address by this pull request.

## How Has This Been Tested?
Yes, the bugfix was tested on GE11-X-S-PAK-0005 data taken by the QC8 team.

## Checklist:

### DIFF
--- a/anaUltraScurve.py
+++ b/anaUltraScurve.py
@@ -580,7 +580,8 @@ if __name__ == '__main__':
                 # Set TObjects linked to TBranches
                 holder_curve = fitter.scanHistos[vfat][chan]
                 holder_curve.Copy(scurve_h)
-                scurve_fit = fitter.getFunc(vfat,chan).Clone('scurveFit_vfat%i_chan%i'%(vfat,chan))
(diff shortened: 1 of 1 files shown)


bdorney (approved):

jsturdy (approved):

jsturdy:
It appears that you are now out-of-date due to an illegal push by @bdorney 3db691e

PR #215, opened 2019-06-06
New Feature: dacAnalysis raises ValueError if one or more DACs Out of Range

bdorney:
## Description

Following discussion in GEM Electronics meeting today it was suggested by VFAT3 team that we should reject VFATs found to have one or more DAC's that would need a DAC value out of the DAC range to achieve the proper bias voltage/current.

### Types of changes

The `dacAnalysis` function will now run to completion; but just before returning it will throw a `ValueError` if one or more DAC's are found to need a DAC value to achieve the correct bias voltage/current that would place the DAC out of it's range. 

## Motivation and Context

We need to be able to reject VFATs that have this DAC behavior; this provides an exception that can be caught by the `testConnectivity.py` routine to stop the procedure and warn the user.

## How Has This Been Tested?

Tested but don't know of a setup where this occurs at present.

## Checklist:

### DIFF
--- a/anaDACScan.py
+++ b/anaDACScan.py
@@ -81,9 +81,10 @@ if __name__ == '__main__':
         scandate = 'noscandate'
 
     from gempython.gemplotting.utils.anautilities import dacAnalysis
+    from gempython.gemplotting.utils.exceptions import VFATDACBiasCannotBeReached
     try:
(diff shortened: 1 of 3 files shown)


mexanick (approved):
Cursory

jsturdy (approved):
Cursory

PR #223, opened 2019-06-21
[Bug Fix] Corrected bug in nominalDacScalingFactors

bdorney:
## Description

In QC7 we noticed that for some VFATs the DAC scan for CFG_HYST would always fail to find a valid value; after some discussion in person with VFAT3 designers we determined that the value for the scaling factor needs to be updated.

### Types of changes

## Motivation and Context

So inside *some* DACs there is a factor that is inside the DAC itself that will change the input current/voltage; this current/voltage that the DAC actually uses is not accessible by either ADC when making a DAC scan but it is related to the actual current used by the fixed factor.

VFAT3 designers admitted this isn't very well described in the manual; but we went through and cross-checked each of the scaling factors.

It was found that only `CFG_HYST` was in error and it should be noted that for `CFG_CAL_DAC` the value we have in the `nominalDacScalingFactors` dictionary is only valid for the `voltageStep` injection mode and is *not* valid for the currentPulse mode.

## How Has This Been Tested?

Change was trivial

## Checklist:

### DIFF
--- a/utils/anaInfo.py
+++ b/utils/anaInfo.py
@@ -40,7 +40,7 @@ nominalDacValues = {
 
 #: From Tables 12 and 13 from the VFAT3 manual
 nominalDacScalingFactors = {
-        "CFG_CAL_DAC":10,
+        "CFG_CAL_DAC":10, # Valid only for currentPulse; if voltageStep this is 1
         "CFG_BIAS_PRE_I_BIT":0.2,
         "CFG_BIAS_PRE_I_BLCC":100,
         "CFG_BIAS_PRE_I_BSF":0.25,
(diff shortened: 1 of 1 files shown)


jsturdy (approved):
How do you want to treat the `CFG_CAL_DAC` depending on the pulse mode?

bdorney:
> How do you want to treat the `CFG_CAL_DAC` depending on the pulse mode?

Considering it's harder to convert voltage to charge than current to charge I disabled the scanning of `CFG_CAL_DAC` from the voltage buffer.  See:

https://github.com/cms-gem-daq-project/cmsgemos/blob/48e2f269dac0947be0101d2a9f0d943752e2ce61/gempython/tools/hw_constants.py#L17-L44

PR #231, opened 2019-07-09
[feature] Finding inflection point

bregnery:
This new feature finds the inflection point (knee) in the SBit rate scan and marks this point with a vertical line on the SBit rate scan graph. But this functionality is only available for the channelor case.

## Description
I created a new function, findInflectionPoint, and altered the sbitRateAnalysis function in `utils/anautilities.py`. The new function uses `numpy.gradient` along with some loops to find the inflection point. The altered sbitRateAnalysis draws a line in the sbit graph at the inflection point location. These changes are for the channelor case only.

### Types of changes

## Motivation and Context
We wanted to find the inflection point.

## How Has This Been Tested?
I used a python virtual environment on my account on gem904daq01. The code was tested using sbit rate data for the channelor case.

## Checklist:

### DIFF
--- a/utils/anautilities.py
+++ b/utils/anautilities.py
@@ -450,7 +450,7 @@ def filePathExists(searchPath, subPath=None, debug=False):
 
 def findInflectionPts(graph):
     '''
-    Find and return the "knee", inflection point of the SBit Rate scan
+    Find and return the "knee", inflection point, of the SBit Rate scan
     graph should be a TGraph or TH1F of the SBit Rate scan for 1 vfat 
     '''
 
@@ -1221,7 +1221,7 @@ def make2x4Canvas(name, initialContent = None, initialDrawOpt = '', secondaryCon
     canv.Update()
     return canv
 
-def make3x8Canvas(name, initialContent = None, initialDrawOpt = '', secondaryContent = None, secondaryDrawOpt = '', canv=None):
+def make3x8Canvas(name, initialContent = None, initialDrawOpt = '', secondaryContent = None, secondaryDrawOpt = '', canv = None ):
     """
     Creates a 3x8 canvas for summary plots.
 
@@ -1235,7 +1235,7 @@ def make3x8Canvas(name, initialContent = None, initialDrawOpt = '', secondaryCon
 
     import ROOT as r
     from ..mapping.chamberInfo import chamber_vfatPos2PadIdx
-    
+   
     if canv is None:
         canv = r.TCanvas(name,name,500*8,500*3)
         canv.Divide(8,3)
@@ -1840,6 +1840,8 @@ def sbitRateAnalysis(chamber_config, rateTree, cutOffRate=0.0, debug=False, outf
                 # Get Inflection Points /////////////////////////////////////////////////
                 #========================================================================
 
+                print("Find Inflection Point")
+                print "VFAT #: ", vfat
 		# Pseudocode
 		# make some container to store inflection Pts (e.g. nested dict like dict_dacValsBelowCutOff)
 		# the channel or case is easy:
@@ -1850,11 +1852,8 @@ def sbitRateAnalysis(chamber_config, rateTree, cutOffRate=0.0, debug=False, outf
 			# Then for channel bins in this tmp 2D histogram loop over them
 				# Make a projection using TH2D::ProjectionX or Y depending on which axis is vfatCH
 				# Add this projection to 1 tmp1D histogram using TH1F::Add()
-		# else:
-                print("Find Inflection Point")
-                print "VFAT #: ", vfat
+
                 if perchannel == False :
-	            #dict_InflectPt[dacName][ohKey][vfat] = findInflectionPts(dict_Rate1DVsDACNameX[dacName][ohKey][vfat])
 	            dict_dacInflectPts[dacName][ohKey][vfat] = findInflectionPts(dict_Rate1DVsDACNameX[dacName][ohKey][vfat])
 
 
@@ -1875,6 +1874,11 @@ def sbitRateAnalysis(chamber_config, rateTree, cutOffRate=0.0, debug=False, outf
                     pass
                 pass
             pass
+
+        #==========================================================================================
+        # Make Graphs /////////////////////////////////////////////////////////////////////////////
+        #==========================================================================================
+           
         for idx in range(len(dacNameArray)):
             dacName = np.asscalar(dacNameArray[idx])
             if perchannel:
@@ -1882,9 +1886,29 @@ def sbitRateAnalysis(chamber_config, rateTree, cutOffRate=0.0, debug=False, outf
                 for vfat in range(1,25):
                     canv_Summary2D.cd(vfat).SetLogz()
             else:
-                canv_Summary1D = make3x8Canvas("canv_Summary_Rate1D_vs_{0}".format(dacName),dict_Rate1DVsDACNameX[dacName][ohKey],"APE1")
-                for vfat in range(1,25) :
-                    canv_Summary1D.cd(vfat).SetLogy()
+                canv_Summary1D = make3x8Canvas("canv_Summary_Rate1D_vs_{0}".format(dacName),dict_Rate1DVsDACNameX[dacName][ohKey],"APE1" )
+                #for vfat in range(1,25):
+                for vfat in range(0,24):
+                    canv_Summary1D.cd(vfat + 1).SetLogy()
+
+                    # make TH1F into TGraph
+                    graph = dict_Rate1DVsDACNameX[dacName][ohKey][vfat] 
+                    if type(graph) == r.TH1F :
+                        graph = r.TGraph(graph)
+
+                    # get maximum y value
+                    y = graph.GetY()
+                    y = np.array(y)
+                    ymax = np.amax(y)
+
+                    # Draw a line on the graphs
+                    canv_Summary1D.cd(vfat + 1)
+                    print "Inflect x value: ", dict_dacInflectPts[dacName][ohKey][vfat][0]
+                    kneeLine = r.TLine(dict_dacInflectPts[dacName][ohKey][vfat][0], 1.0, dict_dacInflectPts[dacName][ohKey][vfat][0], ymax)
+                    kneeLine.SetLineColor(2)
+                    #kneeLine.Draw()
+                    #kneeLine.DrawLine(dict_dacInflectPts[dacName][ohKey][vfat][0], 1, dict_dacInflectPts[dacName][ohKey][vfat][0], ymax)
+                    kneeLine.DrawLineNDC(dict_dacInflectPts[dacName][ohKey][vfat][0], 1, dict_dacInflectPts[dacName][ohKey][vfat][0], ymax)
 
                 #====================================================================================
                 # Make Graphs ///////////////////////////////////////////////////////////////////////
@@ -1930,7 +1954,6 @@ def sbitRateAnalysis(chamber_config, rateTree, cutOffRate=0.0, debug=False, outf
                 fig.savefig('InflectionPointStudy.png')
                 fig.savefig('InflectionPointStudy.pdf')
   
-      
 
             if scandate == 'noscandate':
                 if perchannel:
--- a/utils/anautilities.py
+++ b/utils/anautilities.py
@@ -456,6 +456,9 @@ def findInflectionPts(graph):
     import numpy as np
     from itertools import groupby
 
+    # Allow for yellow and red warning colors
+    from gempython.utils.gemlogger import printYellow, printRed
+
     # make TH1F into TGraph
     if type(graph) == root.TH1F :
         graph = root.TGraph(graph)
@@ -477,6 +480,7 @@ def findInflectionPts(graph):
     # Sum the split arrays and find the most negative sum
     negSum = []
     bigNegSum = 0
+    bigIdx = 0
     for iNegArr in range(0, len(negGrad) ) :
         tmpNegSum = sum(negGrad[iNegArr] )
         negSum.append(tmpNegSum)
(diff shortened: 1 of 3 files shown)


bdorney:
Needs rebase ontop of develop

bdorney:
Do something like:

```
git fetch -p --all
git checkout develop
git pull <name_of_central_remote> develop
get checkout <name_of_your_local_feature_branch>
git rebase -i develop
```

You may need to manually resolve conflicts (follow on screen instructions).

Once `rebase` is finished and all conflicts have been resolved you'll need to force update your remote (because your commit history will have changed):

```
git push -f <name_of_your_remote> <name_of_your_local_feature_branch>
```

Only do this last step once you're sure everything is working!!!

jsturdy:
> Do something like:
> 
> ```
> git fetch -p --all
> git checkout develop
> git pull <name_of_central_remote> develop
> get checkout <name_of_your_local_feature_branch>
> git rebase -i develop
> ```

I have tried to recommend/stress that people should not keep personal versions (from their forks) of certain branches (`develop` is one of them), if you want a `develop` branch, you should ensure that it tracks the `cms-gem-daq-project/<repo>` `develop` branch (likewise for other branches).

```
* c3c7979        (HEAD -> bregnery/feature/inflectionPoint) Prepare for pull request
* e7265c3        Channelor case is ready
* a63f1b7        Working on perchannel case
* 4eb556c        Trying to update to use TLine
* 93b12e6        Added Inflection point finder
| * 581e241      (gemdaq/develop) Feature Updated ana_scans.py and made most analysis routines importable (#225)
| * e8c3f02      updating DB query to VFAT prod DB to fix bug caused by relying only on max(RUN_NUMBER), fixes issue #229 (#230)
|/  
* 135f4d8        (bregnery/develop) New Feature: Plotting Time Series HV Plots from DCS Monitoring Files (#222)

```

```
git remote add gemdaq git@github.com:cms-gem-daq-project/gem-plotting-tools.git
git checkout -t gemdaq/develop
## then every time you want to ensure your features are up to date
git fetch -p --all
git checkout develop
git pull
git checkout <my feature branch>
git rebase -i develop
```

What it seems that you did was to rebase your commits on top of **your** `develop` branch (possibly the intermediate steps @bdorney outlined didn't succeed), so they are still conflicting and out of date.

bdorney:
Indeed `name_of_central_remote` was meant to be from the cms-gem-daq-project.

jsturdy:
Clearly problems with rebase + refactor indicate that we will require some more guidelines

jsturdy:
I would like @bdorney's comment on how these additions should interplay with other tools, especially tools that would like to programatically extract the `CFG_THR_ARM_DAC` value corresponding to the knee, to understand if anything further would be required in this PR.
After that, you will, need to rebase again... (and I don't yet know what the triage should be between this PR and #238, as I suspect there will be conflicts during either rebase)

bdorney:
> I would like @bdorney's comment on how these additions should interplay with other tools, especially tools that would like to programatically extract the `CFG_THR_ARM_DAC` value corresponding to the knee, to understand if anything further would be required in this PR.

I'm not sure I understand your question; the original idea behind this knee finding was to determine the lower limit on where to perform the ARM DAC calibration.  But this is not extracted algorithmically presently.  The script behind this takes a user input:

https://github.com/cms-gem-daq-project/sw_utils/blob/develop/scripts/calibrateArmDac.sh

> After that, you will, need to rebase again... (and I don't yet know what the triage should be between this PR and #238, as I suspect there will be conflicts during either rebase)

Triage should just be to keep functions in required locations.  If this is not clear I can fork from @bregnery and do the cleanup myself and then submit a PR to his branch.

jsturdy (dismissed):
Conditionally approved: @bregnery as it seems that there is no additional feature to be added here, can you `rebase` again?

jsturdy (dismissed):

utils/anautilities.py
```diff
@@ -442,6 +442,59 @@ def filePathExists(searchPath, subPath=None, debug=False):
             print "Found %s"%s(testPath)
         return True
 
+#========================================================================================
+# Find Inflection Point /////////////////////////////////////////////////////////////////
+#----------------------------------------------------------------------------------------
+# Author(s): Brendan Regnery, Brian Dorney, Jared Sturdy ////////////////////////////////
+#----------------------------------------------------------------------------------------
```
bdorney: these comments should follow the sphinx style syntax.  The best way I think is to incorporate yourself into the `moduleauthor` field at the top of the `RST` block at the top of the file rather than the "per function" level.

utils/anautilities.py
```diff
@@ -442,6 +442,59 @@ def filePathExists(searchPath, subPath=None, debug=False):
...
+    x = np.array(x, dtype=float)
+    y = np.array(y, dtype=float)
+
+    # Calculate the gradient of y as a function of x
+    # Documentation here: https://docs.scipy.org/doc/numpy/reference/generated/numpy.gradient.html
+    grad = np.gradient(y, x)
+
+    # Split the array when the slope changes sign
+    posGrad = [list(g) for k, g in groupby(grad, lambda x: x > 0) if k]
+    negGrad = [list(g) for k, g in groupby(grad, lambda x: x < 0) if k]
```
bdorney, line 478: What are k & g supposed to be here?  Single letter variable names (even in list comprehension statements) are generally not a good idea.

utils/anautilities.py
```diff
@@ -442,6 +442,59 @@ def filePathExists(searchPath, subPath=None, debug=False):
...
+            bigNegSum = tmpNegSum
+            bigIdx =  iNegArr
+
+    # Get the inflection point
+    # We are defining the inflection point as the point where the most negative gradient sum begins
+    inflxGrad = negGrad[bigIdx][0]
+    for iGrad in range(0, len(grad) ) :
+        if grad[iGrad] == inflxGrad :
+            inflxIdx = iGrad
+    return [x[inflxIdx], y[inflxIdx] ]
```
bdorney: you're returning this as a list but as I understand this is an ordered pair; prefer `tuple` return type.

Also is there any case in which this would fail to find an ordered pair?  If so then won't `inflxIdx` be undefined and generate a `NameError`?

utils/anautilities.py
```diff
@@ -442,6 +442,59 @@ def filePathExists(searchPath, subPath=None, debug=False):
...
+        if tmpNegSum < bigNegSum :
+            bigNegSum = tmpNegSum
+            bigIdx =  iNegArr
+
+    # Get the inflection point
+    # We are defining the inflection point as the point where the most negative gradient sum begins
+    inflxGrad = negGrad[bigIdx][0]
+    for iGrad in range(0, len(grad) ) :
+        if grad[iGrad] == inflxGrad :
+            inflxIdx = iGrad
```
bdorney: Wouldn't it be more efficient to replace lines 493-495 with [`list::index()`](https://docs.python.org/2/tutorial/datastructures.html)? e.g.

```python
try:
   inflxIdx = grad.index(inflxGrad)
except ValueError
   print("Warning: inflxGrad value {:f} not found in grad container".format(inflxGrad))
   # Do appropriate error handling
```

utils/anautilities.py
jsturdy: I tend to (strongly) disagree in this case, primarily because the variables don't exist outside this single line scope, and the documentation, if needed, should be looked up in the `itertools.groupby` docs.
Here the comment already tells in a succinct way, what the lists are (well, it could explicitly state whether it stores the index of the input `grad` array, or the partitions of gradients for every sign change, but that's candy)

bdorney (changes requested):
small changes requested to implementation and comment location.

utils/threshAlgos.py
```diff
@@ -433,20 +433,20 @@ def anaUltraThreshold(args,thrFilename,GEBtype="short",outputDir=None,fileScurve
     # Do we return analyzed TTree?
     from multiprocessing import current_process
     if (current_process().name == 'MainProcess'):
-        return 
+        return
```
jsturdy: @bdorney c'est quoi ce truc?

bdorney: mmhmm seems my `.vimrc` isn't doing what I thought it was...I thought it should be stripping trailing white spaces automatically...I guess this is *not* the case :D

jsturdy (dismissed):
OK from my side, modulo comment on addition by @bdorney that was highlighted by a whitespace change in this PR

utils/threshAlgos.py
jsturdy: Nah, I meant this and the next line (double `return` statements)
(I don't really like *how* commented-on lines are selected by github...)

bdorney: the more `return` statements you have the faster you get your output right? right!?

utils/anautilities.py
```diff
@@ -442,6 +442,62 @@ def filePathExists(searchPath, subPath=None, debug=False):
...
+    try: 
+        inflxIdx = np.where(grad == inflxGrad) #return the index at the specified value
+    except ValueError:
+        print("Warning: the inflection point gradient (inflxGrad) value {:f} was not found in the gradient list".format(inflxGrad))
+    try:
+        inflxPnt = (x[inflxIdx], y[inflxIdx] )
+    except NameError:
+        print("Error: No Inflection Point was not found")
+
+    return inflxPnt
```
bdorney, line 512: this will still throw an exception since lines 494-497 don't actually resolve the `NameError`. I think the thing to do would be to do a:

```python
try:
        inflxPnt = (x[inflxIdx], y[inflxIdx] )
except NameError:
        printRed("Error: No Inflection Point was not found")
        # then re-raise the NameError that can be caught by the calling code
        # Quick google search gives me: http://www.ianbicking.org/blog/2007/09/re-raising-exceptions.html
else:
       return inflxPnt
```

utils/threshAlgos.py
bdorney: (indeed the `return thrAnaTree` is what is desired here)

utils/threshAlgos.py
```diff
@@ -1229,19 +1251,42 @@ def sbitRateAnalysis(chamber_config, rateTree, cutOffRate=0.0, debug=False, outf
...
+                    y = graph.GetY()
+                    y = np.array(y)
+                    ymax = np.amax(y)
+
+                    # Draw a line on the graphs
+                    kneeLine.append(r.TLine(dict_dacInflectPts[dacName][ohKey][vfat][0], 10.0, dict_dacInflectPts[dacName][ohKey][vfat][0], ymax) )
+                    kneeLine[vfat].SetLineColor(2)
+                    kneeLine[vfat].SetVertical()
+                    canv_Summary1D.cd(chamber_vfatPos2PadIdx[vfat] )
+                    kneeLine[vfat].Draw()
```
bdorney, line 1431: is this line position provided to the user in any tabular format or output data that could be read that would *not* require opening the TFile?

utils/anautilities.py
bdorney: you can import `printRed` I believe from:

```python
from gempython.utils.gemlogger import printRed
```

utils/threshAlgos.py
bregnery: I do not quite follow what you are looking for here.

jsturdy: How does someone running this analysis extract from the output the value of the `ARM_DAC` that the inflection point is at?
Can you print a table that says something like

| `VFAT` | `ARM_DAC` inflection point |
|----|----|
| 0 | 32 |
| 1 | 54 |
| 2 | 12 |

bregnery: Do you want me to put this in a text file? Or just have it print these values as the code runs?

bregnery: Update from @bdorney:

Do both. As the code run print the following:

| OHN | VFATN | ARM DAC Inflection Pt |
|--------|-----------|-------|
|   0     |     0       | 20  |
|   3     |    17	  | 45  |

	
and the output text file needs to be sent to the appropriate location under $DATA_PATH e.g. $DATA_PATH/detectorName/....etc...

For the text file drop the OHN column, since it will be only for one link.

bregnery: Okay, I added it and pushed

utils/threshAlgos.py
```diff
@@ -1035,6 +1035,16 @@ def sbitRateAnalysis(chamber_config, rateTree, cutOffRate=0.0, debug=False, outf
...
 
+    # Map TPad to correct vfat
+    from ..mapping.chamberInfo import chamber_vfatPos2PadIdx
+    from tabulate import tabulate
+
+    from gempython.gemplotting.utils.anautilities import findInflectionPts
+    from gempython.gemplotting.utils.anautilities import make3x8Canvas
+
+    # Allow for yellow warning color
+    from gempython.utils.gemlogger import printYellow
```
bdorney, line 1177: I would have these import statements be moved to where they are first used...this improves readability so you don't have to go through the code to understand where something comes from/what something is.

utils/threshAlgos.py
```diff
@@ -1229,19 +1253,48 @@ def sbitRateAnalysis(chamber_config, rateTree, cutOffRate=0.0, debug=False, outf
                     pass
                 pass
             pass
+
+        # print points in a table
+        if perchannel == False:
+            print(tabulate(inflectTable, headers = ['Geo Addr', 'VFAT Number', 'ARM DAC Inflection Pt'], tablefmt='orgtbl') )
+            inflectTableFile = file("{0}/{1}/inflectionPointTable.txt".format(elogPath,chamber_config[ohKey],scandate), "w")
```
bdorney: I was hoping this file would follow the output conventions in the rest of the script.  Explicitly see [lines 1150-1160](https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/231/files#diff-b72e0013378b9ec95b76d710ee83b7c2R1150-R1160).  Apologies for not being clear earlier

utils/threshAlgos.py
```diff
@@ -1229,19 +1253,48 @@ def sbitRateAnalysis(chamber_config, rateTree, cutOffRate=0.0, debug=False, outf
...
+            inflectTableFile.write(tabulate(inflectTable, headers = ['Geo Addr', 'VFAT Number', 'ARM DAC Inflection Pt'], tablefmt='orgtbl') )
```
bdorney: It seems inflectTable has information for *all* links (maybe I'm wrong).  But you're writing this file *per* link.  For the output file I would prefer to *not* have this information for all links but only the per link info in the per link file.

utils/threshAlgos.py
bdorney: still needs to be resolved.

bdorney (changes requested):
still a few outstanding issues; readability of imports can be improved and segmenting output inflection text files.

utils/threshAlgos.py
jsturdy: I do (and prefer) the opposite, sort of following `c/c++` style, but at the scope of relevance (`class`, `function`, file)...
[PEP8](https://www.python.org/dev/peps/pep-0008/#imports)  says they should *always* be at the top of the file, and as there isn't an established "style" here, it can be left up to the individual developer, with a slight lean towards the PEP

bdorney: this comment can be ignored then...would suggest that this type of style guide info makes it into the manifesto.

utils/threshAlgos.py
mexanick: Indeed, extra return has to be removed!

mexanick (changes requested):
One small issue with return statement

jsturdy (approved):

utils/threshAlgos.py
bregnery: The commit after this addresses this now and handles some additional error cases. 

https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/231/commits/09da4b11772349a57ef42208f654c5540c742066

bdorney (approved):

bdorney:
@mexanick anything outstanding on your side?

mexanick (approved):

Issue #235, opened 2019-07-10
ARM DAC calibration should consider channel masks


AndrewLevin:
We would like the ARM DAC calibration to fit only good data.

## Brief summary of issue

Currently, the ARM DAC calibration script https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/macros/calibrateThrDac.py performs the ARM DAC calibration based on TGraphs loaded from the root file that is output from the scurve analysis. These TGraphs are filled with all 128 channels, not considering the channel masks, e.g.:

https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/581e2414844bbba0ce5a122f4c67d7183b9d3daf/utils/scurveAlgos.py#L622-L636 So, we are fitting partially good and partially bad data, including channels which are masked as in-the-pedestal, which would affect the low ARM DAC region where problems have been reported by the QC8 team.

## Proposed Solution

@bdorney has proposed that we first redesign the loading of data in the ARM DAC calibration script such that it is based on the TTree, rather than the TGraphs, that is output from the scurve analysis. Then we apply the same quality cuts to the channel fits as are currently applied in the scurve analysis. The cut values are currently configurable as `argparse` options with default values, and @bdorney proposed to move them into https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/utils/anaoptions.py and import them into both the scurve analysis and the ARM DAC calibration.

I agree with this proposed solution.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

AndrewLevin:
Actually, given that we are storing the masks in the scurveFitTree, I think we can just re-use them rather than re-calculating them.

AndrewLevin:
Also, I noticed that the upper limit of the pedestal in the fit is set to nevents

https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/fitting/fitScanData.py#L305

which I believe is not even correct unitwise. It should be `self.calDAC2Q_m[vfat]*(256)+self.calDAC2Q_b[vfat]`, right?

bdorney:
> Also, I noticed that the upper limit of the pedestal in the fit is set to nevents
> 
> https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/fitting/fitScanData.py#L305
> 
[...]

PR #239, opened 2019-07-11
Apply scurve quality masks in ARM DAC calibration

AndrewLevin:
Applies quality cuts to the scurve fits that are used in the ARM DAC calibration with configurable cut values. 

## Description
This pull request makes the following changes:

1. Fixes an obvious mistake with the upper limit of the pedestal parameter in the scurve fit.
2. Adds the number of events to the TTree that is output from the scurve analysis. This is needed to apply the quality cut on the effective pedestal.
3. Moves the default cut values for the 4 mask criteria into `anaInfo.py`. The default values were taken from https://github.com/cms-gem-daq-project/sw_utils/blob/develop/scripts/calibrateArmDac.sh, not from the default values set are currently set in the `argparse` argument construction.
4. Moves the `argparse` block associated with these cut values into `anaoptions.py`.
5. Redesigns the feeding of data into the ARM DAC calibration such that it is based on the TTrees instead of the TGraphErrors that are output from the scurve analysis.
6. Removes the scurve fits for which any of the four quality mask criteria are satisfied.

### Types of changes

## Motivation and Context
Intended to resolve issue https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/235

## How Has This Been Tested?
Yes, I tested both the updated scurve analysis and the ARM DAC calibration on previously collected ARM DAC calibration data from the detector GE11-X-S-CERN-0010. The "canvScurveMeanVsThrDac_Summary" ARM DAC calibration plots are shown below.

With the head of develop:

With this pull request:

There are slight changes to the plots e.g. VFATs 5, 6, and 10. I would conclude that these changes are improvements, given that Gaussian fits to the scurveMean distribution that previously failed are now succeeding, and some of large error bars are reduced.

## Checklist:

### DIFF
--- a/ana_scans.py
+++ b/ana_scans.py
@@ -833,15 +833,6 @@ if __name__ == '__main__':
     cpuUsagee.add_argument("--light", action="store_true", help="Analysis uses only 25%% of available cores")
     cpuUsagee.add_argument("--medium", action="store_true", help="Analysis uses only 50%% of available cores")
     cpuUsagee.add_argument("--heavy", action="store_true", help="Analysis uses only 75%% of available cores")
-    
-    # create the parent parser for channel masks for scurve analysis
-    parser_scurveChanMasks = argparse.ArgumentParser(add_help = False)
-
-    chanMaskGroup = parser_scurveChanMasks.add_argument_group(title="Options for channel mask decisions", description="Parameters which specify how Dead, Noisy, and High Pedestal Channels are charaterized")
-    chanMaskGroup.add_argument("--maxEffPedPercent", type=float, default=0.02, help="Percentage, threshold for setting the HighEffPed mask reason, if channel (effPed > maxEffPedPercent * nevts) then HighEffPed is set")
-    chanMaskGroup.add_argument("--highNoiseCut", type=float, default=1.5, help="Threshold in fC for setting the HighNoise maskReason, if channel (scurve_sigma > highNoiseCut) then HighNoise is set")
-    chanMaskGroup.add_argument("--deadChanCutLow", type=float, default=0,help="If channel (deadChanCutLow < scurve_sigma < deadChanCutHigh) then DeadChannel is set")
-    chanMaskGroup.add_argument("--deadChanCutHigh", type=float, default=0, help="If channel (deadChanCutHigh < scurve_sigma < deadChanCutHigh) then DeadChannel is set")
 
     # create the parent parser for making output plots w.r.t ASIC channel or panasonic connector pin
     parser_stripChanOrPinType = argparse.ArgumentParser(add_help = False)
@@ -853,6 +844,8 @@ if __name__ == '__main__':
     parser_zscore = argparse.ArgumentParser(add_help = False)
     parser_zscore.add_argument("-z","--zscore", type=float, default=10,help="Z-Score for Outlier Identification in MAD Algo, used to set HotChannel bit")
 
+    from gempython.gemplotting.utils.anaoptions import parser_scurveChanMasks
+    
     # List of parent parsers specifically scurce analysis
     listOfParentParsers4Scurves = [parser_fileAndConfig, parser_parallelAna, parser_scurveChanMasks, parser_stripChanOrPinType, parser_zscore]
 
--- a/fitting/fitScanData.py
+++ b/fitting/fitScanData.py
@@ -302,11 +302,11 @@ class ScanDataFitter(DeadChannelFinder):
                     if self.isVFAT3:
                         fitTF1.SetParLimits(0, self.calDAC2Q_m[vfat]*(256)+self.calDAC2Q_b[vfat], self.calDAC2Q_m[vfat]*(1)+self.calDAC2Q_b[vfat])
                         fitTF1.SetParLimits(1, 0.0, self.calDAC2Q_m[vfat]*(128)+self.calDAC2Q_b[vfat])
-                        fitTF1.SetParLimits(2, -0.01, self.Nev[vfat][ch])
+                        fitTF1.SetParLimits(2, -0.01, self.calDAC2Q_m[vfat]*(1)+self.calDAC2Q_b[vfat])
                     else:
                         fitTF1.SetParLimits(0, -0.01, self.calDAC2Q_m[vfat]*(256)+self.calDAC2Q_b[vfat])
                         fitTF1.SetParLimits(1, 0.0,  self.calDAC2Q_m[vfat]*(128)+self.calDAC2Q_b[vfat])
-                        fitTF1.SetParLimits(2, -0.01, self.Nev[vfat][ch])
+                        fitTF1.SetParLimits(2, -0.01, self.calDAC2Q_m[vfat]*(256)+self.calDAC2Q_b[vfat])
                         pass
 
                     fitTF1.SetParLimits(3, 0.75*init_guess_p3, 1.25*init_guess_p3)
--- a/macros/calibrateThrDac.py
+++ b/macros/calibrateThrDac.py
@@ -85,9 +85,7 @@ if __name__ == '__main__':
     import argparse
     from gempython.gemplotting.utils.anaoptions import parser_scurveChanMasks
     parser = argparse.ArgumentParser(description='Arguments to supply to calibrateThrDac.py',parents=[parser_scurveChanMasks])
-    parser.add_argument("filename", type=str, dest="inputFile",
-            help="Tab delimited file specifying the input list of scandates, in three column format, specifying chamberName, scandate, and either CFG_THR_ARM_DAC or CFG_THR_ZCC_DAC value")
-
+    parser.add_argument("inputFile", type=str, help="Tab delimited file specifying the input list of scandates, in three column format, specifying chamberName, scandate, and either CFG_THR_ARM_DAC or CFG_THR_ZCC_DAC value")
     parser.add_argument("-d","--debug", action="store_true", help="Prints additional debugging information")
     parser.add_argument("--fitRange", type=str, default="0,255", 
             help="Two comma separated integers which specify the range of 'CFG_THR_*_DAC' to use in fitting when deriving the calibration curve")
@@ -104,8 +102,8 @@ if __name__ == '__main__':
     envCheck('ELOG_PATH')
 
     # Determine outputDir
-    if "/" in args.filename:
-        outputDir = args.filename[0:args.filename.rfind("/")+1]
+    if "/" in args.inputFile:
+        outputDir = args.inputFile[0:args.inputFile.rfind("/")+1]
     else:
         from gempython.gemplotting.utils.anautilities import getElogPath
         outputDir = getElogPath()
--- a/macros/calibrateThrDac.py
+++ b/macros/calibrateThrDac.py
@@ -112,7 +112,6 @@ if __name__ == '__main__':
     from gempython.gemplotting.utils.threshAlgos import calibrateThrDAC
     from gempython.utils.gemlogger import printGreen, printRed
     from gempython.utils.wrappers import runCommand
-    from gempython.gemplotting.utils.namespace import Namespace
     import os, sys, traceback
     try:
         args.outputDir = outputDir
--- a/macros/calibrateThrDac.py
+++ b/macros/calibrateThrDac.py
@@ -93,7 +93,7 @@ if __name__ == '__main__':
             help="Two comma separated integers which specify the range of 'CFG_THR_*_DAC' to use in fitting when deriving the calibration curve")
     parser.add_argument("--listOfVFATs", type=str, default=None,
             help="If provided the VFATID will be taken from this file rather than scurveTree.  Tab delimited file, first line is a column header, subsequent lines specify respectively VFAT position and VFAT serial number.  Lines beginning with the '#' character will be skipped")
-    parser.add_argument("--numOfGoodChannelsMin", type=int, default=10, help="Minimum number of channels that must be good (unmasked) for an armDacVal point to be use in the calibration procedure.")
+    parser.add_argument("--numOfGoodChansMin", type=int, default=10, help="Minimum number of channels that must be good (unmasked) for an armDacVal point to be use in the calibration procedure.")
     parser.add_argument("--noLeg", action="store_true", help="Do not draw a TLegend on the output plots")
     parser.add_argument("--savePlots", action="store_true", help="Make *.png file for all plots that will be saved in the output TFile")
     args = parser.parse_args()
--- a/macros/calibrateThrDac.py
+++ b/macros/calibrateThrDac.py
@@ -93,6 +93,7 @@ if __name__ == '__main__':
             help="Two comma separated integers which specify the range of 'CFG_THR_*_DAC' to use in fitting when deriving the calibration curve")
     parser.add_argument("--listOfVFATs", type=str, default=None,
             help="If provided the VFATID will be taken from this file rather than scurveTree.  Tab delimited file, first line is a column header, subsequent lines specify respectively VFAT position and VFAT serial number.  Lines beginning with the '#' character will be skipped")
+    parser.add_argument("--numOfGoodChannelsMin", type=int, default=10, help="Minimum number of channels that must be good (unmasked) for an armDacVal point to be use in the calibration procedure.")
     parser.add_argument("--noLeg", action="store_true", help="Do not draw a TLegend on the output plots")
     parser.add_argument("--savePlots", action="store_true", help="Make *.png file for all plots that will be saved in the output TFile")
     args = parser.parse_args()
--- a/macros/calibrateThrDac.py
+++ b/macros/calibrateThrDac.py
@@ -85,7 +85,7 @@ if __name__ == '__main__':
     import argparse
     from gempython.gemplotting.utils.anaoptions import parser_scurveChanMasks
     parser = argparse.ArgumentParser(description='Arguments to supply to calibrateThrDac.py',parents=[parser_scurveChanMasks])
-    parser.add_argument("filename", type=str, 
+    parser.add_argument("filename", type=str, dest="inputFile",
             help="Tab delimited file specifying the input list of scandates, in three column format, specifying chamberName, scandate, and either CFG_THR_ARM_DAC or CFG_THR_ZCC_DAC value")
 
     parser.add_argument("-d","--debug", action="store_true", help="Prints additional debugging information")
@@ -116,23 +116,9 @@ if __name__ == '__main__':
     from gempython.gemplotting.utils.namespace import Namespace
     import os, sys, traceback
     try:
-
-        args = Namespace(
-            inputFile=args.filename,
-            fitRange=args.fitRange,
-            listOfVFATs=args.listOfVFATs,
-            maxEffPedPercent=args.maxEffPedPercent,
-            highNoiseCut=args.highNoiseCut,
-            deadChanCutLow=args.deadChanCutLow,
-            deadChanCutHigh=args.deadChanCutHigh,
-            noLeg=args.noLeg,
-            outputDir=outputDir,
-            savePlots=args.savePlots,
-            debug=args.debug
-        )
+        args.outputDir = outputDir
 
         retCode = calibrateThrDAC(args)
-
     except IOError as err:
         printRed("IOError: {0}".format(err.message))
         printRed("Analysis failed with error code {0}".format(retCode))
--- a/macros/calibrateThrDac.py
+++ b/macros/calibrateThrDac.py
@@ -83,7 +83,8 @@ Internals
 if __name__ == '__main__':
     # create the parser
     import argparse
-    parser = argparse.ArgumentParser(description='Arguments to supply to calibrateThrDac.py')
+    from gempython.gemplotting.utils.anaoptions import parser_scurveChanMasks
+    parser = argparse.ArgumentParser(description='Arguments to supply to calibrateThrDac.py',parents=[parser_scurveChanMasks])
     parser.add_argument("filename", type=str, 
             help="Tab delimited file specifying the input list of scandates, in three column format, specifying chamberName, scandate, and either CFG_THR_ARM_DAC or CFG_THR_ZCC_DAC value")
 
@@ -112,16 +113,26 @@ if __name__ == '__main__':
     from gempython.gemplotting.utils.threshAlgos import calibrateThrDAC
     from gempython.utils.gemlogger import printGreen, printRed
     from gempython.utils.wrappers import runCommand
+    from gempython.gemplotting.utils.namespace import Namespace
     import os, sys, traceback
     try:
-        retCode = calibrateThrDAC(
-                inputFile=args.filename,
-                fitRange=args.fitRange,
-                listOfVFATs=args.listOfVFATs,
-                noLeg=args.noLeg,
-                outputDir=outputDir,
-                savePlots=args.savePlots,
-                debug=args.debug)
+
+        args = Namespace(
+            inputFile=args.filename,
+            fitRange=args.fitRange,
+            listOfVFATs=args.listOfVFATs,
+            maxEffPedPercent=args.maxEffPedPercent,
+            highNoiseCut=args.highNoiseCut,
+            deadChanCutLow=args.deadChanCutLow,
+            deadChanCutHigh=args.deadChanCutHigh,
+            noLeg=args.noLeg,
+            outputDir=outputDir,
+            savePlots=args.savePlots,
+            debug=args.debug
+        )
+
+        retCode = calibrateThrDAC(args)
+
     except IOError as err:
         printRed("IOError: {0}".format(err.message))
         printRed("Analysis failed with error code {0}".format(retCode))
--- a/macros/calibrateThrDac.py
+++ b/macros/calibrateThrDac.py
@@ -113,9 +113,8 @@ if __name__ == '__main__':
     from gempython.utils.gemlogger import printGreen, printRed
     from gempython.utils.wrappers import runCommand
     import os, sys, traceback
+    args.outputDir = outputDir
     try:
-        args.outputDir = outputDir
-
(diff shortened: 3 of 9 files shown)


utils/anaoptions.py
```diff
@@ -11,15 +11,25 @@
...
-parent_parser.add_argument("-o", "--outfilename", type=str, help="Specify Output Filename")
+parentParser1.add_argument("-d", "--debug", action="store_true", help="print extra debugging information")
+parentParser1.add_argument("-o", "--outfilename", type=str, help="Specify Output Filename")
 
-stripChanOrPinType = parent_parser.add_mutually_exclusive_group(required=False)
+stripChanOrPinType = parentParser1.add_mutually_exclusive_group(required=False)
 stripChanOrPinType.add_argument("-c","--channels", action="store_true", help="Make plots vs channels instead of strips")
 stripChanOrPinType.add_argument("-p","--panasonic", action="store_true", dest="PanPin",help="Make plots vs Panasonic pins instead of strips")
+
+parentParser2 = argparse.ArgumentParser(add_help = False)
```
bdorney: Previously the name `parent_parser` was a descriptive and unique variable name that all things "inherited" from; but by inserting a second one and changing to names `parentParser1` and `parentParser2` you have really degraded the overall readability of the code base; what I was going for more was a common sense copy/paste from `ana_scans.py`: e.g.

```python
    # create the parent parser for channel masks for scurve analysis	
    parser_scurveChanMasks = argparse.ArgumentParser(add_help = False)	

     chanMaskGroup = parser_scurveChanMasks.add_argument_group(title="Options for channel mask decisions", description="Parameters which specify how Dead, Noisy, and High Pedestal Channels are charaterized")	
    chanMaskGroup.add_argument("--maxEffPedPercent", type=float, default=0.02, help="Percentage, threshold for setting the HighEffPed mask reason, if channel (effPed > maxEffPedPercent * nevts) then HighEffPed is set")	
    chanMaskGroup.add_argument("--highNoiseCut", type=float, default=1.5, help="Threshold in fC for setting the HighNoise maskReason, if channel (scurve_sigma > highNoiseCut) then HighNoise is set")	
    chanMaskGroup.add_argument("--deadChanCutLow", type=float, default=0,help="If channel (deadChanCutLow < scurve_sigma < deadChanCutHigh) then DeadChannel is set")	
    chanMaskGroup.add_argument("--deadChanCutHigh", type=float, default=0, help="If channel (deadChanCutHigh < scurve_sigma < deadChanCutHigh) then DeadChannel is set")
```   

And then in the three files discussed to import `parser_scurveChanMasks` and add this to the list of parent parsers that are provided to various analysis tools.

Unfortunately you've rolled all your changes into a *single* commit which isn't compartmentalized so you cannot easily roll back these changes.

Please do the above and revert changes, e.g. `parentParser1` -> `parent_parser` then revert changes to files not mentioned in issue #235.  Then squash this change commit so that the history of those files you will revert is not affected; this might not be the correct `git` protocol; @jsturdy can better advise here.

fitting/fitScanData.py
```diff
@@ -302,11 +302,11 @@ def fit(self, debug=False):
                     if self.isVFAT3:
                         fitTF1.SetParLimits(0, self.calDAC2Q_m[vfat]*(256)+self.calDAC2Q_b[vfat], self.calDAC2Q_m[vfat]*(1)+self.calDAC2Q_b[vfat])
                         fitTF1.SetParLimits(1, 0.0, self.calDAC2Q_m[vfat]*(128)+self.calDAC2Q_b[vfat])
-                        fitTF1.SetParLimits(2, -0.01, self.Nev[vfat][ch])
+                        fitTF1.SetParLimits(2, -0.01, self.calDAC2Q_m[vfat]*(1)+self.calDAC2Q_b[vfat])
                     else:
                         fitTF1.SetParLimits(0, -0.01, self.calDAC2Q_m[vfat]*(256)+self.calDAC2Q_b[vfat])
                         fitTF1.SetParLimits(1, 0.0,  self.calDAC2Q_m[vfat]*(128)+self.calDAC2Q_b[vfat])
-                        fitTF1.SetParLimits(2, -0.01, self.Nev[vfat][ch])
+                        fitTF1.SetParLimits(2, -0.01, self.calDAC2Q_m[vfat]*(256)+self.calDAC2Q_b[vfat])
```
bdorney, line 309: And did you see any real improvement to the fitting procedure with this change?

ana_scans.py
```diff
@@ -841,8 +832,10 @@ def trimParallelAna(args):
     parser_zscore = argparse.ArgumentParser(add_help = False)
     parser_zscore.add_argument("-z","--zscore", type=float, default=10,help="Z-Score for Outlier Identification in MAD Algo, used to set HotChannel bit")
 
+    from gempython.gemplotting.utils.anaoptions import parentParser2
+    
     # List of parent parsers specifically scurce analysis
-    listOfParentParsers4Scurves = [parser_fileAndConfig, parser_parallelAna, parser_scurveChanMasks, parser_stripChanOrPinType, parser_zscore]
+    listOfParentParsers4Scurves = [parser_fileAndConfig, parser_parallelAna, parentParser2, parser_stripChanOrPinType, parser_zscore]
```
bdorney: revert name change as discussed [here](https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/239/files#r302651222)

macros/calibrateThrDac.py
```diff
@@ -83,7 +83,8 @@
 if __name__ == '__main__':
     # create the parser
     import argparse
-    parser = argparse.ArgumentParser(description='Arguments to supply to calibrateThrDac.py')
+    from gempython.gemplotting.utils.anaoptions import parentParser2
+    parser = argparse.ArgumentParser(description='Arguments to supply to calibrateThrDac.py',parents=[parentParser2])
```
bdorney: name change

utils/anaInfo.py
```diff
@@ -17,6 +17,12 @@
 scurveMeanMin = 0.1 #: points are removed if they satisfy scurveMean < scurveMeanMin
 scurveMeanFracErrMin = 0.001 #: points are removed if they satisfy scurveMeanError/scurveMean < scurveFracErrMin
 
+#: The default values for the cuts that determine the scurve fit quality masks
+maxEffPedPercentDefault=0.1
+highNoiseCutDefault=20
+deadChanCutLowDefault=0
+deadChanCutHighDefault=0
```
bdorney, line 25: Are these constants used anywhere else aside from `utils/anaoptions.py`?  If not remove.  Additionally these defaults do not match what was currently implemented.

utils/threshAlgos.py
```diff
@@ -451,7 +451,9 @@ def calibrateThrDACStar(inputs):
     """
     return calibrateThrDAC(*inputs)
 
-def calibrateThrDAC(inputFile,fitRange="0,255",listOfVFATs=None,noLeg=False,outputDir=None,savePlots=False,debug=False):
+from anaInfo import maxEffPedPercentDefault, highNoiseCutDefault, deadChanCutLowDefault, deadChanCutHighDefault
+
+def calibrateThrDAC(inputFile,fitRange="0,255",listOfVFATs=None,maxEffPedPercent=maxEffPedPercentDefault,highNoiseCut=highNoiseCutDefault,deadChanCutLow=deadChanCutLowDefault,deadChanCutHigh=deadChanCutHighDefault,noLeg=False,outputDir=None,savePlots=False,debug=False):
```
bdorney: consider passing a namespace rather than a large argument list and then checking if expected entries exist and and if they do not exist either add them with a default option or raise an appropriate exception.

utils/scurveAlgos.py
```diff
@@ -478,6 +478,8 @@ def anaUltraScurve(args, scurveFilename, calFile=None, GEBtype="short", outputDi
         scurveFitTree.Branch( 'pedestal', pedestal, 'pedestal/F')
         ped_eff = array( 'f', [ 0 ] )
         scurveFitTree.Branch( 'ped_eff', ped_eff, 'ped_eff/F')
+        nevents = array( 'i', [ 0 ] )
+        scurveFitTree.Branch( 'nevents', nevents, 'nevents/I')
```
bdorney: Adding a new branch, which increases memory for the output file unnecessarily (e.g. because `nevents` will be the same for entries in the `TTree`) it's probably better to simply change `ped_eff` to store a percentage rather than an integer.  The pro's and cons of this should have been discussed in the issue.

utils/threshAlgos.py
```diff
@@ -625,10 +637,43 @@ def calibrateThrDAC(inputFile,fitRange="0,255",listOfVFATs=None,noLeg=False,outp
...
+                thisVFAT_ENCMean = np.mean(scurveFitDataThisVfat["noise"])
+                thisVFAT_ENCStd = np.std(scurveFitDataThisVfat["noise"])
+                histENC = r.TH1F("scurveSigma_vfat%i"%vfat,"VFAT %i;S-Curve Sigma #left(fC#right);N"%vfat,
+                                 40, thisVFAT_ENCMean - 5. * thisVFAT_ENCStd, thisVFAT_ENCMean + 5. * thisVFAT_ENCStd )
+        
+            for idy in range(0,len(scurveFitDataThisVfat)):
+                scurveMean = scurveFitDataThisVfat[idy]['threshold']
+                scurveSigma = scurveFitDataThisVfat[idy]['noise']
+                histThresh.Fill(scurveMean)
+                histENC.Fill(scurveSigma)
```
bdorney, line 733: What is it that you're adding with this block? It wasn't discussed in the issue.

bdorney:
So in addition to the comments I've put above I've listed this as "blocked" and "on hold" @jsturdy has requested that PR's should be first discussed in issues and algorithm's outlined before submitting a PR.  There have been several changes here that where not originally outlined in the issue.

Before making the PR it would have been good to see a thorough comparison of these plots (at the very least on the same `TPad`) not to mention something more indepth (e.g. how do pull's/residuals between data & fit behave) and then some plan to attempt to compare trimming before/after this change.

utils/anaoptions.py
AndrewLevin: done

fitting/fitScanData.py
AndrewLevin: No, I did not expect any change to the fitting procedure, but we just got lucky in some sense because the number of events we typically use (100) is larger than the pedestal values.

ana_scans.py
AndrewLevin: done

macros/calibrateThrDac.py
AndrewLevin: done

utils/anaInfo.py
AndrewLevin: they are also used in utils/threshAlgos.py to set the arguments of the function calibrateThrDAC(...) if they are not provided

AndrewLevin: the defaults are changed back to the ones currently implemented (I took the ones above from https://github.com/cms-gem-daq-project/sw_utils/blob/develop/scripts/calibrateArmDac.sh)

utils/threshAlgos.py
AndrewLevin: done

utils/scurveAlgos.py
AndrewLevin: `ped_eff` is redefined as `ped_eff/nevents` now

utils/threshAlgos.py
AndrewLevin: We are loading data into the ARM DAC calibration script from the TTree, instead of the TGraphs because we need to apply the quality cuts.

bdorney: As I mentioned in your follow-up comment on the github issue it would be good to add some comments here to "guide" the reader.

AndrewLevin: code streamlined and more comments added

AndrewLevin:
I found that in two cases (vfat1 and vfat2 in slide 8 of https://indico.cern.ch/event/837005/contributions/3509259/attachments/1884624/3106257/ALevin_calibrateThrDac_SWDevMeeting_23Jul2019.pdf), after applying the channel masks, a Gaussian fit was performed on a TGraph which was constructed from a single channel, which resulted in an unphysical negative Gaussian sigma. These points were already excluded from the fit, but they were still shown as non-zero points on the plot. Now, I have added protection against this by requiring that at least 10 channels are unmasked, otherwise the Gaussian fit is not attempted. The effect of this change can be seen below for the same example as in the slides (GE11-X-S-PAK-0007):

Before:

After:

bdorney:
Looks like this needs a `rebase`.

Please let us know when this is done and it's ready for re-review.

AndrewLevin:
It is ready for re-review. I have uploaded some slides for tomorrow's meeting: https://indico.cern.ch/event/838248/contributions/3515801/attachments/1887448/3111845/ALevin_calibrateThrDac_SWDevMeeting_30Jul2019.pdf

macros/calibrateThrDac.py
```diff
@@ -112,16 +113,26 @@
...
+            listOfVFATs=args.listOfVFATs,
+            maxEffPedPercent=args.maxEffPedPercent,
+            highNoiseCut=args.highNoiseCut,
+            deadChanCutLow=args.deadChanCutLow,
+            deadChanCutHigh=args.deadChanCutHigh,
+            noLeg=args.noLeg,
+            outputDir=outputDir,
+            savePlots=args.savePlots,
+            debug=args.debug
+        )
```
bdorney: Here it seems you're just reseting `args` to itself and changing the name of one member and adding another (e.g. `inputFile` and `outputDir`).  Rather than reseting the whole object it would be simpler to change line 88 to:

```python
parser.add_argument("filename", type=str, dest="inputFile",
```

Then the `filename` argument is stored in `args.inputFile` and then to just add `outputDir` as a single line, i.e.:

```python
args.outputDir = outputDir
``` 

This is much simpler than what you've done in lines 120-132.

Finally typically you should place the minimum amount of code inside of a `try` block and have what is inside the `try` block be only relevant to your exception handling.  Right now the exception handling is designed to catch only `IOError`s and manipulating `args` has nothing to do with this.  So the only thing that should be in this `try` block should be line 134.

utils/namespace.py
```diff
@@ -0,0 +1,3 @@
+class Namespace:
+    def __init__(self, **kwargs):
+        self.__dict__.update(kwargs)
```
bdorney: based on proposed changes [here](https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/239/files#r308269996) I would drop this file from the PR.

utils/threshAlgos.py
```diff
@@ -472,39 +474,69 @@ def calibrateThrDAC(inputFile,fitRange="0,255",listOfVFATs=None,noLeg=False,outp
         savePlots     - Make a '*.png' file for all plots that will be saved in the output TFile
     """
```
bdorney, line 483: You need to update the function documentation to indicate what `args` is and what fields are expected.

utils/threshAlgos.py
```diff
@@ -472,39 +474,69 @@ def calibrateThrDAC(inputFile,fitRange="0,255",listOfVFATs=None,noLeg=False,outp
...
 
+    if not hasattr(args,"inputFile"):
+        print("No inputFile provided to calibrateThrDac")
+        sys.exit(1)
```
bdorney: it would be more appropriate to provide the exit code [OS_NOINPUT](https://docs.python.org/2/library/os.html#os.EX_NOINPUT) here.

utils/threshAlgos.py
```diff
@@ -472,39 +474,69 @@ def calibrateThrDAC(inputFile,fitRange="0,255",listOfVFATs=None,noLeg=False,outp
...
+        
+    if not hasattr(args,"fitRange"):
+        args.fitRange = "0,255"
```
bdorney: change to:

```python
if not hasattr(args,"fitRange"):
   args.fitRange = [0, 255]
else:
   args.fitRange = [int(item) for item in args.fitRange.split(",")]
```

Then drop line 861

utils/threshAlgos.py
```diff
@@ -626,10 +667,45 @@ def calibrateThrDAC(inputFile,fitRange="0,255",listOfVFATs=None,noLeg=False,outp
...
+                
+            histThresh = r.TH1F("scurveMean_vfat%i"%vfat,"VFAT %i;S-Curve Mean #left(fC#right);N"%vfat, 
+                                40, thisVFAT_ThreshMean - 5. * thisVFAT_ThreshStd, thisVFAT_ThreshMean + 5. * thisVFAT_ThreshStd )
+            histENC = r.TH1F("scurveSigma_vfat%i"%vfat,"VFAT %i;S-Curve Sigma #left(fC#right);N"%vfat,
+                                 40, thisVFAT_ENCMean - 5. * thisVFAT_ENCStd, thisVFAT_ENCMean + 5. * thisVFAT_ENCStd )
+                
+            #fill histograms with the scurve means and the scurve sigmas that pass the quality cuts
+            #leave TH1Fs and TGraphs empty for ARM DAC points with too few unmasked channels
+            from gempython.gemplotting.utils.anaInfo import numOfGoodChannelsMin
+            if len(scurveFitDataThisVfat) >= numOfGoodChannelsMin:
```
bdorney: `numOfGoodChannelsMin` defaults to 10; but what if a user had intentionally only scanned 1 channel at each `CFG_THR_ARM_DAC` value?  They would be unable to use this tool in the analysis.  This should be an optional input argument that the user may provide that defaults to 10, and if the input namespace does not have this attribute it should be set to `numOfGoodChannelsMin`.

bdorney (changes requested):
some simplifications can be made and an additional optional argument should be provided.

utils/namespace.py
AndrewLevin: done

utils/threshAlgos.py
AndrewLevin: done

utils/threshAlgos.py
AndrewLevin: done

utils/threshAlgos.py
AndrewLevin: done

utils/threshAlgos.py
AndrewLevin: Done. However note that the case of scanning a single channel is not really supported, we are performing Gaussian fits and then applying cuts to the errors of those Gaussian fits. If we would like to support the case of scanning a single channel, I can make another github issue and pull request.

macros/calibrateThrDac.py
AndrewLevin: Done.

AndrewLevin:
All of your requested changes have been implemented.

macros/calibrateThrDac.py
AndrewLevin: Instead of adding the destination for the `filename` argument, I just changed `filename` to `inputFile`. I think adding a different destination for a positional argument is not allowed.

bdorney: I would not change the interface that the user sees.  And the `dest` argument for `add_argument` can be used regardless of whether or not the the argument is positional or optional.

https://docs.python.org/2/library/argparse.html#the-add-argument-method

bdorney:
So following feedback from SW Dev meeting we agreed (@jsturdy, myself and @AndrewLevin) to:

- Do not plot the points with zero values,
- Add a table, like in scurve analysis, that shows the number of rejected channels per VFAT by category, see [Lines 383-443](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/8a2c3b8a9d1e1dce6c98490a84f8ae022a14051f/utils/scurveAlgos.py#L383-L443)

AndrewLevin:
I have made those changes in the last two commits.

TGraphs with zeroed points removed:

Bad channel summary information printout:

```
| vfatN | armDacVal | Dead Chan | High Noise | High Eff Ped | Fit At Init Val |
| :---: | :-------: | :-------: | :--------: | :----------: | :-------------: |
| -1 | 17 | 48 | 1671 | 2935 | 48 |
| -1 | 20 | 47 | 1974 | 2934 | 47 |
| -1 | 23 | 48 | 2211 | 2936 | 48 |
| -1 | 25 | 46 | 2272 | 2927 | 46 |
| -1 | 30 | 49 | 596 | 1563 | 49 |
| -1 | 40 | 48 | 349 | 803 | 48 |
| -1 | 50 | 49 | 308 | 481 | 49 |
| -1 | 60 | 48 | 282 | 360 | 48 |
| -1 | 70 | 49 | 254 | 271 | 49 |
| -1 | 80 | 48 | 240 | 161 | 48 |
| -1 | 100 | 50 | 222 | 72 | 50 |
| -1 | 125 | 49 | 228 | 55 | 49 |
| -1 | 150 | 49 | 196 | 49 | 49 |
| -1 | 175 | 77 | 210 | 77 | 77 |
| 0 | 17 | 0 | 57 | 128 | 0 |
| 0 | 20 | 0 | 80 | 128 | 0 |
| 0 | 23 | 0 | 76 | 128 | 0 |
| 0 | 25 | 0 | 72 | 128 | 0 |
| 0 | 30 | 0 | 3 | 38 | 0 |
| 0 | 40 | 0 | 1 | 15 | 0 |
| 0 | 50 | 0 | 1 | 7 | 0 |
| 0 | 60 | 0 | 1 | 7 | 0 |
| 0 | 70 | 0 | 1 | 7 | 0 |
| 0 | 80 | 0 | 1 | 2 | 0 |
| 0 | 100 | 0 | 1 | 0 | 0 |
| 0 | 125 | 0 | 1 | 0 | 0 |
| 0 | 150 | 0 | 1 | 0 | 0 |
| 0 | 175 | 0 | 1 | 0 | 0 |
| 1 | 17 | 0 | 45 | 128 | 0 |
| 1 | 20 | 0 | 52 | 128 | 0 |
| 1 | 23 | 0 | 54 | 128 | 0 |
| 1 | 25 | 0 | 58 | 128 | 0 |
| 1 | 30 | 0 | 1 | 36 | 0 |
| 1 | 40 | 0 | 1 | 9 | 0 |
| 1 | 50 | 0 | 1 | 7 | 0 |
| 1 | 60 | 0 | 1 | 9 | 0 |
| 1 | 70 | 0 | 1 | 8 | 0 |
| 1 | 80 | 0 | 1 | 0 | 0 |
| 1 | 100 | 0 | 1 | 0 | 0 |
| 1 | 125 | 0 | 1 | 0 | 0 |
| 1 | 150 | 0 | 1 | 0 | 0 |
| 1 | 175 | 0 | 1 | 0 | 0 |
...
```

jsturdy:
I don't understand the table; ~~I expect one entry per VFAT~~ (ah, one per VFAT per DAC point), no VFATs indexed as `-1`, and **at most** a value of 128 in each of the columns.
If more than one link is possible by this tool, then as many additional fields as necessary to make the results unique (`shelf`/`AMC`/`link`)

AndrewLevin:
The tool handles a single detector at a time.

AndrewLevin:
vfatN = -1 is collective information for all VFATs

bdorney:
> vfatN = -1 is collective information for all VFATs

For this case instead of reporting `-1` I would report `All` in that column in this case.

AndrewLevin:
Done:

```
| vfatN | armDacVal | Dead Chan | High Noise | High Eff Ped | Fit At Init Val |
| :---: | :-------: | :-------: | :--------: | :----------: | :-------------: |
| All | 17 | 48 | 1671 | 2935 | 48 |
| All | 20 | 47 | 1974 | 2934 | 47 |
| All | 23 | 48 | 2211 | 2936 | 48 |
| All | 25 | 46 | 2272 | 2927 | 46 |
| All | 30 | 49 | 596 | 1563 | 49 |
| All | 40 | 48 | 349 | 803 | 48 |
| All | 50 | 49 | 308 | 481 | 49 |
| All | 60 | 48 | 282 | 360 | 48 |
| All | 70 | 49 | 254 | 271 | 49 |
| All | 80 | 48 | 240 | 161 | 48 |
| All | 100 | 50 | 222 | 72 | 50 |
| All | 125 | 49 | 228 | 55 | 49 |
| All | 150 | 49 | 196 | 49 | 49 |
| All | 175 | 77 | 210 | 77 | 77 |
| 0 | 17 | 0 | 57 | 128 | 0 |
| 0 | 20 | 0 | 80 | 128 | 0 |
| 0 | 23 | 0 | 76 | 128 | 0 |
| 0 | 25 | 0 | 72 | 128 | 0 |
| 0 | 30 | 0 | 3 | 38 | 0 |
| 0 | 40 | 0 | 1 | 15 | 0 |
| 0 | 50 | 0 | 1 | 7 | 0 |
| 0 | 60 | 0 | 1 | 7 | 0 |
| 0 | 70 | 0 | 1 | 7 | 0 |
| 0 | 80 | 0 | 1 | 2 | 0 |
| 0 | 100 | 0 | 1 | 0 | 0 |
| 0 | 125 | 0 | 1 | 0 | 0 |
| 0 | 150 | 0 | 1 | 0 | 0 |
| 0 | 175 | 0 | 1 | 0 | 0 |
| 1 | 17 | 0 | 45 | 128 | 0 |
| 1 | 20 | 0 | 52 | 128 | 0 |
| 1 | 23 | 0 | 54 | 128 | 0 |
| 1 | 25 | 0 | 58 | 128 | 0 |
| 1 | 30 | 0 | 1 | 36 | 0 |
| 1 | 40 | 0 | 1 | 9 | 0 |
| 1 | 50 | 0 | 1 | 7 | 0 |
| 1 | 60 | 0 | 1 | 9 | 0 |
| 1 | 70 | 0 | 1 | 8 | 0 |
| 1 | 80 | 0 | 1 | 0 | 0 |
| 1 | 100 | 0 | 1 | 0 | 0 |
| 1 | 125 | 0 | 1 | 0 | 0 |
| 1 | 150 | 0 | 1 | 0 | 0 |
| 1 | 175 | 0 | 1 | 0 | 0 |
...
```

bdorney (approved):
fine from my side

Issue #240, opened 2019-07-15
[Maintenance/Bug Report] Restrict cx_Oracle to version 5.2 or lower.


bdorney:
## Brief summary of issue

We might need to add a restriction on the allowed versions of `cx_Oracle` due to incompatibilities between newer/older versions of the package.

### Types of issue

- [x] Bug report (report an issue with the code)
- [ ] Feature request (request for change which adds functionality)

## Expected Behavior

`TypeError` shouldn't be generated at runtime.

## Current Behavior

QC7 team reported issue in QC7 machine:

https://mattermost.web.cern.ch/cms-gem-ops/pl/bruym98zg3d8tdczrm856pwpca

@jsturdy traced this to an incompatible version of `cx_Oracle` package:

https://mattermost.web.cern.ch/cms-gem-ops/pl/f3o8tbfawigrmfpa6r94wsn7pc

## Possible Solution (for bugs)

Restrict versions allowed for the `cx_Oracle`.

## Context (for feature requests)

We should not install packages that do not track their dependencies adequately leading to problems at runtime.

## Your Environment

* Version used: a51324ac284f894979aa5f53260411437dcf6068
* Shell used:

jsturdy:
Don't know about "lower", I just know that there is a problem with the version of the oracle client that is pulled in with `7.1.2`

lpetre-ulb:
On the ULB setup, we are using the following packages versions successfully:

``` bash
(pyvenv) [lpetre@gem-daq ~]$ pip list | grep -e Alchemy -e pandas -e Oracle
cx-Oracle                        7.2.0
pandas                           0.24.2
SQLAlchemy                       1.3.6
(pyvenv) [lpetre@gem-daq ~]$ getCalInfoFromDB.py 1 4 4
Open pickled address table if available  /home/lpetre/data/address_table/amc_address_table_top.pickle...
Initializing AMC gem-shelf01-amc04
Length of returned view does not match length of input vfat List
VFATs not found: []
<class 'pandas.core.frame.DataFrame'>
Int64Index: 24 entries, 0 to 23
Data columns (total 10 columns):
vfatN            24 non-null int64
vfat3_ser_num    24 non-null object
vfat3_barcode    24 non-null object
iref             24 non-null int64
adc0m            24 non-null float64
adc1m            24 non-null float64
[...]
    vfatN vfat3_ser_num vfat3_barcode  iref    adc0m    adc1m    adc0b  \
[...]

PR #241, opened 2019-07-31
Resolving Version Mismatch w/cx_Oracle & sqlalchemy (Issue #240) and Providing Tool for Comparing Results from anaUltraLatency.py

bdorney:
## Description

Closes issue #240 al-la suggestion by @jsturdy [here](https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/240#issuecomment-520614525) adding tool for comparing results from `anaUltraLatency.py`.

New tool: `plotLatComp.py` with help menu:

```bash
% plotLatComp.py -h
usage: plotLatComp.py [-h] [-a] [-e EVENTSPERLAT] [-n] [-s] filename

Arguments to supply to plotLatComp.py

positional arguments:
  filename              Tab delimited filename specifying list of scandates to
                        compare. First line understood as column header. First
                        column is chamber_name, second is scandate, and third
                        will be used to fill a TLegend. See documentation for
                        parseListOfScanDatesFile(...) for more details

optional arguments:
  -h, --help            show this help message and exit
  -a, --alphaLabels     Provide this argument if alphanumeric characters exist
                        in the third column of the input file
  -e EVENTSPERLAT, --eventsPerLat EVENTSPERLAT
                        Number of events per latency bin; used to set height
                        of y-axis on all output plots
  -n, --noLeg           Do not draw a TLegend on the output plots
  -s, --savePlots       Make *.png file for all plots that will be saved in
                        the output TFile
```

### Types of changes

Breaking because previously the `$ENV` variable `GEM_ONLINE_DB_CONN` was of the form:

```
export GEM_ONLINE_DB_CONN="oracle://DBUSER:DBPASS@"
```

Where `DBUSER` and `DBPASS` where the username and password respectively.  Now `GEM_ONLINE_DB_CONN` is expected to be:

```
export GEM_ONLINE_DB_CONN="DBUSER/DBPASS@"
```

This is due to the differences in syntax between `cx_Oracle` and `sqlalchemy`.

## Motivation and Context

#240 and we needed a tool to easily compare results from `anaUltraLatency.py`.

## How Has This Been Tested?

Example call of `plotLatComp.py`:

```bash
plotLatComp.py -a -e 100 listOfScanDates_latResults_weirdRateEffect.txt
```

LatScan_DifferentRates

## Checklist:


pkg/setup.cfg
```diff
@@ -67,7 +67,7 @@ release = ___release___
 build_requires = python >= 2.6
 requires = python >= 2.6
            cmsgemos_gempython >= 0.3.1
-           cx_Oracle
+           cx_Oracle <= 5.2.1
```
jsturdy: I don't believe this is a strict requirement, however, it **is** a requirement that couples with the version of `sqlalchemy` that is installed (if `sqlalchemy` is too old, then it requires an older version of `cx_Oracle`)
In other words, the error is not in this codebase, rather in the fact that there isn't a proper restriction in `sqlalchemy` to limit the version of `cx_Oracle` based on the installed version of `sqlalchemy`

bdorney: So you're suggesting the requirement on version number should move to `sqlalchemy`?

jsturdy: No, rather I'm saying that I wouldn't implement this change here at all, because the problem *isn't* with this package... (the brute force ~~solution~~ hack would be to restrict *both* to exactly the versions being used...)

bdorney: What would be your preferred solution then?  From what I understood the problem was an upstream change in `cx_Oracle` and `sqlalchemy` that made an issue for `pandas` `DataFrame::read_sql`.  The change here resolves this problem and restricts the packages to what is natively available in `cc7` which I understood was one of your guiding stars, e.g. do not introduce additional dependencies required on GEM machines which would necessitate new RPM's and dependencies to be built and packaged for `puppet`.

jsturdy: > What would be your preferred solution then?
> > No, rather I'm saying that I wouldn't implement this change here at all

I think I already stated my preference...
Yes, your change here does ~~"fix"~~ mitigate the issue for `cc7`, but in 3-4 months, we'll have a `cc8` build for testing, so my earlier preference to test against different versions (`sqlalchemy` and `cx_Oracle`) to find a robust solution that would work regardless, is my higher preference.  At present, it's trivial to ensure that the `cx_Oracle` version is pegged on the machines being used.
In fact, I just checked and on most machines, the `cx_Oracle` version is `5.2.1-1`, while on two (`gem904qc8dqm` and `gem904daq04`) the version is `7.1-2`

mexanick:
Needs a rebase

pkg/setup.cfg
bdorney: > Yes, your change here does "fix" mitigate the issue for cc7, but in 3-4 months, we'll have a cc8 build for testing, so my earlier preference to test against different versions (sqlalchemy and cx_Oracle) to find a robust solution that would work regardless, is my higher preference. At present, it's trivial to ensure that the cx_Oracle version is pegged on the machines being used.

I would suggest that you open a separate issue to address this in the future and allow this  for the time being as we have limited resources and this mitigation allows work to continue without taking time away from higher priority tasks.

bdorney:
@mexanick this has been `rebase`d

pkg/setup.cfg
jsturdy: How about we just change as per https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/240#issuecomment-520614525 and bypass the need for an explicit versionlock altogether

mexanick: second this ^^

mexanick (dismissed):

pkg/setup.cfg
bdorney: > How about we just change as per [#240 (comment)](https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/240#issuecomment-520614525) and bypass the need for an explicit versionlock altogether

I've implemented this solution and the offending commit was been removed.

bdorney:
I've `rebase`d this and addressed concerns.  Ready for re-review @jsturdy and @mexanick

mexanick (approved):
Cursory

jsturdy (approved):

PR #243, opened 2019-08-13
getDirByAnaType raises KeyError for bad anaType

bdorney:
## Description

Function `getDirByAnaType()` will no longer exit if input argument `anaType` is not in list of accepted types (`ana_config`) but will instead raise a `KeyError` which can be caught as an exception.

### Types of changes

Breaking because instead of exiting it will raise an error.

## Motivation and Context

QOL; allow developers opportunity to catch bad `anaType` and handle this instead of automatically exiting.

## How Has This Been Tested?

Change was trivial.

## Checklist:


jsturdy:
Is there an example where this functionality will be beneficial (catching rather than simply exiting)?

bdorney:
> Is there an example where this functionality will be beneficial (catching rather than simply exiting)?

Off the top of my head no.  But I think in general it might be better coding practice to generate an exception in an API function then exit in an manner that does *not* raise a `SystemExit` exception.  Especially as new developers come in and may not be familiar with the API...it may be difficult for them to trace down what's going on and raising an exception will ensure a stack trace.

utils/anautilities.py
```diff
@@ -572,14 +572,9 @@ def getDataPath():
...
     
-    import os
-
     # Check anaType is understood
     if anaType not in ana_config.keys():
-        print "getDirByAnaType() - Invalid analysis specificed, please select only from the list:"
-        print ana_config.keys()
-        exit(os.EX_USAGE)
-        pass
+        raise KeyError("getDirByAnaType - Invalid anaType argument; please select only from the list: {:s}".format(ana_config.keys()))
```
jsturdy, line 577: Probably useful to print the type that was specified

jsturdy (changes requested):

lpetre-ulb:
Should this PR be merged or closed? The change seems trivial, but the PR require a rebase and a small change.

PR #249, opened 2019-09-12
use the detName branch in analyses

### LINKED ISSUE #273: Feature Request: Add TBranch to output TTree created by sbitRateScanAllLinks(...) to store Detector Serial Number
## Brief summary of issue

Right now the data taken by function [`sbitRateScanAllLinks()`](https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/rpc-playground/utils/scanUtils.py#L378-L566) contains multiple OH's but does not include the detector serial number in the output path.  This makes the risk that some later time this information could be lost.  Request to add a TBranch in the output file that stores this information.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

## Expected Behavior

The `sbitRateScanAllLinks` would need to import `chamber_config` dictionary or take it as an input argument (as [`sbitRateAnalysis`](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/9ce5139a7d9b1a657646e00b8753ea52807286c4/utils/threshAlgos.py#L1139-L1453) does).  This would provide the mapping of geographic address to chamber name.

Then the [`gemSbitRateTreeStructure`](https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/15cf95544f2c2e0d98864ae8a65f166d80f9d245/utils/treeStructure.py#L213-L266) object would need a data member for storing chamber name and an associated `TBranch`.

This would then be filled at in `sbitRateAnalysis` at:

 - VFAT level: [L496-L506](https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/15cf95544f2c2e0d98864ae8a65f166d80f9d245/utils/scanUtils.py#L496-L506)
 - Overall rate: [L525-L535](https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/15cf95544f2c2e0d98864ae8a65f166d80f9d245/utils/scanUtils.py#L525-L535)

## Current Behavior

Detector serial number is not stored in output raw file.

## Context (for feature requests)

See above.

## Your Environment

* Version used: 15cf95544f2c2e0d98864ae8a65f166d80f9d245
* Shell used:

AndrewLevin:
A pull request to `vfatqc-python-scripts` (https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/276) adds a detector serial number branch to the output trees from all of the scans. This pull request uses this new branch in the analyses

## Description

In the analysis of the DAC scans and the sbitRate scans, in which there are multiple links, the `detName` branch is used to create the output directory paths.

In the analyses of the scurve scans, threshold scans, and latency scans, the `detName` branch is copied from the input tree to the output tree.

Ensures backwards compatibility by checking whether there is a `detName` branch in the input tree before trying to access it.

Also, this fixes several bugs which assumed that all 24 VFATs were present in the input.

### Types of changes

## Motivation and Context
This, together with https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/276,  are intended to resolve https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/273

## How Has This Been Tested?
This was tested to analyze an scurve scan, a DAC scan, a threshold scan, a latency scan, and an sbitRate scan.

## Checklist:

### DIFF
--- a/utils/threshAlgos.py
+++ b/utils/threshAlgos.py
@@ -1238,6 +1238,13 @@ def sbitRateAnalysis(chamber_config, rateTree, cutOffRate=0.0, debug=False, outf
     list_bNames.remove('vfatN')
     crateMap = np.unique(rp.tree2array(tree=rateTree,branches=list_bNames))
 
+    #map to go from the ohKey to the detector serial number
+    detNamesMap = {}
+    
+    if hasDetNameBranch:
+        for entry in crateMap:
+            detNamesMap[(entry['shelf'],entry['slot'],entry['link'])] = entry['detName'][0]
+    
     # get nonzero VFATs
     dict_nonzeroVFATs = {}
     for entry in crateMap:
@@ -1490,4 +1497,28 @@ def sbitRateAnalysis(chamber_config, rateTree, cutOffRate=0.0, debug=False, outf
         outputFiles[ohKey].Close()
         pass
 
-    return (perchannel, dict_dacValsBelowCutOff)
+    for ohKey,innerDictByVFATKey in dict_dacValsBelowCutOff["THR_ARM_DAC"].iteritems():
+        if hasDetNameBranch:
+            detName = detNamesMap[ohKey]
+        else:
+            detName = chamber_config[ohKey]                            
+
+        from gempython.utils.gemlogger import printGreen    
+            
+        if scandate == 'noscandate':
+            vfatConfg = open("{0}/{1}/vfatConfig.txt".format(elogPath,detName),'w')
+            printGreen("Output Data for {0} can be found in:\n\t{1}/{0}\n".format(detName,elogPath))
+        else:    
+            if perchannel:
+                strDirName = getDirByAnaType("sbitRatech", detName)
+            else:
+                strDirName = getDirByAnaType("sbitRateor", detName)
+            vfatConfg = open("{0}/{1}/vfatConfig.txt".format(strDirName,scandate),'w')
+            printGreen("Output Data for {0} can be found in:\n\t{1}/{2}\n".format(detName,strDirName,scandate))
+            
+        vfatConfg.write("vfatN/I:vt1/I:trimRange/I\n")
+        for vfat,armDACVal in innerDictByVFATKey.iteritems():
+            vfatConfg.write('%i\t%i\t%i\n'%(vfat, armDACVal,0))
+        vfatConfg.close()    
+        
+    return 
--- a/utils/threshAlgos.py
+++ b/utils/threshAlgos.py
@@ -134,9 +134,11 @@ def anaUltraThreshold(args,thrFilename,GEBtype="short",outputDir=None,fileScurve
     listOfBranches = thrTree.GetListOfBranches()
     if 'vfatID' in listOfBranches:
         array_chipID = np.unique(rp.tree2array(thrTree, branches = [ 'vfatID','vfatN' ] ))
-        array_chipID = np.sort(array_chipID, order='vfatN')['vfatID']
+        dict_chipID = {}
+        for entry in array_chipID:
+            dict_chipID[entry['vfatN']]=entry['vfatID']
     else:
-        array_chipID = np.zeros(24)
+        dict_chipID = { vfat:0 for vfat in range(24) }
 
     r.TH1.SetDefaultSumw2(False)
     r.gROOT.SetBatch(True)
@@ -146,7 +148,11 @@ def anaUltraThreshold(args,thrFilename,GEBtype="short",outputDir=None,fileScurve
     dict_hMaxThrDAC_NoOutlier = {}
     dict_chanMaxThrDAC = {}
     for vfat in range(0,24):
-        chipID = array_chipID[vfat]
+        if vfat not in dict_chipID:
+            dict_h2D_thrDAC[vfat] = r.TH2D()
+            continue
+        
+        chipID = dict_chipID[vfat]
         dict_h2D_thrDAC[vfat] = r.TH2D(
                 'h_thrDAC_vs_ROBstr_VFAT{0}'.format(vfat),
                 'VFAT{0} chipID {1};{2};{3} [DAC units]'.format(vfat,chipID,stripChanOrPinName[1],dacName),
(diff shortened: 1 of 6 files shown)


AndrewLevin:
I found another place where `chamber_config` was used, which is updated now.

AndrewLevin:
In the analysis of the sbitThresh scans, the writing of the vfatConfig was duplicated in `anaSBitThresh.py` and in `ana_scans.py`:

https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/ana_scans.py#L117-L137
https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/anaSBitThresh.py#L100-L121

I don't see any reason for this duplication, so I have moved it to the `sbitRateAnalysis` function in `threshAlgos.py`. 

This is related to the detector name branch because the vfatConfig is written to a directory that includes the detector name.

It is also relevant that in many places we are indexing dictionaries based on the OH key. For example:

https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/utils/threshAlgos.py#L1218
https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/utils/threshAlgos.py#L1259

and many other places. We may want to change the indexing to be based on the detector name instead. I thought that the OH key is more guaranteed to be unique, so I didn't make this change here.

jsturdy:
> In the analysis of the sbitThresh scans, the writing of the vfatConfig was duplicated in `anaSBitThresh.py` and in `ana_scans.py`:
> 
> https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/ana_scans.py#L117-L137
> https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/anaSBitThresh.py#L100-L121
> 
> I don't see any reason for this duplication, so I have moved it to the `sbitRateAnalysis` function in `threshAlgos.py`.
> 

OK, seems OK to me, I have to admit that I'm not familiar with the workings of this codebase enough to understand if there was some historical reason for this.
Provided you've tested (at least the likely) use cases and not encountered any issues, then your change is probably good for the long term.

> This is related to the detector name branch because the vfatConfig is written to a directory that includes the detector name.
> 
> It is also relevant that in many places we are indexing dictionaries based on the OH key. For example:
> 
> https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/utils/threshAlgos.py#L1218
> https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/utils/threshAlgos.py#L1259
> 
> and many other places. We may want to change the indexing to be based on the detector name instead. I thought that the OH key is more guaranteed to be unique, so I didn't make this change here.

If you'd like, feel free to also do this change and see if it causes any hiccups (may as well put it in a separate PR so we can get this one through), again, I have no idea if making this change could have unintended side effects

jsturdy (commented):
Curiosity, do you know why `ana_scans` for the `sbitRateAnalysis` doesn't call/execute `anaSBitThresh.py`?

utils/threshAlgos.py
```diff
@@ -169,19 +175,18 @@ def anaUltraThreshold(args,thrFilename,GEBtype="short",outputDir=None,fileScurve
...
-                array_chipID[event.vfatN] = 0
-
         stripPinOrChan = dict_vfatChanLUT[event.vfatN][stripChanOrPinType][event.vfatCH]
         dict_h2D_thrDAC[event.vfatN].Fill(stripPinOrChan,event.vth1,event.Nhits)
         pass
 
-    # Make output TTree
+    # Make output TFile and make output TTree
     from array import array
+    outFile = r.TFile("{0}/{1}".format(outputDir,args.outfilename),"RECREATE")
```
jsturdy, line 184: Why is this now needed?

utils/threshAlgos.py
```diff
@@ -1187,8 +1209,16 @@ def sbitRateAnalysis(chamber_config, rateTree, cutOffRate=0.0, debug=False, outf
...
 
+    #for backwards compatibility, handle input trees that do not have a detName branch by finding the detName using the chamber_config
+    hasDetNameBranch = False
+    if 'detName' in rateTree.GetListOfBranches():
+        hasDetNameBranch = True
+
+    list_bNames = ['dacValX','link','nameX','rate','shelf','slot','vfatCH','vfatN','detName']
+    if not hasDetNameBranch:
+        list_bNames.remove('detName')
+
```
jsturdy, line 1223: Can this block be abstracted in the several places it is used?

utils/threshAlgos.py
```diff
@@ -1225,15 +1262,18 @@ def sbitRateAnalysis(chamber_config, rateTree, cutOffRate=0.0, debug=False, outf
     from gempython.utils.wrappers import runCommand
     from gempython.gemplotting.utils.anautilities import getDirByAnaType
     for entry in crateMap:
-        ohKey = (entry['shelf'],entry['slot'],entry['link'])
+        if hasDetNameBranch:
+            detName = entry['detName'][0]
+        else:
+            detName = chamber_config[(entry['shelf'],entry['slot'],entry['link'])]
```
jsturdy: Likewise for this block, used many times, usually the same syntax

jsturdy (commented):
Couple of clarifications and question on potential streamlining.

utils/threshAlgos.py
AndrewLevin: I have added a function to do this now.

utils/threshAlgos.py
AndrewLevin: I think this is done only twice and in separate files.

utils/threshAlgos.py
AndrewLevin: It was moved from the bottom of the script to here in order to avoid errors:

```
Error in <TTree::Fill>: Failed filling branch:thrAnaTree.detName, nbytes=-1, entry=1680
 This error is symptomatic of a Tree created as a memory-resident Tree
 Instead of doing:
    TTree *T = new TTree(...)
    TFile *f = new TFile(...)
 you should do:
    TFile *f = new TFile(...)
    TTree *T = new TTree(...)
```

AndrewLevin:
> Curiosity, do you know why `ana_scans` for the `sbitRateAnalysis` doesn't call/execute `anaSBitThresh.py`?

To me it seems like poor style to have python call bash which then calls python again, so maybe the real question is why does `anaSBitThresh.py` exist. 

`run_scans.py` does call the other `.py` scripts while `ana_scans.py` imports functions. I guess using the importable functions are more useful for multithreading and also we are importing and running some of these analyses in `testConnectivity.py`.

jsturdy:
> > Curiosity, do you know why `ana_scans` for the `sbitRateAnalysis` doesn't call/execute `anaSBitThresh.py`?
> 
> To me it seems like poor style to have python call bash which then calls python again, so maybe the real question is why does `anaSBitThresh.py` exist.
> 
> `run_scans.py` does call the other `.py` scripts while `ana_scans.py` imports functions. I guess using the importable functions are more useful for multithreading and also we are importing and running some of these analyses in `testConnectivity.py`.

I know there was effort put into making more things `import`able, and I don't know where this ended, but perhaps it more done in the plotting side and not yet in the scan side...

utils/threshAlgos.py
jsturdy: OK, for now I guess it's fine, but I'd consider consolidating this in the future

jsturdy (changes requested):
Needs `rebase`, then can be approved

AndrewLevin:
It is rebased now.

jsturdy (approved):
:+1:

PR #251, opened 2019-10-10
rethrow exceptions from cx_Oracle

### LINKED ISSUE #250: [bug report] Issue with multiprocessing
## Brief summary of issue
It seems that between `` and `` a new behaviour was introduced, either in a return type or in how the multiprocessing is done.
Need to understand which commit brought the regression, and if it is a bug, or an expected change that needed changes elsewhere.

### Types of issue
- [x] Bug report (report an issue with the code)

## Expected Behavior
No errors

## Current Behavior
```
ana_scans.py scurve -s 2019.10.01.19.41 -c --chamberConfig --medium
Launching scurve analysis processes, this may take some time, please be patient
Exception in thread Thread-3:
Traceback (most recent call last):
  File "/usr/lib64/python2.7/threading.py", line 812, in __bootstrap_inner
    self.run()
  File "/usr/lib64/python2.7/threading.py", line 765, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/usr/lib64/python2.7/multiprocessing/pool.py", line 376, in _handle_results
    task = get()
UnpicklingError: NEWOBJ class argument has NULL tp_new
```

### Steps to Reproduce (for bugs)
1. run `ana_scans.py scurve -s 2019.10.01.19.41 -c --chamberConfig --medium` on data taken at QC8

## Possible Solution (for bugs)
These may be useful (same error message, and components we are using)
* https://stackoverflow.com/questions/1953261/weird-unpickling-error-when-using-multiprocessing
* https://stackoverflow.com/questions/11442191/parallelizing-a-numpy-vector-operation
* https://stackoverflow.com/questions/23572477/use-cx-oracle-and-multiprocessing-to-query-data-concurrently

## Your Environment
* Version used: `0e47e29`

AndrewLevin:
`cx_Oracle` throws exceptions objects that cannot be pickled, which is a problem for our multithreading framework, so this pull requests rethrows exceptions that cannot be pickled.

## Description
In the function `getVFAT3CalInfo`, a try-catch block is added around the call to `getVFAT3ProdSumView`. When an exception is caught, it tries to pickle the exception, and checks whether a `TypeError` exception is throw by the pickler. If there is no `TypeError`, the original exception is rethrown. If there is a `TypeError`, then a new exception is created which records the type and the message of the original exception.

### Types of changes

## Motivation and Context
Partially resolves https://github.com/cms-gem-daq-project/gem-plotting-tools/issues/250. The underlying cause of the exception reported in that issue is not understood.

## How Has This Been Tested?
Yes:

```
Launching scurve analysis processes, this may take some time, please be patient
Caught <type 'exceptions.Exception'>: This is a rethrown exception. The original exception type was: <class 'cx_Oracle.DatabaseError'>. The original exception message was: ORA-01017: invalid username/password; logon denied
, terminating workers
```

## Checklist:

### DIFF
--- a/utils/dbutils.py
+++ b/utils/dbutils.py
@@ -102,8 +102,21 @@ def getVFAT3CalInfo(vfatList, debug=False):
     debug       - Prints additional info if true
     """
 
-    df_vfatCalInfo = getVFAT3ProdSumView(vfatList, debug)
-
+    #When using multithreading, the threads pass information back in pickled format. Some exceptions in cx_Oracle cannot be pickled.
+    #Thus, we check here whether the exception can be pickled, and if not rethrow it as an exception that can be.
+    try:
+        df_vfatCalInfo = getVFAT3ProdSumView(vfatList, debug)
+    except Exception as err:
+        import pickle
+        try:
+            pickle.dumps(err)
+        except TypeError:
+            raise Exception("This is a rethrown exception. The original exception type was: "+str(type(err)) +". The original exception message was: "+str(err.message))
+        else:
+            raise err
+            
+            
+        
     return df_vfatCalInfo[['vfatN','vfat3_ser_num', 'vfat3_barcode', 'iref', 'adc0m', 'adc1m', 'adc0b', 'adc1b', 'cal_dacm', 'cal_dacb']]
 
 def getVFAT3ConfView(vfatList, debug=False):


jsturdy (approved):
Are there any other places that we need to add protection for DB exceptions?
Might be safer to find the places where the `connect` method is actually called, and just protect those

AndrewLevin:
> Are there any other places that we need to add protection for DB exceptions?
> Might be safer to find the places where the `connect` method is actually called, and just protect those

I think the `connect` method is only called in one place in `dbutils`, but I think similar exceptions could be throw in other interactions with the db, such as querying the db after we connect.

Issue #226, opened 2019-06-25
Feature Request: GE2/1 Compatibility for Plotting


sbutalla:
## Brief summary of issue

Currently, the `anaUltraScurve.py` and related scripts are compatible with GE1/1 only. The GE2/1 has only 12 VFATs per module, while the GE1/1 has 24, so half of the plots are not populated and can therefore be discarded when plotting the final result for GE2/1 data.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

## Expected Behavior

In order to change this, import the hardware constants from `gempython.tools`.
The user can add an option specifying the `--gemType`, being either the GE11 or GE21 (ME0 to be implemented later), as well as the `detType` argument (i.e., long/short for GE11, module number for GE21). After this, the number of VFATs would be assigned to a variable which is used to set the maximum of each of the loops that controls plotting for the individual VFATs.

## Current Behavior

The analysis script will create plots for the correct number of VFATs depending on the GEM detector, which will be provided as an option. I.e., for the GE2/1 plot only 12 VFATs per module, and for the GE1/1 plot 24 VFATs.

## Possible Solution (for bugs)

Retrieve the hardware constants from `hw_constants.py`, add option for `gemVariant` and use the `vfatsPerGemVariant` to assign the number of VFATs for the proper `gemType`, and set the default value of `gemType` to GE11 to maintain backwards compatibility:
```python
from gempython.tools.hw_constants import gemVariants, vfatsPerGemVariant
parser.add_option("--gemType", dest="gemType", type="string", default="ge11", help="String that defines the GEM variant, available from the list: {0}".format(gemVariants.keys()))
parser.add_option("--detType",type="string",help="Detector type within gemType. If gemType is 'ge11' then this should be from list {0}; if gemType is 'ge21' then this should be from list {1}; and if type is 'me0' then this should be from the list {2}".format(gemVariants['ge11'],gemVariants['ge21'],gemVariants['me0']),default=None)

#Assign number of VFATs for gemType
    if args.gemType == "ge11":
        vfatNum = {0}.format(vfatsPerGemVariant)
    elif args.gemType == "ge21":
        vfatNum = {0}.format(vfatsPerGemVariant)
    else:
        print("me0 gemType not currently implemented, exiting.")
        return
```
For every loop that is plotting distributions, use the maximum range of the loop as `vfatNum`, i.e.,

[...]

PR #252, opened 2019-10-14
Ge21 generalizing

dteague:
PR for changing the gem plotting code to be compatible with GE21 plotting

## Description
The gem-plotting code is hardcoded with many magic numbers that are specific to the GE11, so with changes in the tree structure to add gemtype and detectortype, the plotting code can be expanded to allow for GE21 plotting that is determined on the fly by this gemType variable.

The bulk of the code is removing instances of 24, 8, 3, and combinations of these to more generic variables, ie nVFATS, maxiEta, maxiPhi. These variables are derived from `chamber_maxiEtaiPhiPair` in  `./mapping/chamberInfo.py` and from `vfatsPerGemVariant`  in `gempython.tools.hw_constants` All of the mapping/chamberInfo.py variables have been turned into dictionaries that require the gemType to get the correct variables, ie

 ```chamber_vfatPos2iEta   =>    chamber_vfatPos2iEta["ge11"]```
etc.

Since a large code change is the actual canvas creation code, the largest change is to canvas making. The `makeAxBCanvas` has been removed in place of merging this functionality into the `saveSummaryCanvas` to make a new function called `getSummaryCanvas`.  This means the function now always returns a canvas created and there is a flag to decide to save the canvas or not. ie function def is now

```def getSummaryCanvas(dictSummary, dictSummaryPanPin2=None, name='Summary', trimPt=None, drawOpt="colz", gemType="ge11", write2Disk=False)
canv = make3x8Canvas(...)    =>    canv = getSummaryCanvas(..., gemType="ge11")
saveSummaryCanvas(...)    =>    getSummaryCanvas(..., gemType="ge11", write2Disk=True)
```

Also, because the `makeAxBCanvas` function also added histograms to an existing canvas, this functionality has be put to a helper function `addPlotToCanvas` which hopefully will make the code more readable as well. So the change becomes

```
canv = make3x8Canvas(initialContent=hist1, ...)
canv = make3x8Canvas(initialContent=hist2, canv=canv, ...)

=>

canv = getSummaryCanvas(hist1, ..., gemType="ge11")
canv = addPlotToCanvas(canv, hist2, gemType="ge11")
```

Last, many functions include now a gemType variable that needs to be supplied to give the right gemType. The default is "ge11", so the code remains backwards compatible for all cases where the gemType isn't actually supplied.

### Types of changes

## Motivation and Context

This change has mainly be motivated to be backwards compatible and to default to ge11 in all unknown cases. Since ge21 is needing plots specifically made for it (and not ones derived from ge11 algorithms), having code that can flexibly handle ge21 and in the future, me0 is a must.

For further information on this, look at issue #226 

## How Has This Been Tested?
For testing, the code was run over ge11 and ge21 files with the internal gemType flag flipped (ge21 files didn't have that tree in it). 

NOTE: currently, only the main run files (ie files in the top directory) are changed. Files in the `./macros` area have not been touched. Here is a list of what has been done.

- `anaDACScan.py`, `anaUltraScurve.py`, `anaSBitThresh.py`, `anaUltraLatency.py` - work for both
- `anaUltraThreshold.py`, `anaSBitMonitor.py` - work for ge11, could not fine ge21 file
- `anaSBitReadout.py`, `anaXDAQLatency.py` - could not find input file for either

## Checklist:

### DIFF
--- a/anaSBitMonitor.py
+++ b/anaSBitMonitor.py
@@ -20,7 +20,6 @@ if __name__ == '__main__':
     filename = args.infilename[:-5]
     os.system("mkdir " + args.infilename[:-5])
 
-    print filename
     outfilename = args.outfilename
 
     import ROOT as r
--- a/anaSBitMonitor.py
+++ b/anaSBitMonitor.py
@@ -49,7 +49,7 @@ if __name__ == '__main__':
     ##### END
     from gempython.tools.hw_constants import vfatsPerGemVariant
     nVFATS = vfatsPerGemVariant[gemType]
-    from gempython.gemplotting.mapping.chamberInfo import CHANNELS_PER_VFATS as maxChans
+    from gempython.gemplotting.mapping.chamberInfo import CHANNELS_PER_VFAT as maxChans
     
     print('Initializing histograms')
     from gempython.utils.nesteddict import nesteddict as ndict, flatten
@@ -491,7 +491,7 @@ if __name__ == '__main__':
     print("| :---: | :------: | :-------: | :----------: |")
     fileMisMappedSbits.write("| :---: | :------: | :-------: | :----------: |\n")
     for idx in range(0,len(wrongSBITMapping)):
-        print("| {0} | {1} | {2} | {3} |".format(
+        print("| {0:5d} | {1:8d} | {2:9d} | {3:12d} |".format(
             wrongSBITMapping[idx]['vfatN'],
             wrongSBITMapping[idx]['vfatSBIT'],
             wrongSBITMapping[idx]['sbitSize'],
@@ -519,7 +519,7 @@ if __name__ == '__main__':
     print("| :--------: | :----------: | :----------: |")
     fileMisMappedVFATs.write("| :--------: | :----------: | :----------: |\n")
     for idx in range(0,len(wrongVFATMapping)):
-        print("| {0} | {1} | {2} |".format(
+        print("| {0:10d} | {1:12d} | {2:12d} |".format(
             wrongVFATMapping[idx]['vfatN'],
             wrongVFATMapping[idx]['vfatObs'],
             wrongVFATMapping[idx]['N_Mismatches']))
--- a/anaSBitMonitor.py
+++ b/anaSBitMonitor.py
@@ -17,8 +17,8 @@ if __name__ == '__main__':
     parser.set_defaults(outfilename="SBitMonitorData.root")
     args = parser.parse_args()
     
-    filename = args.filename[:-5]
-    os.system("mkdir " + args.filename[:-5])
+    filename = args.infilename[:-5]
+    os.system("mkdir " + args.infilename[:-5])
 
     print filename
     outfilename = args.outfilename
@@ -29,7 +29,7 @@ if __name__ == '__main__':
     outF = r.TFile(filename+'/'+outfilename, 'recreate')
     inF = r.TFile(filename+'.root')
 
-    gemType = "ge11"
+    
     
     # Determine the rates scanned
     print('Determining rates tested')
@@ -41,7 +41,14 @@ if __name__ == '__main__':
     calEnableValues = np.unique(initInfo['calEnable'])
     isValidValues = np.unique(initInfo['isValid'])
     ratesUsed = np.unique(initInfo['ratePulsed'])
-
+    ##### NEED TO FIX
+    #### VERY TEMPORARY
+    from gempython.gemplotting.mapping.chamberInfo import gemTypeMapping
+    print rp.tree2array(tree=inF.sbitDataTree, branches ='gemType')
+    gemType = gemTypeMapping[rp.tree2array(tree=inF.sbitDataTree, branches =[ 'gemType' ] )[0][0]]
+    print gemType
+    ##### END
+    
     print('Initializing histograms')
     from gempython.utils.nesteddict import nesteddict as ndict, flatten
 
@@ -302,9 +309,9 @@ if __name__ == '__main__':
         # All Rates
         rateCanvas = getSummaryCanvas(dict_g_rateObsCTP7VsRatePulsed[isValid],
                                       name="rateObservedVsRatePulsed_{0}".format(strValidity),
-                                      drawOpt="APE1")
-        rateCanvas = addPlotToCanvas(rateCanvas, dict_g_rateObsFPGAVsRatePulsed[isValid], drawOpt="PE1")
-        rateCanvas = addPlotToCanvas(rateCanvas, dict_g_rateObsVFATVsRatePulsed[isValid], drawOpt="PE1")
+                                      drawOpt="APE1", gemType=gemType)
+        rateCanvas = addPlotToCanvas(rateCanvas, dict_g_rateObsFPGAVsRatePulsed[isValid], drawOpt="PE1", gemType=gemType)
+        rateCanvas = addPlotToCanvas(rateCanvas, dict_g_rateObsVFATVsRatePulsed[isValid], drawOpt="PE1", gemType=gemType)
         
         # Make the legend
         rateLeg = r.TLegend(0.5,0.5,0.9,0.9)
@@ -328,11 +335,11 @@ if __name__ == '__main__':
         rateCanvas.SaveAs("{0}/rateObservedVsRatePulsed_{1}.png".format(filename,strValidity))
 
         # CalDisable rate 0 case
-        getSummaryCanvas(dict_h_sbitMulti[isValid][0][0], name="{0}/sbitMulti_{1}_calDisabled_0Hz.png".format(filename,strValidity), drawOpt="", write2Disk=True)
-        getSummaryCanvas(dict_h_sbitSize[isValid][0][0], name="{0}/sbitSize_{1}_calDisabled_0Hz.png".format(filename,strValidity), drawOpt="", write2Disk=True)
+        getSummaryCanvas(dict_h_sbitMulti[isValid][0][0], name="{0}/sbitMulti_{1}_calDisabled_0Hz.png".format(filename,strValidity), drawOpt="", gemType=gemType, write2Disk=True)
+        getSummaryCanvas(dict_h_sbitSize[isValid][0][0], name="{0}/sbitSize_{1}_calDisabled_0Hz.png".format(filename,strValidity), drawOpt="", gemType=gemType, write2Disk=True)
 
-        getSummaryCanvas(dict_h_sbitObsVsChanPulsed[isValid][0][0], name="{0}/sbitObsVsChanUnmasked_{1}_calDisabled_0Hz.png".format(filename,strValidity), drawOpt="COLZ", write2Disk=True)
-        getSummaryCanvas(dict_h_sbitMultiVsSbitSize[isValid][0][0], name="{0}/sbitMultiVsSbitSize_{1}_calDisabled_0Hz.png".format(filename,strValidity), drawOpt="COLZ", write2Disk=True)
+        getSummaryCanvas(dict_h_sbitObsVsChanPulsed[isValid][0][0], name="{0}/sbitObsVsChanUnmasked_{1}_calDisabled_0Hz.png".format(filename,strValidity), drawOpt="COLZ", gemType=gemType, write2Disk=True)
+        getSummaryCanvas(dict_h_sbitMultiVsSbitSize[isValid][0][0], name="{0}/sbitMultiVsSbitSize_{1}_calDisabled_0Hz.png".format(filename,strValidity), drawOpt="COLZ", gemType=gemType, write2Disk=True)
         
         for idx,rate in enumerate(ratesUsed):
             # Sum over all rates
@@ -375,11 +382,11 @@ if __name__ == '__main__':
                         dict_h_sbitObsVsChanPulsed[isValid][1][-1][vfat].Add(dict_h_sbitObsVsChanPulsed[isValid][1][rate][vfat])
                         dict_h_sbitMultiVsSbitSize[isValid][1][-1][vfat].Add(dict_h_sbitMultiVsSbitSize[isValid][1][rate][vfat])
         
-                getSummaryCanvas(dict_h_sbitMulti[isValid][1][-1], name="{0}/sbitMulti_{1}_calEnabled_SumOfAllRates.png".format(filename,strValidity), drawOpt="", write2Disk=True)
-                getSummaryCanvas(dict_h_sbitSize[isValid][1][-1], name="{0}/sbitSize_{1}_calEnabled_SumOfAllRates.png".format(filename,strValidity), drawOpt="", write2Disk=True)
+                getSummaryCanvas(dict_h_sbitMulti[isValid][1][-1], name="{0}/sbitMulti_{1}_calEnabled_SumOfAllRates.png".format(filename,strValidity), drawOpt="", gemType=gemType, write2Disk=True)
+                getSummaryCanvas(dict_h_sbitSize[isValid][1][-1], name="{0}/sbitSize_{1}_calEnabled_SumOfAllRates.png".format(filename,strValidity), drawOpt="", gemType=gemType, write2Disk=True)
 
-                getSummaryCanvas(dict_h_sbitObsVsChanPulsed[isValid][1][-1], name="{0}/sbitObsVsChanPulsed_{1}_calEnabled_SumOfAllRates.png".format(filename,strValidity), drawOpt="COLZ", write2Disk=True)
-                getSummaryCanvas(dict_h_sbitMultiVsSbitSize[isValid][1][-1], name="{0}/sbitMultiVsSbitSize_{1}_calEnabled_SumOfAllRates.png".format(filename,strValidity), drawOpt="COLZ", write2Disk=True)
+                getSummaryCanvas(dict_h_sbitObsVsChanPulsed[isValid][1][-1], name="{0}/sbitObsVsChanPulsed_{1}_calEnabled_SumOfAllRates.png".format(filename,strValidity), drawOpt="COLZ", gemType=gemType, write2Disk=True)
+                getSummaryCanvas(dict_h_sbitMultiVsSbitSize[isValid][1][-1], name="{0}/sbitMultiVsSbitSize_{1}_calEnabled_SumOfAllRates.png".format(filename,strValidity), drawOpt="COLZ", gemType=gemType, write2Disk=True)
 
     print("Storing TObjects in output TFile")
     # Per VFAT Plots
--- a/anaSBitMonitor.py
+++ b/anaSBitMonitor.py
@@ -39,8 +39,7 @@ if __name__ == '__main__':
     isValidValues = np.unique(initInfo['isValid'])
     ratesUsed = np.unique(initInfo['ratePulsed'])
 
-    ##### NEED TO FIX
-    #### VERY TEMPORARY
+    ##### FIXME
     from gempython.gemplotting.mapping.chamberInfo import gemTypeMapping
     if 'gemType' not in inF.sbitDataTree.GetListOfBranches():
         gemType = "ge11"
--- a/anaSBitMonitor.py
+++ b/anaSBitMonitor.py
@@ -41,14 +41,17 @@ if __name__ == '__main__':
     calEnableValues = np.unique(initInfo['calEnable'])
     isValidValues = np.unique(initInfo['isValid'])
     ratesUsed = np.unique(initInfo['ratePulsed'])
+
     ##### NEED TO FIX
     #### VERY TEMPORARY
     from gempython.gemplotting.mapping.chamberInfo import gemTypeMapping
-    print rp.tree2array(tree=inF.sbitDataTree, branches ='gemType')
-    gemType = gemTypeMapping[rp.tree2array(tree=inF.sbitDataTree, branches =[ 'gemType' ] )[0][0]]
+    if 'gemType' not in inF.sbitDataTree.GetListOfBranches():
+        gemType = "ge21"
+    else:
+        gemType = gemTypeMapping[rp.tree2array(tree=inF.sbitDataTree, branches =[ 'gemType' ] )[0][0]]
     print gemType
     ##### END
-    
+
     print('Initializing histograms')
     from gempython.utils.nesteddict import nesteddict as ndict, flatten
 
--- a/anaSBitMonitor.py
+++ b/anaSBitMonitor.py
@@ -142,7 +142,7 @@ if __name__ == '__main__':
                         rateMap[rate]=binX+1 # Used later to translate rate to bin position
                     dict_h_chanVsRatePulsed_ZRateObs[isValid][calEnable][vfat].GetXaxis().SetBinLabel(binX+1,formatSciNotation(str(rate)))
 
-                    # VFAT level
+                    # VFAT levpel
                     if (calEnable or (not calEnable and rate == 0.0)):
                         # 1D Obs
                         dict_h_sbitMulti[isValid][calEnable][rate][vfat] = r.TH1F(
@@ -289,7 +289,7 @@ if __name__ == '__main__':
                             dict_h_sbitMultiVsSbitSize[isValid][calEnable][rate][vfat].Fill(size,multi)
     
     print("Making summary plots")
-    from gempython.gemplotting.utils.anautilities import make3x8Canvas, getSummary
+    from gempython.gemplotting.utils.anautilities import addPlotToCanvas, getSummaryCanvas
     for isValid in isValidValues:
         if not isValid and not args.checkInvalid:
             continue
@@ -300,17 +300,11 @@ if __name__ == '__main__':
             strValidity="invalidSbits"
 
         # All Rates
-        rateCanvas = make3x8Canvas(
-                name="rateObservedVsRatePulsed_{0}".format(strValidity),
-                initialContent=dict_g_rateObsCTP7VsRatePulsed[isValid],
-                initialDrawOpt="APE1",
-                secondaryContent=dict_g_rateObsFPGAVsRatePulsed[isValid],
-                secondaryDrawOpt="PE1")
-        rateCanvas = make3x8Canvas(
-                name="",
-                secondaryContent=dict_g_rateObsVFATVsRatePulsed[isValid],
-                secondaryDrawOpt="PE1",
-                canv=rateCanvas)
+        rateCanvas = getSummaryCanvas(dict_g_rateObsCTP7VsRatePulsed[isValid],
+                                      name="rateObservedVsRatePulsed_{0}".format(strValidity),
+                                      drawOpt="APE1")
+        rateCanvas = addPlotToCanvas(rateCanvas, dict_g_rateObsFPGAVsRatePulsed[isValid], drawOpt="PE1")
+        rateCanvas = addPlotToCanvas(rateCanvas, dict_g_rateObsVFATVsRatePulsed[isValid], drawOpt="PE1")
         
         # Make the legend
         rateLeg = r.TLegend(0.5,0.5,0.9,0.9)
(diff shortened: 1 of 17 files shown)


anaSBitMonitor.py
```diff
@@ -40,6 +42,16 @@
     isValidValues = np.unique(initInfo['isValid'])
     ratesUsed = np.unique(initInfo['ratePulsed'])
 
+    ##### NEED TO FIX
+    #### VERY TEMPORARY
+    from gempython.gemplotting.mapping.chamberInfo import gemTypeMapping
+    if 'gemType' not in inF.sbitDataTree.GetListOfBranches():
+        gemType = "ge21"
```
jsturdy: What do `NEED TO FIX` and `VERY TEMPORARY` mean in the context of this PR?
Also, here it appears that a default behaviour is going to be `GE2/1` which should not be the case (all changes should be completely transparent to the current operations)

anaSBitMonitor.py
```diff
@@ -88,10 +102,10 @@
                 dict_h_vfatObsVsVfatPulsed[isValid][calEnable][rate] = r.TH2F(
                         "h_vfatObservedVsVfatPulsed_{0}_{1}Hz".format(postScript,int(rate)),
                         "Summmary - Rate {0} Hz;VFAT {1};VFAT Observed".format(int(rate),strPulsedOrUnmasked),
-                        24,-0.5,23.5,24,-0.5,23.5)
+                        vfatsPerGemVariant[gemType],-0.5,vfatsPerGemVariant[gemType]-0.5, vfatsPerGemVariant[gemType],-0.5,vfatsPerGemVariant[gemType]-0.5)
```
jsturdy: probably, this lookup can be moved outside the loop, no?

anaSBitMonitor.py
```diff
@@ -138,7 +152,7 @@
                         rateMap[rate]=binX+1 # Used later to translate rate to bin position
                     dict_h_chanVsRatePulsed_ZRateObs[isValid][calEnable][vfat].GetXaxis().SetBinLabel(binX+1,formatSciNotation(str(rate)))
 
-                    # VFAT level
+                    # VFAT levpel
```
jsturdy: typo

anaSBitReadout.py
```diff
@@ -241,9 +245,9 @@
     ieta_h_sbitSize = ndict()
     ieta_h_delay = ndict()
 
-    for ieta in range(1, 9):
-        ieta_h_strip[ieta] = r.TH1F("h_ieta{0}_strips_vs_hit".format(ieta), "i#eta = {0} | i#phi (1,2,3)".format(ieta), 384, 0., 384.)
-        ieta_h_ch[ieta] = r.TH1F("h_ieta{0}_chan_vs_hit".format(ieta), "i#eta = {0} | i#phi (1,2,3)".format(ieta), 384, 0., 384.)
+    for ieta in range(1, maxiEta+1):
+        ieta_h_strip[ieta] = r.TH1F("h_ieta{0}_strips_vs_hit".format(ieta), "i#eta = {0} | i#phi (1,2,3)".format(ieta), 128*maxiPhi, 0., 128.*maxiPhi)
```
jsturdy: Probably good to have a `CHANNELS_PER_VFAT` constant defined somewhere, to replace this repetition of the magic number `128` (where `128` refers to the number of VFAT channels...)

anaSBitReadout.py
```diff
@@ -366,57 +371,46 @@
     # Summaries Canvas
     #
     # make3x8Canvas
```
jsturdy: you've changed the semantics, i'd change/remove the comment

anaSBitReadout.py
```diff
@@ -425,7 +419,7 @@
         vfat_h_sbitSize[vfat].Write()
 
     ietaDir.cd()
-    for ieta in range(1, 9):
+    for ieta in range(1, maxiEta):
```
jsturdy: this won't reproduce the same behaviour `maxiEta != 9`
possibly more readable would be `range(maxiEta):` and in the loop the value is used as `ieta+1`

anaXDAQLatency.py
```diff
@@ -45,6 +45,19 @@
...
 
+        ##### NEED TO FIX
+    #### VERY TEMPORARY
+    from gempython.gemplotting.mapping.chamberInfo import gemTypeMapping
+    if 'gemType' not in inFile.latTree.GetListOfBranches():
+        gemType = "ge11"
+    else:
+        gemType = gemTypeMapping[rp.tree2array(tree=inFile.latTree, branches =[ 'gemType' ] )[0][0]]
+    print gemType
+    ##### END
```
jsturdy, line 55: again, what does this comment mean in the context of this PR?

anaXDAQLatency.py
```diff
@@ -74,7 +87,7 @@
     # Make nested containers
     from gempython.utils.nesteddict import nesteddict as ndict
     baseDir = ndict() # baseDir[slot][oh] -> string
-    vfatDirs = ["VFAT-%d"%x for x in range(24)]
+    vfatDirs = ["VFAT-%d"%x for x in range(vfatsPerGemVariant[gemType])]
```
jsturdy: two things:
1) as above, it's probably best to set this `vfatsPerGemVariant` lookup somewhere outside of all the loops, so it's only done once.
1) i would *really* love to get rid of these poorly named `TDirectories` (using `-` in the name breaks a lot of interactive stuff (but this change can't be made here as it would break backwards compatibility)

mapping/chamberInfo.py
```diff
@@ -34,8 +34,12 @@
             #placeholder
         }
 
-# Matches CMS coordinates
-chamber_iEta2VFATPos = {
+##### NEED TO FIX
+#### VERY TEMPORARY
+gemTypeMapping = {0:"ge11", 1:"ge21"}
+##### END
```
jsturdy: again

utils/latAlgos.py
```diff
@@ -71,17 +71,32 @@ def anaUltraLatency(infilename, debug=False, latSigMaskRange=None, latSigRange=N
...
+
+    ##### NEED TO FIX
+    #### VERY TEMPORARY
+    from gempython.gemplotting.mapping.chamberInfo import gemTypeMapping
+    if 'gemType' not in inFile.latTree.GetListOfBranches():
+        gemType = "ge11"
+    else:
+        gemType = gemTypeMapping[rp.tree2array(tree=inFile.latTree, branches =[ 'gemType' ] )[0][0]]
+    print gemType
+    ##### END
```
jsturdy, line 86: again

utils/latAlgos.py
```diff
@@ -129,13 +144,15 @@ def anaUltraLatency(infilename, debug=False, latSigMaskRange=None, latSigRange=N
...
-    for vfat in range(0,24):
+    for vfat in range(0,vfatsPerGemVariant[gemType]):
         for binX in range(1, dict_hVFATHitsVsLat[vfat].GetNbinsX()+1):
             dict_hVFATHitsVsLat[vfat].SetBinError(binX, sqrt(dict_hVFATHitsVsLat[vfat].GetBinContent(binX)))
 
     hHitsVsLat_AllVFATs = dict_hVFATHitsVsLat[0].Clone("hHitsVsLat_AllVFATs")
     hHitsVsLat_AllVFATs.SetTitle("Sum over all VFATs")
-    for vfat in range(1,24):
+    #### QUESTION 
+    # why is this only [1-24)
```
jsturdy: because it is cloned in line 151 from the VFAT0 histogram

utils/scurveAlgos.py
```diff
@@ -137,40 +137,48 @@ def anaUltraScurve(args, scurveFilename, calFile=None, GEBtype="short", outputDi
     # Get ChipID's
     import numpy as np
     import root_numpy as rp
+    ##### NEED TO FIX
+    #### VERY TEMPORARY
+    from gempython.gemplotting.mapping.chamberInfo import gemTypeMapping
+    gemType = gemTypeMapping[rp.tree2array(scurveTree, branches =[ 'gemType' ] )[0][0]]
+    print gemType
+    ##### END
```
jsturdy, line 143: again

utils/threshAlgos.py
```diff
@@ -128,17 +116,41 @@ def anaUltraThreshold(args,thrFilename,GEBtype="short",outputDir=None,fileScurve
...
+    
+    ##### NEED TO FIX
+    #### VERY TEMPORARY
+    from gempython.gemplotting.mapping.chamberInfo import gemTypeMapping
+    if 'gemType' not in thrTree.GetListOfBranches():
+        gemType = "ge11"
+    else:
+        gemType = gemTypeMapping[rp.tree2array(tree=thrTree, branches =[ 'gemType' ] )[0][0]]
+    print gemType
+    ##### END
```
jsturdy, line 142: again

utils/threshAlgos.py
```diff
@@ -553,15 +562,22 @@ def calibrateThrDAC(args):
...
+    from gempython.gemplotting.utils.anautilities import getCyclicColor, getDirByAnaType, filePathExists, parseListOfScanDatesFile
     parsedTuple = parseListOfScanDatesFile(args.inputFile)
     listChamberAndScanDate = parsedTuple[0]
     thrDacName = parsedTuple[1]
     chamberName = listChamberAndScanDate[0][0]
 
+    # ##### NEED TO FIX
+    # #### VERY TEMPORARY
+    gemType = chamberName[:chamberName.find("-")].lower()
+    # ##### END
```
jsturdy, line 587: again

utils/threshAlgos.py
```diff
@@ -1110,21 +1126,24 @@ def calibrateThrDAC(args):
...
     # Make summary canvases, always save these
-    canvScurveMeanByThrDac_Summary = make3x8Canvas("canvScurveMeanByThrDac_Summary",dict_mGraphScurveMean,"APE1")
-    canvScurveSigmaByThrDac_Summary = make3x8Canvas("canvScurveSigmaByThrDac_Summary",dict_mGraphScurveSigma,"APE1")
-    canvScurveMeanVsThrDac_Summary = make3x8Canvas("canvScurveMeanVsThrDac_Summary",dict_ScurveMeanVsThrDac,"APE1",dict_funcScurveMeanVsThrDac)
-    canvScurveSigmaVsThrDac_Summary = make3x8Canvas("canvScurveSigmaVsThrDac_Summary",dict_ScurveSigmaVsThrDac,"APE1")
-    canvScurveMeanVsThrDac_BoxPlot_Summary = make3x8Canvas("canvScurveMeanVsThrDac_BoxPlot_Summary",dict_ScurveMeanVsThrDac_BoxPlot,"candle1")
-    canvScurveSigmaVsThrDac_BoxPlot_Summary = make3x8Canvas("canvScurveSigmaVsThrDac_BoxPlot_Summary",dict_ScurveSigmaVsThrDac_BoxPlot,"candle1")
+    canvScurveMeanByThrDac_Summary = getSummaryCanvas(dict_mGraphScurveMean, name="canvScurveMeanByThrDac_Summary", drawOpt="APE1", gemType=gemType)
+    canvScurveSigmaByThrDac_Summary = getSummaryCanvas(dict_mGraphScurveSigma, name="canvScurveSigmaByThrDac_Summary", drawOpt="APE1", gemType=gemType)
+    ### FIX THIS
```
jsturdy: fix what?

utils/threshAlgos.py
```diff
@@ -1260,6 +1280,16 @@ def getDetName(entry):
...
 
+    ##### NEED TO FIX
+    #### VERY TEMPORARY
+    from gempython.gemplotting.mapping.chamberInfo import gemTypeMapping
+    if 'gemType' not in rateTree.GetListOfBranches():
+        gemType = "ge21"
+    else:
+        gemType = gemTypeMapping[rp.tree2array(tree=rateTree, branches =[ 'gemType' ] )[0][0]]
+    print gemType
+    ##### END
```
jsturdy: again

utils/threshAlgos.py
```diff
@@ -1288,12 +1318,13 @@ def getDetName(entry):
...
     dict_vfatCHVsDACNameX_Rate2D = ndict() #[dacName][ohKey][vfat] = TGraph2D
 
+    from gempython.tools.hw_constants import vfatsPerGemVariant
     # initialize a TGraphErrors and a TF1 for each vfat
     for idx in range(len(dacNameArray)):
         dacName = np.asscalar(dacNameArray[idx])
         for entry in crateMap:
             ohKey = (entry['shelf'],entry['slot'],entry['link'])
-            for vfat in range(0,25): #24th VFAT represents "overall case"
+            for vfat in range(0,vfatsPerGemVariant[gemType]+1): #24th VFAT represents "overall case"
```
jsturdy: unfortunately, there are two conventions for the "all VFATs summary" plot, one is to put it in index -1, one is to put it in index 24
this should be cleaned up everywhere, but this is outside this PR

utils/threshAlgos.py
```diff
@@ -1303,7 +1334,7 @@ def getDetName(entry):
                 dict_Rate1DVsDACNameX[dacName][ohKey][vfat].SetMarkerStyle(23)
                 dict_Rate1DVsDACNameX[dacName][ohKey][vfat].SetMarkerSize(0.8)
                 dict_Rate1DVsDACNameX[dacName][ohKey][vfat].SetLineWidth(2)
-
+                print dacName, ohKey, vfat
```
jsturdy: debugging prints should be removed

jsturdy (changes requested):
* extraneous whitespace should be cleaned up
* all debugging `print` statements should be removed (or added to an appropriate `debug` block)
* all necessary `print`s should be `python3` compatible, i.e., `print()` instead of `print`
* all new string formatting should use `"{}".format()` rather than `"%"()` style

I think there are some additional lookup optimizations that can be made, so that map lookups happen only as needed

jsturdy:
when you go to address comments, you should `rebase` onto the latest `cms-gem-daq-project/develop`

anaSBitMonitor.py
dteague: My errant emacs commands are showing :grin: Will Fix

anaSBitMonitor.py
dteague: I put these comments because I was unsure the formatting of the gemType variable in the tree. I formatted it as a branch with the name "gemType" containing a number that mapped ge11 => 0, ge21 => 1. I didn't think this was how the final code would be done.

Is there a gemType branch implemented in the current tree structure, and if so, what is the format? I can change all these "TEMPORARY"s to that. 

Also, ge21 must have slipped by after some tests, will change

mapping/chamberInfo.py
dteague: This is the adhoc mapping I made for the gemType branch: see above comment are the branch's format

anaSBitMonitor.py
dteague: defined variable nVFATS and used that in this (and other places)

anaSBitReadout.py
dteague: I'm putting this variable in `mapping/ChamberInfo.py` for now, I'll implement that now. This can also be done for strips, ie a STRIPS_PER_IPHI variable if we want
Or is the max strips always going to be equal to the max channel number?

anaSBitReadout.py
dteague: done

anaSBitReadout.py
dteague: fixed

anaXDAQLatency.py
dteague: See above comment on reading in gemType

utils/latAlgos.py
dteague: :heavy_check_mark:

utils/latAlgos.py
dteague: as above

anaXDAQLatency.py
dteague: I made a dedicated `nVFATS` variable for most functions and have them use that instead of the dictionary.

utils/threshAlgos.py
dteague: AH, this was a comment before i added the `addPlotToCanvas` function. Before, the `make3x8Canvas` function was called multiple times on the same canvas to fill all the histograms, so my note was make a funciton to reproduce this! Removed the old comment

dteague:
- I changed most all of the print statements to python 3 format
- I changed most all of the % string formatting to the preferred `.format()` function
- I made a variable for many function for max vfats and max channels (`nVFATS` and `maxChans`). - Of course the naming can be changed easily, but it removes multiple look ups of the vfat dictionary and remove some 128 magic numbers from the code 

I do have some questions as shown by the comments to code blocks, and after those are answered, I can fix and test on files again to make sure everything is working correctly

anaSBitMonitor.py
jsturdy: OK, keep a `FIXME` text in each side of the comment block, so that it can easily be found when doing the eventual cleanup

jsturdy:
OK, it would probably be good to sit together and go over the `rebase` workflow, as your previous attempt succeeded (but in a sub-optimal way, as you created a new merge and now have a bunch of duplicate commits)

dteague:
I looked into this and I removed the extraneous commits and rebased them on the newest changes. The commit log shows only the 6 commits on top of the develop branch, so all seems to have worked, but this might need to be looked over since my git skills are still pretty poor :stuck_out_tongue_closed_eyes:

anaSBitReadout.py
jsturdy: I think you can safely assume that here `128` should always be `CHANNELS_PER_VFAT`, as the partitions are almost surely defined by the locations of the VFATs

anaSBitReadout.py
```diff
@@ -312,16 +322,16 @@
             # filling ieta 1Dhistos
             ieta_h_strip[eta].Fill((phi-1)*128+strip[0])
```
jsturdy: Likewise here, `128` is `CHANNELS_PER_VFAT` (also L331, L332, L365, L367...)

anaSBitReadout.py
```diff
@@ -241,9 +250,9 @@
     ieta_h_sbitSize = ndict()
     ieta_h_delay = ndict()
 
-    for ieta in range(1, 9):
-        ieta_h_strip[ieta] = r.TH1F("h_ieta{0}_strips_vs_hit".format(ieta), "i#eta = {0} | i#phi (1,2,3)".format(ieta), 384, 0., 384.)
-        ieta_h_ch[ieta] = r.TH1F("h_ieta{0}_chan_vs_hit".format(ieta), "i#eta = {0} | i#phi (1,2,3)".format(ieta), 384, 0., 384.)
+    for ieta in range(1, maxiEta+1):
+        ieta_h_strip[ieta] = r.TH1F("h_ieta{0}_strips_vs_hit".format(ieta), "i#eta = {0} | i#phi (1,2,3)".format(ieta), 128*maxiPhi, 0., 128.*maxiPhi)
+        ieta_h_ch[ieta] = r.TH1F("h_ieta{0}_chan_vs_hit".format(ieta), "i#eta = {0} | i#phi (1,2,3)".format(ieta), maxChans*maxiPhi, 0., maxChans*maxiPhi)
         ieta_h_sbitSize[ieta] = r.TH1F("h_ieta{0}_sbitSize_vs_hit".format(ieta), "i#eta = {0} SBIT Size".format(ieta), 8, 0., 8.)
```
jsturdy, line 256: ~~Here *possibly* `8`-> `NUM_IETA_SECTORS`~~
Actually, I think it's the s-bit partition, which is not tied to the `ieta` sectors

anaSBitReadout.py
```diff
@@ -255,11 +264,12 @@
...
-    dict_h2d_ieta_ch[0] = r.TH2I('h_2d_ieta_ch', 'Chanels summary        (i#phi = 1,2,3);chan number;i#eta', 384, 0, 384, 8, 0.5, 8.5)
+    dict_h2d_ieta_strip[0] = r.TH2I('h2d_ieta_strip', 'Strips summary        (i#phi = 1,2,3);strip number;i#eta', 128*maxiPhi, 0, 128*maxiPhi, maxiEta, 0.5, 0.5+maxiEta)
+    dict_h2d_ieta_ch[0] = r.TH2I('h_2d_ieta_ch', 'Channels summary        (i#phi = 1,2,3);chan number;i#eta', maxChans*maxiPhi, 0, maxChans*maxiPhi, maxiEta, 0.5, 0.5+maxiEta)
 
     # loop over all branch names but the first (evnt num)
     from gempython.gemplotting.mapping.chamberInfo import chamber_vfatPos2iEtaiPhi as vfat_to_etaphi
+    
     from gempython.utils.gemlogger import printRed, printYellow
     print("Analyzing Raw Data\nThis may take some time please be patient")
     h_clusterMulti = r.TH1F("h_clusterMulti".format(vfat), "", 9,-0.5,8.5)
```
jsturdy, line 275: Here *possibly* `9`-> `NUM_IETA_SECTORS+1`, probably best to check the semantics when the `Fill` is done

fitting/fitScanData.py
```diff
@@ -313,7 +314,7 @@ def fit(self, debug=False):
                     
                     if debug:
                         if self.isVFAT3:
-                            print "| %i | %i | %i | %i | %f | %f | %f | %f | %f | %f | %f | %f | %f |"%(
+                            print("| {0} | {1} | {2} | {3} | {4} | {5} | {6} | {7} | {8} | {9} | {10} | {11} | {12} |".format(
```
jsturdy, line 319: It wasn't my intention with my comment to have you do **all** the conversions, (just the ones that were in code you added/modified), but it's a good thing to have done anyway :-)
However, here does the formatting stay constant?
I suppose it can guess at the type, just need to make sure there isn't any implicit type conversion happening during the formatting...

dteague: I ended up doing a replace for the files I edited, and changed up the numbers when need be! I can add types to these just to be on the safe side

jsturdy (commented):
A few small checks

anaSBitReadout.py
mexanick: With GE2/1 the eta mapping would be different, as each VFAT will control 64 strips in one eta partition and 64 strips in another, adjacent eta partition in a way, that a trigger pad will consist of 2 strips with the same phi and different eta coordinates. Thus said, the number of channels per VFAT will stay of course the same.

anaSBitMonitor.py
dteague: Done

anaSBitReadout.py
dteague: To make sure I understand how we want to implement this: because we are going to have an equal number of strips to vfats, we should just replace all instances of 128 with CHANNELS_PER_VFAT since the distinction won't matter much, eg:
```
ieta_h_strip[ieta] = r.TH1F("h_ieta{0}_strips_vs_hit".format(ieta), "i#eta = {0} | i#phi (1,2,3)".format(ieta), 128*maxiPhi, 0., 128.*maxiPhi)
...
ieta_h_strip[eta].Fill((phi-1)*128+strip[0])
```

If this is correct, I'll go ahead and make that change

utils/scurveAlgos.py
```diff
@@ -1075,6 +1095,7 @@ def fill2DScurveSummaryPlots(scurveTree, vfatHistos, vfatChanLUT, vfatHistosPanP
         # Fill Summary Histogram 
         if lutType is mappingNames[1] and vfatHistosPanPin2 is not None:
             if (stripPinOrChan < 64):
+                #### Off by one error...?
                 vfatHistos[event.vfatN].SetBinContent(63-(stripPinOrChan+1),chargeBin,event.Nhits)
                 vfatHistos[event.vfatN].SetBinError(63-(stripPinOrChan+1),chargeBin,sqrt(event.Nhits))
```
dteague: This seems like an off by one error, namely the (64-1) and (stripPinOrChan+1) seem to be in opposition, eg 
```stripPinOrChan = 0   =>   bin = 62 ```
```stripPinOrChan = 63   =>   bin = -1```
Want to confirm before I try and change this

anaSBitReadout.py
dteague: Followed instructions from meeting

anaSBitReadout.py
dteague: done

anaSBitReadout.py
dteague: Hard to tell, but I don't think. This variable is based on a loop that runs over the clusterNames which are derived from the branches in file, ie 
https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/c987b20144e9e54619277a40c1d57772e5648474/anaSBitReadout.py#L282-L295

and 

https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/c987b20144e9e54619277a40c1d57772e5648474/anaSBitReadout.py#L190-L193

utils/scurveAlgos.py
jsturdy: I don't know...
* What is the valid range of values for `stripPinOrChan`?
* What do the current histograms look like (especially if you show the over/underflow stats)
* All things being equal, it looks to me like the `+1` should be outside the parentheses, and maybe also `vfatHistos` should be `vfatHistosPanPin2`...

dteague: I did some tests, and having the +1 outside was correct. stripPinOrChan in {0,...,127} and the code works by if done by PanPin, the channels are split between the two histograms, `vfatHistos` and `vfatHistosPanPin2`, each having a domain of [0,63].

All that being said, I made the fix!

dteague:
Ran tests and made minor changes. Tests are the same as before, namely:

- anaDACScan.py, anaUltraScurve.py, anaSBitThresh.py, anaUltraLatency.py - work for both
- anaUltraThreshold.py, anaSBitMonitor.py - work for ge11, could not fine ge21 file
- anaSBitReadout.py, anaXDAQLatency.py - could not find input file for either

I did notice that `anaSBitMonitor` gave strange results for GE11, mainly that all vfats pointed to vfat 7 and were misID'd the same number of times for each vfat. Considering not much was changed with this file, I'm inclined to believe this is a problem with the file I'm running over, not the changes made

All of the gemType reading in is surrounded by comments saying `##### FIXME` so when a proper gemType read in is decided, that can be changed quickly. For all of these tests, I just changed the default value to "ge21" and recompiled.

If anyone has files for run over the files I haven't checked or wants other files, namely in the `macros` area changed, I can do that. Otherwise, I hopefully have fixed for all of the above comments.

jsturdy:
> Considering not much was changed with this file, I'm inclined to believe this is a problem with the file I'm running over, not the changes made

You can verify by using the unmodified code to run the same analysis on the same file (since it's GE1/1)

dteague:
I ran the code in the develop branch, and it seems like this code isn't used much because it doesn't work out of the box! (`args.filename` should be `args.infilename`. It is fixed in this PR). But the develop branch version gives the exact same results, so this is just a strange fluke of the file I think.

jsturdy:
> I ran the code in the develop branch, and it seems like this code isn't used much because it doesn't work out of the box! (`args.filename` should be `args.infilename`. It is fixed in this PR). But the develop branch version gives the exact same results, so this is just a strange fluke of the file I think.

@AndrewLevin there was some issue in the past that was related to this `filename` vs `infilename` if I recall correctly, was this resolved?  is the change that @dteague is making here counter to that discussion?
(@dteague, you can probably search the issues for that string to find what the discussion was about)

dteague:
Pinging for status
Also with #256 , it might be good to fix this concurrently with this PR else GE21 plots can't be graphed (will look into fixes and possibly make another PR)

mexanick:
@dteague the issue #256 requires a separate PR. Would be great if you can provide it

AndrewLevin:
> > I ran the code in the develop branch, and it seems like this code isn't used much because it doesn't work out of the box! (`args.filename` should be `args.infilename`. It is fixed in this PR). But the develop branch version gives the exact same results, so this is just a strange fluke of the file I think.
> 
> @AndrewLevin there was some issue in the past that was related to this `filename` vs `infilename` if I recall correctly, was this resolved? is the change that @dteague is making here counter to that discussion?

I think maybe you are referring to this change

https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/239/files#diff-058f061e69c295595d79241cf963cf20L105-R106

but that has nothing to do with this.

> (@dteague, you can probably search the issues for that string to find what the discussion was about)

jsturdy:
> > > I ran the code in the develop branch, and it seems like this code isn't used much because it doesn't work out of the box! (`args.filename` should be `args.infilename`. It is fixed in this PR). But the develop branch version gives the exact same results, so this is just a strange fluke of the file I think.
> > 
> > 
> > @AndrewLevin there was some issue in the past that was related to this `filename` vs `infilename` if I recall correctly, was this resolved? is the change that @dteague is making here counter to that discussion?
> 
> I think maybe you are referring to this change
> 
> https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/239/files#diff-058f061e69c295595d79241cf963cf20L105-R106
> 
> but that has nothing to do with this.
> 
> > (@dteague, you can probably search the issues for that string to find what the discussion was about)

OK, thanks @AndrewLevin, I just wanted to make sure that this change was properly motivated and not rehashing something historical.

jsturdy (dismissed):

jsturdy:
@dteague, sorry, can you `rebase` this one final time?
just want to make sure it would be a good merge as you've touched lots of files

dteague:
Just took care of that. Hopefully I rebased correctly and all is up to snuff!

jsturdy (approved):

jsturdy (dismissed):
Ready for squash

lpetre-ulb:
While testing the new GE2/1 setup at FIT, Stephen and I encountered an issue with the usage of this PR within the [GE2/1 compatibility PR in `vfatqc-python-tools`](https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/265).

Indeed, in `getVFAT3CalInfo` the `gemType` argument is not set in `testConnectivity.py` nor in any other script; this is an issue for GE2/1 operation. While this seems an issue related to `vfatqc-python-tools` PR, the design in `dbutils.py` generalization could be improved to prevent this kind of problem.

In fact, why is the GEM type even required to be given to the DB util functions? I mean, one should be able to request the data for an arbitrary number of VFATs; 24 for GE1/1, 12 for GE2/1 or n for a manual lookup.

I see that `gemType` is used only here:
https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/6612de6cc575eef2b588bc4678c559c80a13d6a5/utils/dbutils.py#L176

Couldn't `vfatsPerGemVariant[gemType]` be advantageously replaced with `len(vfatList)`? This would make the DB util functions fully generic without any requirement on the GEM type. As a side effect, the PR in `vfatqc-python-scripts` would work without modification.

dteague:
>Couldn't vfatsPerGemVariant[gemType] be advantageously replaced with len(vfatList)? This would make the DB util functions fully generic without any requirement on the GEM type. As a side effect, the PR in vfatqc-python-scripts would work without modification.

Very fair, @lpetre-ulb, that is a much better implementation that I didn't see! Let me fix that quickly

lpetre-ulb:
Thank you for the quick fix @dteague!

jsturdy:
~~@mexanick, do we want to hold this while @sbutalla tests at FIT, or just merge it as is and do bugfixes later?~~
oops! too slow of a refresh! (thanks @lpetre-ulb!)

jsturdy (approved):
now ready for squash

mexanick (approved):
I think it is fine to merge and start testing. Bug reports/fixes should be provided upon necessity.

lpetre-ulb:
@sbutalla and I discovered what seems like a bug yesterday in the DAC scans plots for GE2/1 (at least). The title is a fifth order polynomial and the axis do not have any legend. Stephen, can you open an issue? I don't have access to the plots.

PR #255, opened 2019-10-18
[Bug Fix] In DAC scan analysis, ADC error was not being set correctly

AndrewLevin:
We are applying a scaling factor to the calibrated ADC value, and we should also apply this to its error.

### Types of changes

## Motivation and Context
This pull request should resolve the issues with BLCC DAC scans reported in https://indico.cern.ch/event/855869/contributions/3601787/attachments/1929074/3194549/dac_variations_qc8_vs_p5_18-10.2019.pdf

## How Has This Been Tested?
Yes, I have tested this on coffin data:

With the HEAD of develop:

SummaryGE11-X-S-INDIA-0002_DACScan_CFG_BIAS_PRE_I_BLCC

With this pull request:

SummaryGE11-X-S-INDIA-0002_DACScan_CFG_BIAS_PRE_I_BLCC

## Checklist:

### DIFF
--- a/utils/anautilities.py
+++ b/utils/anautilities.py
@@ -277,6 +277,7 @@ def dacAnalysis(args, dacScanTree, chamber_config, scandate='noscandate'):
                 calibrated_ADC_value -= nominal_iref 
 
             calibrated_ADC_value /= nominalDacScalingFactors[dacName]
+            calibrated_ADC_error /= nominalDacScalingFactors[dacName]
                 
         #the reversal of x and y is intended - we want to plot the dacName variable on the y-axis and the adcName variable on the x-axis
         #the dacName variable is the DAC register that is scanned, and we want to determine its nominal value


jsturdy:
So this was not being done for *any* of the DACs, correct?
Just in this case (and maybe also `CFG_HYST` we saw it more clearly due to the range issue?
(can you also show a before/after of `CFG_HYST` and another that were already appearing "normal"?)

AndrewLevin:
Yes, this was not being done for any of the DACs, but it has the largest effect for those DACs with scaling factors most different from 1. Because we have a lot of data points to fit, the errors are less important unless they are wildly off, which is the case for BLCC, because the scaling factor is 100.

CFG_HYST DAC scan analysis with HEAD of develop:

SummaryGE11-X-S-INDIA-0002_DACScan_CFG_HYST

CFG_HYST DAC scan analysis with this pull request:

SummaryGE11-X-S-INDIA-0002_DACScan_CFG_HYST

CFG_BIAS_SD_I_BSF DAC scan analysis with HEAD of develop:

SummaryGE11-X-S-INDIA-0002_DACScan_CFG_BIAS_SD_I_BSF

CFG_BIAS_SD_I_BSF DAC scan analysis with this pull request:

SummaryGE11-X-S-INDIA-0002_DACScan_CFG_BIAS_SD_I_BSF

jsturdy (approved):

PR #264, opened 2019-11-18
Get `VREF_ADC` values from the database

lpetre-ulb:
## Description
[See cms-gem-daq-project/vfatqc-python-scripts#288 for more details.]

This PR makes the following two changes:

* Don't filter out the `vref_adc` column from the database in `getVFAT3CalInfo`.
* Also fix an long standing misleading error message; even when all VFATs were found in the DB the following warning message was printed out:

``` bash
Length of returned view does not match length of input vfat List
VFATs not found: []
```

### Types of changes

## Motivation and Context

DAC scan failures encountered during commissioning at P5.

## How Has This Been Tested?

`VREF_ADC` values are correctly written to files in the `$GEM_PATH` and on the CTP7 with `getCalInfoFromDB.py`.

## Checklist:


jsturdy:
Does #263 also try to solve the same problem in a different way?
Or is the problem #263 solves independent from the problem(s) addressed by this?

utils/dbutils.py
```diff
@@ -77,7 +77,7 @@ def getGEMDBView(view, vfatList=None, debug=False):
 
     if vfatList is not None:
         # First warn the user which VFATs are *not* found
-        if len(vfatList) != dfGEMView.shape[1]:
+        if len(vfatList) != dfGEMView.shape[0]:
```
jsturdy, line 80: for my curiosity, what is the dimensionality of `dfGEMView` (or what did the original code think/expect to extract from the old index)?

lpetre-ulb:
This PR is not related to #263, but is a dependency of https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/288. The `vref_adc` column was previously filtered out in `getVFAT3CalInfo` while its values are very important (as the past few days have shown ;)

The misleading warning message was [also seen](https://mattermost.web.cern.ch/cms-gem-ops/pl/y7kq7611kbd48p6tfh77kwwoth) during debugging at P5, but is not the same as the one #263 is trying to fix.

utils/dbutils.py
lpetre-ulb: The first index refers to the rows and the second one to the columns. In our case, the dimensionality is therefore (number of VFATs retrieved from DB, number of columns in the requested view).

AndrewLevin (approved):

jsturdy (approved):

PR #267, opened 2019-11-26
chamberInfo.py should be replaced when the rpm is installed

AndrewLevin:
## Description
This pull request simply removes the line that prevents `chamberInfo.py` from being overwritten.

### Types of changes

## Motivation and Context
If `chamberInfo.py` is not replaced, new features in `chamberInfo.py` are not installed.

## How Has This Been Tested?
Yes, I have installed the package in my virtual environment

## Checklist:

### DIFF
--- a/pkg/installrpm.sh
+++ b/pkg/installrpm.sh
@@ -25,11 +25,9 @@ cat <<EOF >>INSTALLED_FILES
 %attr(0755,root,root) /opt/cmsgemos/bin/*.py
 # %attr(0755,root,root) %{python2_sitelib}/gempython/scripts/*.py
 # %attr(0755,root,root) %{python2_sitelib}/gempython/gemplotting/macros/*.py
-# %config(noreplace) %attr(0755,root,root) %{python2_sitelib}/gempython/gemplotting/mapping/chamberInfo.py
 %attr(0755,root,root) /usr/lib/python*/site-packages/gempython/scripts/*.py
 %attr(0755,root,root) /usr/lib/python*/site-packages/gempython/scripts/*.py
 %attr(0755,root,root) /usr/lib/python*/site-packages/gempython/gemplotting/macros/*.py
-%config(noreplace) %attr(0755,root,root) /usr/lib/python*/site-packages/gempython/gemplotting/mapping/chamberInfo.py
 EOF
 echo "Modified INSTALLED_FILES"
 cat INSTALLED_FILES


lpetre-ulb (dismissed):
Shouldn't the following line also be removed to prevent any further issue when un-commenting it?

https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/09b17ebf17aaad977b132eb1f2d96c9b39232154/pkg/installrpm.sh#L28

lpetre-ulb (approved):
Trivial change.

AndrewLevin:
Yes, I guess that commented line is for installation in a virtual environment, although in my virtual environment I don't actually have an environment variable $PYTHON2_SITELIB.

lpetre-ulb:
> Yes, I guess that commented line is for installation in a virtual environment, although in my virtual environment I don't actually have an environment variable $PYTHON2_SITELIB.

Indeed, it is designed to automatically work with different Python `site-packages` locations. The main target is the support of different Python versions. The variable will correctly refer to your virtual environment, even if an RPM should never refer to a specific virtual environment...

The syntax `%{python2_sitelib}` is a RPM macro (and not an environment variable) which is defined as: `python2 -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())"`.

jsturdy:
This behaviour is actually by design, (because this file holds configuration information) it should **not** overwrite a potentially system specific file.
When installing, if `rpm` notices that there are differences, it will create `rpmnew` and `rpmsave` files, but it will not replace the current file [cf](http://ftp.rpm.org/max-rpm/s1-rpm-inside-files-list-directives.html), and when this happens, the command will tell you that it happened, so it's up to the sysadmin to notice that there are changes and resolve them, if desired.

The `%config(noreplace)` must stay, and using the proper `rpm` macros would be good, it just needs to be validated against a future (`cc8`) world where we'll provide a `python3` package

AndrewLevin:
At least in our QC7 and QC8 setups, the system specific information is stored in a separate file called `system_specific_constants.py`, no in `chamberInfo.py`.

jsturdy:
OK, if we've completely removed any "editable" features from this file, then I jumped the gun on closing this.

lpetre-ulb:
Yes, all "editable" dictionaries are moved to the `system_specific_constant` module which must be present in the `PYTHONPATH` (or be importable somehow). `chamberInfo.py` is kept only for backwards compatibility throughout the code base and import the dictionaries from `system_specific_constant` for any setup-specific configuration.

AndrewLevin:
It is possible to edit the dictionaries directly in the `chamberInfo.py` which will be used if the `system_specific_constants.py` is not found, but I think the proper way is to separate the system specific information like we are doing in all of the 904 setups.

jsturdy (approved):
Before installing, should remove symlinks on the system, if present

PR #268, opened 2019-11-26
[Bug Fix] This is to fix an indexing issue causing problems in qc7/8

bregnery:
This is a problem observed in qc7/8 but not in my own venv.

## Description
This should fix the sbitThresh analysis

### Types of changes

## Motivation and Context
Required to work in the qc7/8 environment

## How Has This Been Tested?
```bash
#!/usr/bin/env bash

#=========================================================================================
# Testing Suite for the Inflection Point finder ------------------------------------------
#=========================================================================================


echo "Welcome to Brendan's Testing Suite for the Inflection Point Finder"
echo "------------------------------------------------------------------------------------"
echo "This script executes a bunch of files known to cause issues for the inflection"
echo "point finder"
echo "------------------------------------------------------------------------------------"

echo "Did you remember to update??? With make pip and pip install rpm*.tar.gz?"

echo "File 1"
echo "This tests how the code handles a vfat with a bad vfat"
echo "Please make sure (2,5,4) is filled in your system_specfic_constants.py file"

# This file causes issues because of a zero slope issue
anaSBitThresh.py /data/bigdisk/GEM-Data-Taking/GE11_QC8//sbitRate/channelOR/2019.10.25.13.59/SBitRateData.root

echo "File 2"
echo "This tests how the code handles a detector with missing vfats"
echo "Please make sure (1,2,0) is filled in your system_specfic_constants.py file"

# This file is from the coffin where the detector is missing vfats
anaSBitThresh.py /data/bigdisk/GEM-Data-Taking/GE11_Integration//sbitRate/channelOR//2019.10.24.17.45/SBitRateData.root
```

## Checklist:

### DIFF
--- a/pkg/setup.cfg
+++ b/pkg/setup.cfg
@@ -71,7 +71,7 @@ requires = python >= 2.6
            numpy >= 1.7
            python-pandas
            python-sqlalchemy
-           root_numpy >= 4.7
+           root_numpy >= 4.8
            tabulate >= 0.4.4
 #           tabulate < 0.8 ## need to get this dynamically, as it depends on the python version
 install_script = installrpm.sh
--- a/pkg/setup.cfg
+++ b/pkg/setup.cfg
@@ -68,10 +68,10 @@ build_requires = python >= 2.6
 requires = python >= 2.6
            cmsgemos_gempython >= 0.3.1
            cx_Oracle
-           numpy >= 1.7
+           numpy >= 1.14
            python-pandas
            python-sqlalchemy
-           root_numpy >= 4.8
+           root_numpy >= 4.7
            tabulate >= 0.4.4
 #           tabulate < 0.8 ## need to get this dynamically, as it depends on the python version
 install_script = installrpm.sh
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,9 @@
 cx_Oracle
-numpy>1.7,<1.12;  python_version == '2.6'
-numpy>1.7,<1.12;  python_version == '3.2'
-numpy>1.7,<1.12;  python_version == '3.3'
-numpy>1.16,<1.17; python_version == '2.7'
-numpy>1.7; python_version >= '3.4'
+numpy>1.14,<1.17; python_version == '2.6'
+numpy>1.14,<1.17; python_version == '3.2'
+numpy>1.14,<1.17; python_version == '3.3'
+numpy>1.14,<1.17; python_version == '2.7'
+numpy>1.14; python_version >= '3.4'
 pandas
 ## currently conflicts if numpy isn't installed, as it will download it's own version due to steup_requries
 root_numpy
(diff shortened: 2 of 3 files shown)


jsturdy:
What are the versions of dependencies in the two environments?

bregnery:
how do I figure that out?

When I re-based, I noticed new dependencies.

The old version accepted the following syntax `dict_dacInflectPts[dacName][ohKey][vfat][0]` which returned `[15.0]`, which I see is not quite right, but it handled it. Then some dependency was added that would not accept this input and I had to change it to `dict_dacInflectPts[dacName][ohKey][vfat][0][0]` so it returned `15.0` which is what I really meant for it to be in the first place.

jsturdy:
probably something like `pip freeze` from your environment, as well as the same for QC8
mostly interested in `numpy`/`root_numpy`/`pandas`, not sure there are other relevant deps

bregnery:
This bug also does not affect the p5 computer

bregnery:
For numpy:
```bash
(inflectionEnv)-bash-4.2$ pip show numpy
Name: numpy
Version: 1.14.6
```

jsturdy:
* which version is at P5?
* which version(s) is(are) on QC8/7 in the environment(s) that is(are) being used there?
* which package's API changed and in which version?

once these are answered, the appropriate changes to the requirements must be made as well

jsturdy:
OK, I'm going to make an executive decision and we'll run `numpy==1.16.5` and `root_numpy==4.8.0` on our production machines (so code **must** be tested to work with at least these)

jsturdy:
@bregnery needs another rebase, and can you make sure it works as expected based on the aforementioned versions?

bregnery:
Okay, it works with the new requirements. Now I just need to re-base.

bregnery:
Done

jsturdy (commented):
Can you update the relevant requirements files? (`requirements.txt`, `pkg/setup.cfg`, etc)

jsturdy:
@bregnery bump

bregnery:
updated and it compiles/runs. But @jsturdy, I think that the requirements might not be getting parsed correctly.

pkg/setup.cfg
```diff
@@ -71,7 +71,7 @@ requires = python >= 2.6
            numpy >= 1.7
```
jsturdy: this version should match the one in the requirements, which i think can be `>=1.14` for the features actually in use

pkg/setup.cfg
```diff
@@ -71,7 +71,7 @@ requires = python >= 2.6
...
            python-pandas
            python-sqlalchemy
-           root_numpy >= 4.7
+           root_numpy >= 4.8
```
jsturdy: this also doesn't need to change, things work fine with `4.7`

jsturdy (changes requested):

jsturdy (approved):
:+1:

PR #272, opened 2019-11-29
Error Fix

dteague:
This is so silly. Fix for issue #262 . Only needed to put a "same" in the draw argument. This only fixes the plotting issue, not the chi2 one (some values are fairly large)

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change

## How Has This Been Tested?
Using files provided in issue #262

### DIFF
--- a/utils/anautilities.py
+++ b/utils/anautilities.py
@@ -106,6 +106,7 @@ def dacAnalysis(args, dacScanTree, chamber_config, scandate='noscandate'):
         detName = getDetName(entry)
         return detName[:detName.find('-')].lower()
     gemType = getGemType(crateMap[0])
+    #    gemType='ge11'
     ### END
     from gempython.tools.hw_constants import vfatsPerGemVariant
     nVFATS = vfatsPerGemVariant[gemType]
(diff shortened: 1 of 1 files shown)


jsturdy:
This is confirmed fixed for both GE1/1 and GE2/1 (not sure exactly which files were provided for testing in #262)?

dteague:
For the test files, it was the zipped up [files sent by Laurent](https://github.com/cms-gem-daq-project/gem-plotting-tools/files/3859310/BadDACScanPlots.zip) in the issue #262 . For the fixed output, that can be [seen here](https://dteague.web.cern.ch/dteague/GEMWork/gemplotting_tools/Issue262/2019-12-01_TestStandGE11_out/)

[Output of GE21 test here](https://dteague.web.cern.ch/dteague/GEMWork/gemplotting_tools/Issue262/2019-12-01_GE21-M2-P4_out/) I am working from the US, so some things are acting strange probably as a result, e.g. I can't access bigdisk on the gemdaq machines. So I used a GE21 file and calfile I had lying around for this. On face value, it seems for both cases, the plots look better, though the chi2 error still comes up

jsturdy (approved):
:+1:

PR #273, opened 2019-12-03
Reorganize DAC analysis exception classes

AndrewLevin:
A new exception class `DACAnalysisException` is introduced which combines `VFATDACBiasCannotBeReached` and `VFATDACFitLargeChisquare`.

## Description
The new exception class contains two `bool`s which identify which problem or problems cause the exception, and a message with the details.

### Types of changes

## Motivation and Context
In the case of both a bad bias and a bad fit exception in the same DAC scan, only one of the two exceptions would be thrown, arbitrarily.

## How Has This Been Tested?
I have reanalyzed a DAC scan which has both types of exceptions.

## Checklist:

### DIFF
--- a/utils/exceptions.py
+++ b/utils/exceptions.py
@@ -15,6 +15,18 @@ Documentation
 """
 
 class VFATDACBiasCannotBeReached(ValueError):
+
+    def __init__(self, message, errors):
+        super(VFATDACBiasCannotBeReached, self).__init__(message)
+
+        self.errors = errors
+        return
+
+class DACAnalysisException(ValueError):
+    """
+    This exception is raised when a DAC vs ADC fit returns a large chisquare value
+
+    """
     """
     This exception is raised if DAC analysis determines that to achieve the 
     nominal bias current/voltages in ``nominalDacValues`` dictionary, 
@@ -23,20 +35,11 @@ class VFATDACBiasCannotBeReached(ValueError):
     ``maxVfat3DACSize[dacSelect][0]``.  Info can be found on these ranges in
     the ``maxVfat3DACSize`` dictionary of: ``gempython.tools.hw_constants``
     """
-    def __init__(self, message, errors):
-        super(VFATDACBiasCannotBeReached, self).__init__(message)
+    def __init__(self, isBadFit, isBadBias, message, errors):
+        super(DACAnalysisException, self).__init__(message)
 
-        self.errors = errors
-        return
-
-class VFATDACFitLargeChisquare(ValueError):
-    """
-    This exception is raised when a DAC vs ADC fit returns a large chisquare value
-
-    """
-    def __init__(self, message, errors):
-        super(VFATDACFitLargeChisquare, self).__init__(message)
-
+        self.isBadFit = isBadFit
+        self.isBadBias = isBadBias
         self.errors = errors
         return
 
--- a/utils/exceptions.py
+++ b/utils/exceptions.py
@@ -14,21 +14,11 @@ Documentation
 -------------
 """
 
-class VFATDACBiasCannotBeReached(ValueError):
-
-    def __init__(self, message, errors):
-        super(VFATDACBiasCannotBeReached, self).__init__(message)
-
-        self.errors = errors
-        return
-
 class DACAnalysisException(ValueError):
     """
-    This exception is raised when a DAC vs ADC fit returns a large chisquare value
-
-    """
-    """
-    This exception is raised if DAC analysis determines that to achieve the 
+    This is exception is raised when one or more of the following flags is set.
+    isBadFit: This flag is set when a DAC vs ADC fit returns a large chisquare value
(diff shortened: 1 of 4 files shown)


jsturdy (commented):
It looks like you've changed `VFATDACFitLargeChisquare` to `DACAnalysisException`, but I don't see any downstream changes that were `catch`ing the old exception having been changed.
Is this breaking in that regard?
Does code need to be modified to work with this change if it is merged in?

AndrewLevin:
> It looks like you've changed `VFATDACFitLargeChisquare` to `DACAnalysisException`, but I don't see any downstream changes that were `catch`ing the old exception having been changed.
> Is this breaking in that regard?
> Does code need to be modified to work with this change if it is merged in?

Yes, there is code that catches this exception in `testConnectivity.py`. I will prepare a pull request for `vfatqc-python-scripts`.

mexanick:
@AndrewLevin is this stale?

AndrewLevin:
no, it is not stale

lpetre-ulb:
@AndrewLevin, is this PR still relevant? In that case, a twin PR should be provided in `vfatqc-python-scripts`.

lpetre-ulb:
@AndrewLevin, can we close this PR? Or will the remaining work ever be carried on?

AndrewLevin:
the work is already carried out actually:

https://github.com/AndrewLevin/vfatqc-python-scripts/commit/d7b24bbe4c74b59a8ad970a28aac03181252a120

I will test it tomorrow

utils/exceptions.py
```diff
@@ -15,28 +15,31 @@
 """
 
 class VFATDACBiasCannotBeReached(ValueError):
```
lpetre-ulb: Why does this class still exist? Isn't the spirit of this PR to merge all exceptions from DAC scan analysis into `DACAnalysisException`? From what I see, it is not raised anywhere anymore, but is caught in different places (where `DACAnalysisException` should be, e.g. in `ana_scans.py`).

utils/exceptions.py
```diff
@@ -15,28 +15,31 @@
...
-    """
-    This exception is raised if DAC analysis determines that to achieve the 
-    nominal bias current/voltages in ``nominalDacValues`` dictionary, 
-    of: ``gempython.gemplotting.utils.anaInfo``, the DAC would need to be set
-    to a value outside of the DAC range ``[0, max]`` where ``max`` is
-    ``maxVfat3DACSize[dacSelect][0]``.  Info can be found on these ranges in
-    the ``maxVfat3DACSize`` dictionary of: ``gempython.tools.hw_constants``
-    """
```
lpetre-ulb: If the the class is still destined to remain, why is its docstring removed?

utils/exceptions.py
```diff
@@ -15,28 +15,31 @@
...
-    def __init__(self, message, errors):
-        super(VFATDACFitLargeChisquare, self).__init__(message)
+    """
+    This exception is raised if DAC analysis determines that to achieve the 
+    nominal bias current/voltages in ``nominalDacValues`` dictionary, 
+    of: ``gempython.gemplotting.utils.anaInfo``, the DAC would need to be set
+    to a value outside of the DAC range ``[0, max]`` where ``max`` is
+    ``maxVfat3DACSize[dacSelect][0]``.  Info can be found on these ranges in
+    the ``maxVfat3DACSize`` dictionary of: ``gempython.tools.hw_constants``
+    """
```
lpetre-ulb: The documentation should describe the current behavior of the exception class.

lpetre-ulb (changes requested):

utils/exceptions.py
AndrewLevin: removed

utils/exceptions.py
AndrewLevin: the class is removed

utils/exceptions.py
AndrewLevin: documentation updated

utils/exceptions.py
AndrewLevin: and replaced in ana_scans.py now

lpetre-ulb: `VFATDACBiasCannotBeReached` still present in `anaDACScan.py`.

Could you be more careful with your PR in the legacy branches? The policy is to not accept new features, particularly, for quality of life features. We made an exception for this PR which was already open, but one must be very cautious not to add any new bug.

lpetre-ulb:
@AndrewLevin, have you checked that the changes are thorough and thoroughly tested?

AndrewLevin:
well, I have did some tests, but there are so many dependencies it is hard to know if I missed something (which is why this should done by gitlab CI instead of a human...)

I think it is unlikely that there is an inconsistency now (anyway it is an exception, and would just not be caught), but it is fine with me if you want to not merge it

lpetre-ulb (approved):

Issue #288, opened 2020-02-12
sBitRate scan  analysis not working with GLIB


cgalloni:
The  script to analyze the root file of an sBitThresh scan taken with the GLIB fails with the error:

```
Traceback (most recent call last):
  File "/opt/cmsgemos/bin/ana_scans.py", line 909, in <module>
    args.func(args)
  File "/opt/cmsgemos/bin/ana_scans.py", line 113, in anaSBITThresh
    scandate = args.scandate)
  File "/usr/lib/python2.7/site-packages/gempython/gemplotting/utils/threshAlgos.py", line 1502, in sbitRateAnalysis
    if dict_dacInflectPts[dacName][ohKey][vfat][0] == None:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

## Brief summary of issue

The  script to analyze the root file of an sBitThresh scan taken with the GLIB fails.  This might be related to the fact that for VFAT 0 and 1 the rate at CFG_THR_ARM_DAC 0 is 0, reported in [#286]

### Types of issue

- [X] Bug report (report an issue with the code)
- [ ] Feature request (request for change which adds functionality)

## Expected Behavior

It should produce the output files with the results of the analysis

## Current Behavior

```
ana_scans.py sbitThresh -i /data/bigdisk/GEM-Data-Taking/GE11_QC8///sbitRate/channelOR//2020.02.12.11.33/SBitRateData.root --chamberConfig -m 100
Analyzing: /data/bigdisk/GEM-Data-Taking/GE11_QC8///sbitRate/channelOR//2020.02.12.11.33/SBitRateData.root
Loading info from input TTree
Looping over stored events in rateTree
Determine when SBIT rate falls below 100.0 Hz and writing output data
| Geo Addr   |   VFAT Number |   ARM DAC Inflection Pt |
|------------+---------------+-------------------------|
| (2, 3, 0)  |             0 |                       4 |
| (2, 3, 0)  |             1 |                       3 |
| (2, 3, 0)  |             2 |                       4 |
| (2, 3, 0)  |             3 |                       3 |
| (2, 3, 0)  |             4 |                       3 |
| (2, 3, 0)  |             5 |                       6 |
| (2, 3, 0)  |             6 |                       4 |
| (2, 3, 0)  |             7 |                       3 |
| (2, 3, 0)  |             8 |                       2 |
| (2, 3, 0)  |             9 |                       2 |
[...]

PR #289, opened 2020-02-13
[BugFix] Addressing Multi-valued inflection point issues

bregnery:
## Description
Previously the inflection point finder allowed for multiple values, we have prevented this from happening by selecting the largest value when more than one is selected. Also, it was requested by the shifters for us to remove the red line from the sbit graphs.

### Types of changes

## Motivation and Context
This bug prevented tests from being run at qc7 and needed to be addressed.

## How Has This Been Tested?
```bash
anaSBitThresh.py /data/bigdisk/GEM-Data-Taking/GE11_QC8///sbitRate/channelOR//2020.02.12.11.33/SBitRateData.root

echo "File 1"
echo "This tests how the code handles a vfat with a bad vfat"
echo "Please make sure (2,5,4) is filled in your system_specfic_constants.py file"

# This file causes issues because of a zero slope issue
anaSBitThresh.py /data/bigdisk/GEM-Data-Taking/GE11_QC8//sbitRate/channelOR/2019.10.25.13.59/SBitRateData.root

echo "File 2"
echo "This tests how the code handles a detector with missing vfats"
echo "Please make sure (1,2,0) is filled in your system_specfic_constants.py file"

# This file is from the coffin where the detector is missing vfats
anaSBitThresh.py /data/bigdisk/GEM-Data-Taking/GE11_Integration//sbitRate/channelOR//2019.10.24.17.45/SBitRateData.root
```

## Checklist:

### DIFF
--- a/utils/anautilities.py
+++ b/utils/anautilities.py
@@ -556,8 +556,7 @@ def findInflectionPts(graph):
     # We are defining the inflection point as the point where the most negative gradient sum begins
     try:
         inflxGrad = negGrad[bigIdx][0]
-        inflxIdices = np.where(grad == inflxGrad) #return the index at the specified value
-        inflxIdx = np.amax(inflxIdices) # if there are multiple possible points, take the largest one
+        inflxIdx = np.amax(np.where(grad == inflxGrad) ) #return the index at the specified value, if there are multiple possible points, take the largest one
         inflxPnt = (x[inflxIdx], y[inflxIdx] )
     # Error handling for problematic VFATs
     except IndexError:
--- a/utils/anautilities.py
+++ b/utils/anautilities.py
@@ -556,7 +556,8 @@ def findInflectionPts(graph):
     # We are defining the inflection point as the point where the most negative gradient sum begins
     try:
         inflxGrad = negGrad[bigIdx][0]
-        inflxIdx = np.where(grad == inflxGrad) #return the index at the specified value
+        inflxIdices = np.where(grad == inflxGrad) #return the index at the specified value
+        inflxIdx = np.amax(inflxIdices) # if there are multiple possible points, take the largest one
         inflxPnt = (x[inflxIdx], y[inflxIdx] )
     # Error handling for problematic VFATs
     except IndexError:
--- a/utils/threshAlgos.py
+++ b/utils/threshAlgos.py
@@ -1452,7 +1452,7 @@ def sbitRateAnalysis(chamber_config, rateTree, cutOffRate=0.0, debug=False, outf
                 # channelor case
                 if perchannel == False :
 	            dict_dacInflectPts[dacName][ohKey][vfat] = findInflectionPts(dict_Rate1DVsDACNameX[dacName][ohKey][vfat])
-                    inflectTable.append([ohKey, vfat, dict_dacInflectPts[dacName][ohKey][vfat][0][0] ])
+                    inflectTable.append([ohKey, vfat, dict_dacInflectPts[dacName][ohKey][vfat][0] ])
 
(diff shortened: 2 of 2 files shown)


bregnery:
This is to address issue [#288]

cgalloni:
@bregnery, can you solve the conflicting files? Maybe with a rebase?

bregnery:
Ah, yes, sorry, I rebased to a previous tag

utils/anautilities.py
```diff
@@ -556,7 +556,8 @@ def findInflectionPts(graph):
     # We are defining the inflection point as the point where the most negative gradient sum begins
     try:
         inflxGrad = negGrad[bigIdx][0]
-        inflxIdx = np.where(grad == inflxGrad) #return the index at the specified value
+        inflxIdices = np.where(grad == inflxGrad) #return the index at the specified value
+        inflxIdx = np.amax(inflxIdices) # if there are multiple possible points, take the largest one
```
lpetre-ulb: Is the variable `inflxIdices` necessary?

utils/threshAlgos.py
```diff
@@ -1452,7 +1452,7 @@ def getDetName(entry):
                 # channelor case
                 if perchannel == False :
 	            dict_dacInflectPts[dacName][ohKey][vfat] = findInflectionPts(dict_Rate1DVsDACNameX[dacName][ohKey][vfat])
-                    inflectTable.append([ohKey, vfat, dict_dacInflectPts[dacName][ohKey][vfat][0][0] ])
+                    inflectTable.append([ohKey, vfat, dict_dacInflectPts[dacName][ohKey][vfat][0] ])
```
lpetre-ulb, line 1457: I guess this fixes the same bug as the one described in https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/268?

lpetre-ulb (dismissed):
Non blocking comments.

utils/anautilities.py
bregnery: Good point!

utils/threshAlgos.py
```diff
@@ -1498,24 +1498,13 @@ def getDetName(entry):
                 for vfat in range(0,nVFATS):
                     canv_Summary1D.cd(chamber_vfatPos2PadIdx[gemType][vfat]).SetLogy()
 
-                    # make sure the inflection point is there
```
jsturdy, line 1501: I would probably keep the line drawing functionality, but disable it by default if someone looking at the plot finds it too distracting

utils/threshAlgos.py
bregnery: Yeah, now that it is restricted to have one index, the data structure changed a bit.

utils/threshAlgos.py
bregnery: Okay, I have not added input options in python, so I will have to take a look at how to do this.

bregnery: Should I also disable by default the printing of the table? (The table will still be printed to the file)

bregnery:
I have added an option inside anautilities for drawing the lines on the graph and for printing the table output in the terminal. These options default to false. The inflection point values are still stored in a `.txt` file by default.

I have tested this on sbit scans that have multivalued inflection points, missing vfats, and bad sbit lines.

utils/threshAlgos.py
jsturdy: were there complaints about the table?
i think there are other tools that are printing tables to the output, so i wouldn't change that

bregnery: Okay, then I will default that option to true and leave it as an option if people complain about the table in the future :)

cgalloni (approved):

jsturdy (approved):

